PR #3, opened 2020-10-21
Add Actions for Archiving Dependencies

dragonstyle:
Deno, Pandoc, and TinyTex.

Depends upon setting up Buckets, which is not yet complete.

### DIFF
--- a/.github/tinytex.yml
+++ b/.github/tinytex.yml
@@ -1,0 +1,54 @@
+# This is a basic workflow that is manually triggered
+
+ name: Archive TinyTex
+
+ # Controls when the action will run. Workflow runs when manually triggered using the UI
+ # or API.
+ on:
+   workflow_dispatch:
+     # Inputs the workflow accepts.
+     inputs:
+       version:
+         # Friendly description to be shown in the UI instead of 'name'
+         description: 'TinyTex Version to Archive'
+         # Default value if no value is explicitly provided
+         default: ''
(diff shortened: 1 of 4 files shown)


jjallaire:
LGTM, merge when ready!

jjallaire (approved):

jjallaire:
We should add pandoc-crossref here as well

.github/tinytex.yml
```diff
@@ -0,0 +1,54 @@
...
+     runs-on: ubuntu-latest
+
+     # Steps represent a sequence of tasks that will be executed as part of the job
+     steps:
+     # Runs a single command using the runners shell
+     - name: Download Local Copies
+       run: |
+            for EXT in $EXTS; do
+             FILENAME=TinyTeX-${{ github.event.inputs.version }}.$EXT
+             curl -fail -L -o $FILENAME https://github.com/yihui/tinytex-releases/releases/download/${{ github.event.inputs.version }}/$FILENAME
```
jjallaire, line 36: Which variation are we targeting? (looks like there are 3: https://github.com/yihui/tinytex-releases#releases)

dragonstyle: I targeted the `TinyTeX.*` with this, but looking merely closely, I think we could get away with `TinyTeX-1.*` which would be enough to compile most RMarkdown documents. What do you think?

PR #63, opened 2021-07-26
Remove parenthesis for revealjs slideNumber option

### LINKED ISSUE #62: revealjs quoting of some option is not correct
Take this 
````markdown
---
title: test
format: 
  revealjs:
    slideNumber: 'c/t'
---

## slide 1

## Slide 2
````

It won't work as expected because Quarto is quoting and adding parenthesis `'(c/t)' `

````html
        // Display the page number of the current slide
        slideNumber: '(c/t)',
````

https://github.com/quarto-dev/quarto-cli/blob/d6a6f945df0fb1654942d13ec1e255db23e2e67d/src/format/reveal/format-reveal.ts#L46-L50

The parenthesis needs to be removed.

cc @apreshill I think this is the error you encountered

cderv:
slideNumber in revealjs can be boolean or string. 

Quarto, like rmarkdown, does a special handling of this option. However, there should not be some parenthesis I think (https://revealjs.com/slide-numbers/) 

This fix #62

Tested with `configure-windows.cmd` then rendering an example 
````markdown
---
title: test
format: 
  revealjs:
    slideNumber: 'c/t'
---

## slide 1

## Slide 2
````

### DIFF
--- a/src/format/reveal/format-reveal.ts
+++ b/src/format/reveal/format-reveal.ts
@@ -46,7 +46,7 @@ function revealHtmlPostprocessor() {
         // quote slideNumber
         scriptEl.innerText = scriptEl.innerText.replace(
           /slideNumber: (h[\.\/]v|c(?:\/t)?)/,
-          "slideNumber: '($1)'",
+          "slideNumber: '$1'",
         );
       }
     }


jjallaire:
Thanks!

PR #113, opened 2021-09-22
fix linux install after previous sudo install

jmbuhr:
When running the `./configure-linux.sh` command with sudo powers, quarto gets installed to `/usr/local/bin/quarto`. If you then later on attempt to install without `sudo`, quarto fails to delete the old symlink and doesn't attempt to install to  `~/bin/` as it would normally do because the attempt of symlink removal was outside of the `try` block. This PR adds another `try` block with a helpful error message.

### DIFF
--- a/package/src/common/configure.ts
+++ b/package/src/common/configure.ts
@@ -60,10 +60,14 @@ export async function configure(
     for (let i = 0; i < symlinkPaths.length; i++) {
       const symlinkPath = symlinkPaths[i];
       info(`> Trying ${symlinkPath}`);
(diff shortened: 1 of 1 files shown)


jjallaire:
Thanks!

jjallaire:
Hi @jmbuhr,  I am going back through PRs to make sure we have contributor agreements for all those that we've merged. If you don't mind it would be great to get one from you (thanks again for the contribution). Links to agreement and details on submitting are here: https://quarto.org/about.html#pull-requests. Thanks!

PR #115, opened 2021-09-24
Clean up some knitr code

yihui:
Since [**knitr** v1.34.2](https://github.com/yihui/knitr/blob/master/NEWS.md#changes-in-knitr-version-135), `partition_yaml_options()` has been ported to **knitr**, and `parse_block()` in **knitr** can handle YAML options, so you no longer need to mutate it.

This PR needs to wait until the next version of **knitr** is on CRAN, unless you install the dev version from Github in Quarto.


src/resources/rmd/patch.R
```diff
@@ -40,58 +40,6 @@ if (requireNamespace("htmlwidgets", quietly = TRUE)) {
...
-  
-  # strip trailing comma and whitespace
-  params.src <- sub("\\s*,?\\s*$", "", params.src)
-  
-  # look for other options to forward. note that ideally we could extract *all*
-  # parameters and then pass partitioned$code below, however we can construct
-  # cases where deparsed versions of the options include a newline, which causes
-  # an error. we'll wait and see if this capability is incorporated natively
-  # into knitr parse_block -- if it's not then we can pursue more robust versions
-  # of textual option forwarding that don't run into newlines
```
yihui, line 70: @jjallaire I guess this part is no longer necessary, since I parse these YAML options and merge them right away here: https://github.com/yihui/knitr/blob/c9b7e3acae3248e5a48beddd49e98d37d1cf1517/R/parser.R#L96-L97 (so you don't need to do it in `knitr_options_hook()` in `src/resources/rmd/hooks.R` above) Not sure if I missed or misunderstood anything.

jjallaire:
Great, thanks! I may still need to do the option forwarding (where I convert fig. options into fig- options for internal consumption -- Quarto supports both although it documents `fig-`).

I also will probably take the approach of doing this conditionally for now (i.e. I'll merge your changes now but keep the old codepaths around for earlier versions of knitr). 

So I'll take this PR as the starting point and create a new branch that does the above then assign you to review it when I think it's ready to go.

jjallaire:
Oh it looks like I actually turn `fig-` into `fig.` for internal processing (same idea, support both syntaxes with one set of downstream code)

jjallaire:
Here's the new PR: https://github.com/quarto-dev/quarto-cli/pull/116

jjallaire:
This is now integrated with this commit (https://github.com/quarto-dev/quarto-cli/commit/77cf45fdde0f23f2c8f10161234cbc4110a0e833) plus a follow up commit to take advantage of the params src behavior in knitr 1.34.4 so I'll close. Thanks!!!

PR #116, opened 2021-09-25
use native knitr yaml chunk option handling

jjallaire:
Use the native knitr handling for knitr >= 1.34.2 (the current dev version but not yet the CRAN version). 

@yihui The `original.params.src` as originally forwarded to Quarto included the engine. I noted that what you are currently forwarding does not include the engine so I added this workaround to make the tests pass:

https://github.com/quarto-dev/quarto-cli/compare/feature/knitr-native-yaml-options?expand=1#diff-8db8cafc98f0d728b640552494dd7aac586b9c8ba6f132707c4265276c79ed01R334-R343

If you are okay w/ changing knitr to pass the full original source including the engine then I can remove this bit.

### DIFF
--- a/src/resources/rmd/hooks.R
+++ b/src/resources/rmd/hooks.R
@@ -330,7 +330,17 @@ knitr_hooks <- function(format, resourceDir) {
           x <- trimws(x, "left")
         }
       }
-      x <- paste0("\n```{{", options[["original.params.src"]], "}}\n", yamlCode, x, '\n```')
+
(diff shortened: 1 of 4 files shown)


yihui:
> If you are okay w/ changing knitr to pass the full original source including the engine then I can remove this bit.

Sure! I just did it in knitr 1.34.3: https://github.com/yihui/knitr/commit/c081ef9c07bda8718479dd2821cc16a7b9eaf405

jjallaire:
Great, thanks!

PR #144, opened 2021-10-20
Support other attributes from RevealHighlight

cderv:
This PR brings support for other attributes from https://revealjs.com/code/

* `data-no-escape`
* `data-trim`

When passed as fenced code attributes, Pandoc will set them on the `<pre>`. We need to move them on the `<code>` so that this would work 

````markdown
```{.clojure data-trim=''}

    (def lazy-fib
      (concat
       [0 1]
       ((fn rfib [a b]
            (lazy-cons (+ a b) (rfib b (+ a b)))) 0 1)))
        
```
````

### DIFF
--- a/src/format/reveal/format-reveal.ts
+++ b/src/format/reveal/format-reveal.ts
@@ -309,6 +309,8 @@ function revealHighlightDefault(theme: string | undefined) {
 
 function revealHighlightHtmlPostprocessor() {
   const kDataLineNumbers = "data-line-numbers";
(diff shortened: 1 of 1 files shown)


cderv:
This makes sense only if we keep using RevealHighlight plugin obviously

cderv:
Closing this as we don't use highlight.js anymore after 5b6510d0da880da7c4d7f0fe18dc755369e518ea

Issue #92, opened 2021-08-17
--type: alias "website" as an alias for "site"


aronatkins:
The documentation talks about "website", yet the CLI only supports "site". Should the CLI support "website" as equivalent to "site"?

```
quarto create-project --type website
ERROR: Error: Project type must be one of default, site, book, but got "website".
    at Object.value (file:///Users/aron/dev/rstudio/quarto-cli/src/command/create-project/cmd.ts:35:17)
    at parseFlags (https://deno.land/x/cliffy@v0.19.3/flags/flags.ts:168:34)
    at Command.parseFlags (https://deno.land/x/cliffy@v0.19.3/command/command.ts:1113:22)
    at Command.parse (https://deno.land/x/cliffy@v0.19.3/command/command.ts:896:58)
    at Command.parse (https://deno.land/x/cliffy@v0.19.3/command/command.ts:870:33)
    at quarto (file:///Users/aron/dev/rstudio/quarto-cli/src/quarto.ts:59:64)
    at file:///Users/aron/dev/rstudio/quarto-cli/src/quarto.ts:76:11
```

jjallaire:
Yeah I think we should indeed do this. I'm not sure why we went with `site` given how that's not the natural thing to call it in. prose. Perhaps the real fix is to make it `website` in the CLI and just alias `site` for backwards compatibility.

PR #160, opened 2021-10-30
rename 'site' to 'website' in project config

jjallaire:


jjallaire:
@aronatkins This PR addresses https://github.com/quarto-dev/quarto-cli/issues/92

This also changes the format of `_quarto.yml` to `type: website` and `website` as the root key for website project config (the old format is still supported to migrate existing projects).

Does RSC depend on `type: site` in any way or is everything mediated through calls to `quarto render`, `quarto inspect`, etc.?

aronatkins:
RSC does not currently depend on `type`.

PR #161, opened 2021-11-02
Add process to update reveal.js in `update-html-dependencies`

cderv:
At first I did it to update on a github release tag but it seems a commit is better.
Other example on commit is for Bootstrap  and it is using `git`. I think we can do without GIT and did it accordingly for reveal.js. 

this is also in light of an issue discussed with @dragonstyle where on Windows `Deno.removeSync` does not manage to clean a folder cloned using Git. (because of Git access right on filesytem were read only files are created). 
This lead to adding Try Catch so the the scripts runs on Windows without terminating early due to error.

Regarding the reveal.js update: 

* We get `css/`, `dist/` and `plugin/`
* We remove unwanted `.css` in the `dist/theme/` folder

I choose to update the existing `updateGithubSourceCodeDependency()` function to support a commit in addition of a release tag. I used a new optional `commit` argument set to `false` by default. Another idea was do detect if the env var was a git commit or a version number using for example `.` detection as commit won't have any dots usually. It could be on the number of character too. Anyway, tell me if this is a better idea.

If the process seems valid, I could also change the way we get the bootstrap deps to use the same zip download. This would avoid using Git and would avoid the "Permission Denied" issue on Windows.

### DIFF
--- a/configuration
+++ b/configuration
@@ -25,6 +25,7 @@ export POPPER_JS=2.9.1
 export CLIPBOARD_JS=2.0.8
 export TIPPY_JS=6.3.0
 export PDF_JS=2.8.335
+export REVEAL_JS=4.1.2
 
 # javascript search dependencies
 export AUTOCOMPLETE_JS=1.4.0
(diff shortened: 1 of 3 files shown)


dragonstyle:
This looks great. I think I should update bslib to use this now that it supports commits!

cderv:
> I think I should update bslib to use this now that it supports commits!

once we do that, if there is not need to git clone, the remove problem should be solved. I'll check once the change is done for bootstrap

PR #173, opened 2021-11-08
First pass for FR translation + new translations

cderv:
Not sure about the context for all the keys.  I would need to render a check if the translations fit in their context. I'll amend later if I see something off. 

For example, not sure about the difference between 
```
search-more-match-text: "more match"
search-more-matches-text: "more matches"
```
`more` being plural, it would go with a singular word.


cderv:
I added also in this PR translation for a new "Copy to Clipboard Tooltip"

PR #176, opened 2021-11-10
Add translations for de

skeydan:
I agree with @yitao-li and @cderv that this is surprisingly difficult (not knowing the pragmatic context).

At least I had the advantage of being able to base my translation on two languages already (unfortunately, not knowing Mandarin).

### Questions/comments:
 
1.

Sometimes, I'm using punctuation, to more strongly convey the intention. For example:

```
callout-important-caption: "Wichtig!"
callout-danger-caption: "Achtung! Gefährlich!"
```

This assumes these are tooltips - which should be correct, right? - but should be fine even if not. 

However, these:

```
search-matching-documents-text: "Gefundene Dokumente:"
search-more-match-text: "Weitere Ergebnisse in diesem Dokument:"
search-more-matches-text: "Weitere Ergebnisse in diesem Dokument:"
```

assume that these are headings, or just text, but not tooltips, for example. (Such that the colons make sense.)
Is this assumption correct?

2. 
I agree with @cderv that the difference between these:

search-more-match-text
search-more-matches-text

is not clear. Perhaps we're misunderstanding something? (I followed @cderv in taking both as meaning the same, but perhaps that's not intended?) 

3.

I left the prefixes as-is, as I don't know of any common abbreviations for those anyway (in German even less than in English :-)).


jjallaire:
Thank you! Comments/ answers to various questions that have arisen:

1) I don't think you need the punctuation for warning/danger as these will already have an icon and color to provide emphasis (see https://quarto.org/docs/authoring/callouts.html#callout-types)

2) For the search translations, I don't thin you want the colon because these headings are indications that there are more matches that _could_ be shown (but are not shown by default). e.g. see: https://quarto.org/docs/websites/images/navbar-overlay.png

For the more matches ambiguity, one of them will have the number 1 in front of it and the other will have more than one, for example: "1 more match in this document" vs. "5 more matches in this document"

jjallaire:
Those prefixes are abbreviations for the full text version (e.g. "Figure" => "fig."

Given that, would it be better to use e.g.  "Schaubild" =>  "sch."

Without this I'm worried the user won't be able to interpolate the meaning of the english abbreviation. If you don't want to use an abbreviation you could even translate as "schaubild" ?

jjallaire:
@skeydan Does this look like the correct translation of "Appendices" to German: https://github.com/quarto-dev/quarto-cli/commit/42cf9d5e4d9d04f2a8725d5e47595ca39439c72b#diff-d2656269a433d8e95b71b772ef5b7526dc08396a7eb79c154ed1a8f6e5cf8285

skeydan:
totally!

PR #180, opened 2021-11-11
fix configuration for symlink bin directories

sijow:
Calling `ensureDirSync` on `"/home/user/bin"` for example will fail if `bin` is a symlink to a directory (because it's a file). Calling it on `"/home/user/bin/"` (by appending a path separator) will force resolving the link so bin is seen as a directory (if the link target is a directory). This mechanism is described at https://man7.org/linux/man-pages/man7/path_resolution.7.html 

This fixes the build on my computer where `~/bin` is a symlink.

### DIFF
--- a/package/src/common/configure.ts
+++ b/package/src/common/configure.ts
@@ -4,7 +4,7 @@
 * Copyright (C) 2020 by RStudio, PBC
 *
 */
-import { dirname, join } from "path/mod.ts";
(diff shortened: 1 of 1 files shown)


dragonstyle:
Thank you for the pull request! If you haven't already, would you mind completing a contributor agreement as described here:

https://quarto.org/about.html#pull-requests

Thanks again!

knuesel:
Sure! Just sent.

PR #189, opened 2021-11-18
Add patch for leaflet due to reveal scaling feature

cderv:
This PR add a path for leaflet when used in revealjs. 

Reveal with use scaling feature and this mess up the pointer on the map with tooltip
lealfet-tooltip-nok

With this plugin, it will "cancel" the zooming for the leaflet container so that the pointer x / y coordinates are correct to place the tooltip. 

lealfet-tooltip-ok

However it still need to be tweaked so that it is recalculated each time the slide scale changes using 
````
Reveal.on( 'resize', event => {
  // event.scale, event.oldScale, event.size
} );
````

Hence the draft PR for now

### DIFF
--- a/src/resources/formats/revealjs/plugins/support/support.js
+++ b/src/resources/formats/revealjs/plugins/support/support.js
@@ -78,23 +78,33 @@ window.QuartoSupport = function () {
     // check if leaflet is used
     if (window.L) {
       L.Map.addInitHook(function () {
-        const slides = deck.getSlidesElement();
-        const scale = deck.getScale();
+        function unScale(slides, scale) {
+          console.log(this);
+          const container = this.getContainer();
 
-        const container = this.getContainer();
+          // Cancel revealjs scaling on map container by doing the opposite of what it sets
+          // * zoom will be used for scale > 1
+          // * transform will be used for scale < 1
+          // As we change on every resize, we remove other value if it was previously set
+          if (slides.style.zoom) {
+            if (slides.style.transform) slides.style.transform = null;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jjallaire** [comment]
Looks good, ping me when the resize hook is in and I'll merge.

**Comment by cderv** [comment]
Ok I think this is good not - I manage to add the hook, hope it is done correctly. At least, it is working as I expect. 

I believe this will work with any format using leaflet - I tested with R and **leaflet** but there is nothing tied to htmlwidget. That is why I wanted to use `L.Map.getContainer()` and not find the div of class leaflet, just in case Leaflet map is added another way.

File: src/resources/formats/revealjs/plugins/support/support.js [hunk 1]
```diff
@@ -73,13 +73,54 @@ window.QuartoSupport = function () {
...
+        }
+
+        // bind the leaflet Map object to unscale
+        const unScale2 = unScale.bind(this);
+
+        // Unscale at initialization
+        unScale2(deck.getSlidesElement(), deck.getScale());
+
+        // And unscale each time presentation is resized
+        Reveal.on("resize", function (ev) {
```
cderv, line 109: Is `deck` == `Reveal` in the function and it should be `deck.on()`  instead ?

jjallaire: It should be `deck.on` (that allows plugins to deal w/ multiple decks on a page)

cderv:
```suggestion
        deck.on("resize", function (ev) {
```

PR #191, opened 2021-11-19
Add a fig-responsive execute option

cderv:
This is defaulting to TRUE for HTML bootstrap format only. This will mean no responsiveness for htmlwidgets in revealjs format in 
R

Adding `fig-responsive` as an execute option means that the new assignment of `htmlwidgets:::resolveSizing()` happens conditionally if the option is TRUE. Changes are done in `patch.R` and `execute.R`. 

Then `fig-responsive` default to FALSE for HTML bootstrap was trickier. Doing this meant to be able to set `format.execute["fig-responsive"] = TRUE` only for bootstrap format. In order to do that, I added a `executeFilter` following the logic of `metadataFilter` to be able to modify `execute` default option based on metadata set by the user. In our case the `theme` metadata defining is the format is Bootstrap HTML or not. I did not used `formatHasBootstrap()` because I was not sure how to use from within the format creation itself.  I did this logic for setting the default value conditionally to the theme metadata. 

@jjallaire I am opening this PR so that you can check the logic considering our previous discussion (I did not see how this would have fit in `bootstrapExtras` )

I need to tweak to not make it responsive if user has change the figure sizes.


src/format/html/format-html.ts
```diff
@@ -74,6 +77,23 @@ export function htmlFormat(
   return mergeConfigs(
     createHtmlFormat(figwidth, figheight),
     {
+      execute: {
+        [kFigResponsive]: true,
+      },
```
cderv, line 82:
```suggestion
```

This should not be there as we set the default in `executeFilter`

src/config/constants.ts
```diff
@@ -329,6 +329,8 @@ export const kCrossrefAppendixDelim = "appendix-delim";
 export const kCrossrefChaptersAlpha = "chapters-alpha";
 export const kCrossrefChapterId = "chapter-id";
 
+export const kFigResponsive = "fig-responsive";
```
cderv, line 332: Not sure if this config should go somewhere in one of the definition interface now it is a metadata.
Currently, it is nowhere defined that this is boolean. Should it go somewhere in `types.ts` ?

src/format/formats-shared.ts
```diff
@@ -96,6 +100,9 @@ export function createHtmlPresentationFormat(
   return mergeConfigs(
     createHtmlFormat(figwidth, figheight),
     {
+      metadata: {
+        [kFigResponsive]: false,
+      },
```
cderv, line 105: This should do it to set as default for all HTML presentation

src/format/html/format-html.ts
```diff
@@ -75,11 +76,11 @@ export function htmlFormat(
...
-        if (!!metadata[kMinimal] && metadata[kTheme] === undefined) {
-          return {
-            ...metadata,
-            theme: "none",
-          };
+        if (!metadata[kMinimal]) {
+          return mergeConfigs(metadata, {
+            [kFigResponsive]: false,
+            theme: metadata[kTheme] === undefined ? "none" : metadata[kTheme],
+          });
```
cderv, line 83: It will now be set to TRUE by default except when minimal is used. There is no more Bootstrap conditional as discussed

src/format/html/format-html.ts
```diff
@@ -75,11 +76,11 @@ export function htmlFormat(
     createHtmlFormat(figwidth, figheight),
     {
       metadataFilter: (metadata: Metadata) => {
-        if (!!metadata[kMinimal] && metadata[kTheme] === undefined) {
-          return {
-            ...metadata,
-            theme: "none",
-          };
+        if (metadata[kMinimal] === true) {
```
cderv, line 79: deno linter throw me a comment if I use `!!metadata[kMinimal]` to be only `!metadata[kMinimal]`. 
I used explicit form to avoid confusion as in 
https://github.com/quarto-dev/quarto-cli/blob/217e24c7d50731cdcbeec493a5543b7a679be59b/src/format/html/format-html.ts#L395

src/config/constants.ts
jjallaire: Yes, this should go in the `defaultFormat()` function. There is a bunch of other stuff you need to do for any new option -- I can check out the branch and do this once you give me the thumbs up it's okay to do

src/resources/rmd/execute.R
```diff
@@ -491,6 +494,42 @@ apply_slides_patch <- function(includes) {
...
+      default_knitr_options <-  knitr::opts_chunk$get()[size_options]
+      
+      resolveSizing <- function(x, sp, standalone, knitrOptions = NULL) {
+          # default sizing resolution
+          sizing <- htmlwidgets_resolveSizing(x, sp, standalone, knitrOptions)
+          
+          # Do not alter if user has change fig.height or fig.width option
+          if (!is.null(knitrOptions) && !identical(default_knitr_options, knitrOptions[size_options])) {
+            return(sizing)
+          }
```
cderv, line 514: This will do it per chunk. This means that if the `fig.*` options are changed on a specific chunk, it will not be modified but all the other chunk would have the transformation. 

If option is change globally `knitr::opts_chunk$set`, then all chunk would not be responsive. 

However, it is not perfect because `knitr::opts_chunk$get` will only get us the default knitr option, but not the default for the format that can be changed by Quarto. 

We could probably remove this and the two other related line above and do that later;

cderv (commented):
This is ready for the main part. 

Only the cheking of default option is not yet 100% complete.

src/config/constants.ts
cderv: Ok great. I tried to find the different places, but I must not have taken the correct other option as example. 

Or maybe because `defaultFormat()` does have yet an option for metadata. Anyway, I'll look at your change to learn

jjallaire: ah, good point. we don't actually provide defaults for metadata values

PR #194, opened 2021-11-20
Merge directory level metadata into project

dragonstyle:
Collect any metadata from `_metadata.yml` files along the path between the project and the input and merge into the project.


src/project/project-context.ts
```diff
@@ -296,54 +303,104 @@ export async function projectMetadataForInputFile(
   }
```
jjallaire, line 304: `projectContext` can return undefined if the file isn't in a project.

src/command/render/render.ts
```diff
@@ -918,13 +919,26 @@ async function resolveFormats(
...
+    ? directoryMetadataForInputFile(
+      project.dir,
+      dirname(target.input),
+    )
+    : {};
+  const projMetadata = mergeConfigs(
+    projMetadataForInput,
+    directoryMetadataForInput,
+  );
+
```
jjallaire, line 941: There is another consideration here concerning the logic for ordering the formats. Let's discuss realtime on Monday.

Issue #198, opened 2021-11-23
quarto can't parse my metadata since v2.232 


benabel:
Hi and first thanks for this awesome project.

I have a problem since version [0.2.232](https://github.com/quarto-dev/quarto-cli/commit/fd6d8cd9709c23f38bc76617fab29fc578358033) introduced an error on my metadata parsing:

```markdown
---
matter: "NSI"
classe: "Terminale"
partie: "P1&P4"
sectionPrefix: "Partie~"
subsectionPrefix: "Exercice~"
format:
  latex:
    papersize: a5
    classoption:
      - fontsize=10pt
      - DIV=calc
---

# Devoir 3
```

When i run quarto I go onto this:

```
ERROR: YAMLError: can not read a block mapping entry; a multiline key may not be an implicit key at line 6, column 7:
    format:
```

However this is valid `yaml` and pandoc can parse it. `pandoc -t json devoir3.qmd | jq '.meta' `.

dragonstyle:
Thanks for taking the time to report this! Starting with that release, we do throw a hard exception when the YAML fails to parse (vs swallowing the error), so this is a likely culprit.

When I use the simple example above, however, I'm not reproducing the issue. That is making me think that some other YAML / metadata is in play (either further down in the document, in a `_quarto.yml` project file, or elsewhere. Can you share the complete file that reproduces the issue or project file that goes with this (if it is in a project)?

benabel:
Thank you for your quick response, in fact this was due to an error in another file of the project. I found which file in the project had a invalid metadate throwing an error as I proposed in PR #199.

benabel:
It is fixed in #187.

PR #199, opened 2021-11-23
Throw error when yaml Metadata can't be parsed

benabel:
Related issue #198


jjallaire:
Thanks @benabel. @cscheid actually has a branch that is about to be merged (https://github.com/quarto-dev/quarto-cli/pull/187) that does schema and syntax validation of YAML (so not only do we throw but we throw with the context of exactly what failed to parse or validate and why.

I will close this PR in deference to those changes -- @cscheid maybe just take a look at this PR to make sure it's intent/function is captured in your branch.

benabel:
Ok I tried the feature/yaml-validation branch, it allows to debug easily the issue evoqued in #198.

PR #218, opened 2021-12-13
figcaption styling in slides 

cderv:
Currently in Quarto presentation we got this 
* One image 
```markdown
## Insert an image

Without stretch feature

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg)
```

image

* 2 images
```markdown
## No strech when more than one image

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg)

And another 

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg)
```

image

It seems it would be better to align the `figcaption` where there is one with the image

This PR does that. 

I would also style the caption differently (maybe smaller font size ?) - would that be something to add ? 

Obviously, when I get the `.stretch` working, we either remove the caption or keep it, and I still think it should be centered. 

After this PR. 
image

image

### DIFF
--- a/src/resources/formats/html/_quarto-rules.scss
+++ b/src/resources/formats/html/_quarto-rules.scss
@@ -80,14 +80,23 @@
   width: 100%;
   margin-bottom: 0;
 }
-.quarto-figure-left > figure > p {
-  text-align: left;
+.quarto-figure-left > figure {
+  > p,
+  > figcaption {
+    text-align: left;
+  }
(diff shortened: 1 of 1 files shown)


jjallaire:
I agree about the smaller font for captions in reveal. There is a `styles.html` in the reveal resources folder where you can add this (maybe use ~ 0.5em?)

PR #224, opened 2021-12-15
Automatically add stretch class on images when only one in slides

cderv:
This PR is leveraging the stretch feature from revealjs: https://revealjs.com/layout/#stretch

## Current Status

Sharing this so that we can discuss and you can test. 

## Idea

Automatically apply the strech class on images when there is only one image in slides. The reveal feature is normally based on adding the class manually to the image, and currently only one image can be stretch per slide.

The plan is to use a post process of the HTML to modify the HTML code so that reveal process can apply as expected.

## How it works for reveal ? 

reveal will apply a transformation when `Reveal.layout()` for this stretch class 
https://github.com/hakimel/reveal.js/blob/b7d65be051072c4f36334403f0cdc19f88ce09c3/js/reveal.js#L954

* Detect any **direct child** of `<section>` with the class `stretch` (or `r-streched`) 
* Calculate remaining Height
* Apply to the element using inline style attributes

## How image works in Quarto ? 

Pandoc has a specific detection and process for img of class `stretch`. Usually non caption `` are put inside a `<p>` as any other element in Pandoc, and caption image inside a figure element (`implicit_figure` extension), but when class stretch is there, pandoc will put the `` first level, completely not considering the caption. 

Quarto will do as pandoc for non captioned, but will have its own process for figure environment putting the img into a div, keeping the caption in `<figure>` and `<figcaption>`. When plot are created using execution engine, the figure div is further nest inside other div. 

This means that adding the `stretch` class on image is not the only thing to do to make it works.

## What this PR is doing

* If we detect only one `` in a slide
* We add the `stretch` class if not already there 
* Then we move the `` has direct child of the slide `<section>`
* We search for a potential `ficaption` and we move its content inside a `<p class="caption">` just below the ``

## (De)activation switch 

For testing but also could be useful to deactivate manually, I added a `stretch-auto` option (maybe shorter `stretch: true` is enough. 

## Testing

Below the file with which I tested

<details>
<summary>Details</summary>

````markdown
---
title: "test stretched"
format: 
  revealjs: 
    auto-stretch: true
keep-md: true
---

## Insert an image

Without stretch feature

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg)

## Insert an image with code conten below

Without stretch feature

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg)

```{r, echo = TRUE}
1 + 1 
```

## Insert an image with no caption

Without stretch feature

![](https://revealjs.com/images/logo/reveal-black-text.svg)

## Insert an image with right alignment

Without stretch feature

![](https://revealjs.com/images/logo/reveal-black-text.svg){fig-align='right'}

## Insert an image with caption text

![Elephant](https://raw.githubusercontent.com/quarto-dev/quarto-web/main/docs/authoring/elephant.png)

## Column layout

:::: {.columns}

::: {.column width="40%"}
On the right a nice elephant
:::

::: {.column width="60%"}
![Elephant](https://raw.githubusercontent.com/quarto-dev/quarto-web/main/docs/authoring/elephant.png)
:::

::::

## Insert an image with alt text

![](https://raw.githubusercontent.com/quarto-dev/quarto-web/main/docs/authoring/elephant.png){fig-alt="A drawing of an elephant."}

## Insert an image with alt text, caption and title

![Elephant](https://raw.githubusercontent.com/quarto-dev/quarto-web/main/docs/authoring/elephant.png "Title: An elephant"){fig-alt="A drawing of an elephant."}

## Insert an image specifying `.strech` explicitly (no caption)

With adding stretch feature

![](https://revealjs.com/images/logo/reveal-black-text.svg){.stretch}

## Insert an image specifying `.strech` explicitly (with caption)

Using the `.strech` class

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg){.stretch}

## No strech when more than one image

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg)

And another

![revealjs logo](https://revealjs.com/images/logo/reveal-black-text.svg)

## Using knitr to create a plot

```{r, echo = TRUE, fig.cap = "Knitr figure"}
plot(cars)
```

## Using knitr to create a plot without echo

```{r, echo = FALSE, fig.cap = "Knitr figure"}
plot(cars)
```

## Using knitr to create a plot without other text

```{r, echo = FALSE, fig.cap = "Knitr figure"}
plot(cars)
```

On this plot, we have interesting informations.

## Using knitr to specify alignement

```{r, echo = FALSE, fig.cap = "Knitr figure", fig.align='right'}
plot(cars)
```

## Slide with python code

```{python}
#| fig-cap: "Polar axis plot"
#| fig-alt: "A line plot on a polar axis"

import numpy as np

r = np.arange(0, 2, 0.01)
theta = 2 * np.pi * r
fig, ax = plt.subplots(subplot_kw={'projection': 'polar'})
ax.plot(theta, r)
ax.grid(True)
plt.show()
```

## Is a widget breaking the feature ? 

```{r}
library(plotly)
fig = list(
  data = list(
    list(
      x = c(1, 2, 3),
      y = c(1, 3, 2),
      type = 'bar'
    )
  ),
  layout = list(
    title = 'A Figure Specified By R List',
    plot_bgcolor='#e5ecf6', 
         xaxis = list( 
           zerolinecolor = '#ffff', 
           zerolinewidth = 2, 
           gridcolor = 'ffff'), 
         yaxis = list( 
           zerolinecolor = '#ffff', 
           zerolinewidth = 2, 
           gridcolor = 'ffff')
  )
)
# To display the figure defined by this list, use the plotly_build function
plotly_build(fig)
```

````

</details>

### CODE REVIEW

File: src/format/reveal/format-reveal.ts [hunk 1]
```diff
@@ -37,6 +37,7 @@ import {
   revealMuliplexPreviewFile,
   revealMultiplexExtras,
 } from "./format-reveal-multiplex.ts";
+import { info, warning } from "https://deno.land/std@0.97.0/log/mod.ts";
```
jjallaire, line 40: this should be imported from `log/mod.ts` (not the full url)

src/resources/formats/revealjs/styles.html
```diff
@@ -71,4 +71,28 @@
   .reveal .slide figure > figcaption {
     font-size: 0.5em
   }
+
+  .reveal .slide img.stretch + p.caption {
```
jjallaire, line 75: Remove the `text-align` as we are no longer centering figure captions

src/format/reveal/format-reveal.ts
cderv: This should be removed in fact

src/resources/formats/revealjs/styles.html
cderv: Removed and the other rule is moved to quarto.scss

src/format/reveal/format-reveal.ts
```diff
@@ -223,6 +224,13 @@ export function revealjsFormat() {
           metadataOverride.previewLinks = false;
         }
 
+        // specify auto-stretch if there is no boolean 'auto-stretch'
+        const autoStretch =
+          typeof (format.metadata[kAutoStretch]) !== "boolean";
+        if (autoStretch) {
+          metadataOverride[kAutoStretch] = true;
+        }
```
cderv, line 232: I change the way the default `auto-stretch` value is done to do it like `controlsAuto ` so that it is `true` by default, even in the case `auto-stretch` is set to a non boolean. Previously, if user set to non boolean, there was no check of the value

Is that ok ?

jjallaire: I don't think we need to do this here (that's there because `controls` is also a valid reveal setting that only takes a boolean so we need the sideband `controlsAuto`.

I think it's preferable to do it the way you had it before (note that we are going to validate the schema so they'll actually get a runtime error if they set it to a non-boolean)

cderv: > (note that we are going to validate the schema so they'll actually get a runtime error if they set it to a non-boolean)

Oh that is the part I did not get - I found the file. Do I need to something in `\src\resources\schema\format-metadata-reveal.yml` now ? Or this will be for later ?

jjallaire: That work has continued on the `feature/yaml-validation` branch so you can add the entry to that branch (or I have a note to take care of this myself as I am frequently on that branch)

cderv:
Ok, since our last discussion I have made the changes: 

* Option is now `auto-stretch`: it will control the automatic addition of the class `stretch` on images in the presentation. 
* but no matter the option, all images with the class `stretch` will be unwrapped on first level of the slide.
* However, images in columns layout are left alone 
* CSS rule is put in the scss file
* Empty leftover are removed (thanks @dragonstyle for the help!) 

Last thing would be the caption. For now I do nothing and caption are not specifically aligned. This mean we could get this when no stretch
image

or this when image is made bigger
image

Probably caption will be longer, but in slide I would tend to think shorter. It is quite subjective so I left it like that. Only the font size is set to `#{$presentation-font-smaller}em`

I edit the top comment with my latest example qmd test file for this. Related to this, should tests be added somewhere for features like ? I don't know how the test suits works for now.

jjallaire:
Okay, this all looks great! I agree we should leave the caption as-is. I will merge this now and we can handle widgets/ojs in another PR.

jjallaire:
I noticed that reveal supports both `stretch` and `r-stretch` (although they only document the latter). Since we also document e.g. `r-fit` I think should also support both but document `r-stretch`. Here's a small change that does this: https://github.com/quarto-dev/quarto-cli/commit/cb602eeb4c71af4480b3386e04124ded76c52de2 (LMK if this looks okay to you)

jjallaire:
Updated docs are here: https://quarto.org/docs/presentations/revealjs/advanced.html#stretch

Some questions:

(1) I say in the docs that `auto-stretch` only applies to top level images. Is that correct? I think it should be (as e.g. an inline image should not be stretched). 

(2) We exclude auto-stretch behavior for columns, should we also do so for layout blocks?

(3) We probably need a way for users to opt-out of auto stretch for a single slide or image. How about a `.nostretch` class (that we'd look for on either the `<section>` or on the individual image)

jjallaire:
Another thing I think we should do is not apply stretching to images that have an explicit `height` set on them. (IIRC knitr will only set the width on generated markdown?). Let's discuss realtime if this doesn't seem like an obviously correct idea.

cderv:
> I noticed that reveal supports both `stretch` and `r-stretch` (although they only document the latter). Since we also document e.g. `r-fit` I think should also support both but document `r-stretch`. 

I went with `stretch` because Pandoc only supports this class from revealjs. https://pandoc.org/MANUAL.html#extension-implicit_figures

https://github.com/jgm/pandoc/blob/4b220d592c163efb0bf5a0e1db2ed7d6be9c69e7/src/Text/Pandoc/Writers/HTML.hs#L737

So I thought at first we should use this class.
But now that we do the unwrapping ourself I think we don't need Pandoc support. So that should be fine. Thanks !

cderv:
> Another thing I think we should do is not apply stretching to images that have an explicit `height` set on them. (IIRC knitr will only set the width on generated markdown?). Let's discuss realtime if this doesn't seem like an obviously correct idea.

Good point. I need to check what **knitr** is doing exactly in the different cases but I also think that only width will be set by default.

cderv:
> (1) I say in the docs that auto-stretch only applies to top level images. Is that correct? I think it should be (as e.g. an inline image should not be stretched).

I forgot to test this case with inline images. Current does do anything special to ignore them. So I need to check how that behave and adjust. 

> (2) We exclude auto-stretch behavior for columns, should we also do so for layout blocks?

So this : https://quarto.org/docs/authoring/figures-and-layout.html#block-layout but with only one image in the block ? 
I assumed those layout would be used with several images. We should indeed ignore them. 

> (3) We probably need a way for users to opt-out of auto stretch for a single slide or image. How about a .nostretch class (that we'd look for on either the <section> or on the individual image)

As it applies only on slides with one image, I would go with the `.nostretch` class in the image.

jjallaire:
My only concern about only supporting the image is that for code chunks it might be tricky for people to get classes on the image (unless we already do this pretty cleanly). If we did it at the slide level then we could also support a cell option for `auto-stretch: false`

cderv:
> for code chunks it might be tricky for people to get classes on the image (unless we already do this pretty cleanly)

For code chunk, one need to use the `out.extra` chunk option. The value will be used as content to add into `` in HTML or inserted into `\includegraphics[]` in LaTeX. 

````markdown
---
title: "test"
output: 
  html_document:
    keep_md: true
---

```{r}
#| out.extra: .nostretch
plot(cars)
```
````

would produce this in `md` file
````
<img src="test_files/figure-html/unnamed-chunk-1-1.png" .nostretch />
````

It could probably be make more explicit with a mechanism in quarto. 

But I see the advantage of slide level configuration.

**Comment by jjallaire** [comment]
As written: `` this won't actually work (it's not valid HTML). It would need to be `` so therefore the user would need to write:

````markdown
```{r}
#| out.extra: "class='nostretch'"
plot(cars)
```
````

Which I think is kind of obscure. +1 on the real motivation here being slide level config.

cderv:
Oh indeed. I went to fast. I'll try to add the slide level config.

cderv:
For reference, I took care of all the above comment in #231.

PR #234, opened 2021-12-22
deno: write to stdin in 4k chunks

jjallaire:
@cscheid This resolves the issue you observed with writing large files/buffers to stdin. Could you do a review? (I find that i/o loops are very commonly the source of unexpected bugs so would love the sanity check).

I rendered `quarto-web` with both the 4k window size in the code now but also w/ a much smaller window (3 bytes) to make sure we are windowing correctly and getting the end boundary right.


jjallaire:
@cscheid I went ahead and merged this b/c I am fairly confident given my tests, but if you could just give the code a once over that would be great (note that another unrelated commit got inadvertently added here, you can ignore that).

PR #285, opened 2022-01-25
add custom R syntax file with new pipe support

cderv:
This PR add the `r.xml` file with new pipe support. 

I have PR this file in KDE repo (https://invent.kde.org/frameworks/syntax-highlighting/-/merge_requests/295) and it has been merged. It means this should be available in next skylighting version probably and hopefully next Pandoc version. 

So maybe you prefer to wait for this change to be included into Pandoc.


jjallaire:
Thanks! I'll merge it now then remove it when Pandoc has this included.

PR #289, opened 2022-01-25
Update import_map.json for deno stdlib 0.122.0

jmablog:
Currently (quarto version 0.3.46), the deno stdlib doesn't work when importing via Typescript for use in post-render files, with the error message "error: Specifier not found in cache: "https://deno.land/std@0.119.0/fs/ensure_dir.ts", --cached-only is specified."

Looking through the source code, it seems a few days ago the stdlib for deno was updated to 0.122.0 in 'src/resources/deno_std/deno_std.lock' but not in the import map here.


jjallaire:
Thanks!

PR #301, opened 2022-02-01
Feed Support for Listings

dragonstyle:
- Generate feeds for listing pages (optionally)
- Feeds support `partial` or `full-content`
- Feeds are page level, so will collect all the items from any listings on the page
- You can also generate category specific feeds like so:
```
listing:
  feed:
    categories:
      - category1
      - category2
```
- In addition, now listing pages will automatically render whenever a page they reference is rendered (or when a new file that matches their contents is rendered).

### DIFF
--- a/src/project/types/website/listing/website-listing-shared.ts
+++ b/src/project/types/website/listing/website-listing-shared.ts
@@ -102,6 +102,7 @@ export interface ListingFeedOptions {
   [kType]?: "partial" | "full";
   [kLanguage]?: string;
   [kDescription]?: string;
+  [kFieldCategories]?: string[];
 }
 
 export interface ListingSharedOptions {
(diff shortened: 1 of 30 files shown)


src/project/types/website/listing/website-listing-shared.ts
```diff
@@ -87,9 +96,20 @@ export type CategoryStyle =
   | "category-unnumbered"
   | "category-cloud";
 
+export interface ListingFeedOptions {
+  [kTitle]?: string;
+  [kItems]?: number;
+  [kType]: "partial" | "full";
```
jjallaire, line 102: Perhaps this should be `summary` and `full` ?

src/project/types/website/website-config.ts
```diff
@@ -103,9 +104,11 @@ export interface OpenGraphConfig {
 
 type WebsiteConfigKey =
   | "title"
+  | "image"
```
jjallaire, line 107: Be sure to add `image` and `site-description` to the website schema (or just note that it should be done when creating the documentation)

dragonstyle: Will do as a part of listing schema/doc update.

PR #322, opened 2022-02-08
Listing Translation Ru

dragonstyle:


dragonstyle:
Hi @gjirokastrelbasanit34 - if you have a few moments and would be willing, could you review these suggested translations for some new capabilities in Quarto (pinging you just because I see that you've contributed translations in the past)?

I'm happy to answer questions to provide context for the usage of these strings, or you can also check documentation for the feature here:

https://quarto.org/docs/websites/website-listings.html#overview

PR #324, opened 2022-02-08
Listing Translation Zh

dragonstyle:


dragonstyle:
Hi @yitao-li - if you have a few moments and would be willing, could you review these suggested translations for some new capabilities in Quarto (pinging you just because I see that you've contributed translations in the past)?

I'm happy to answer questions to provide context for the usage of these strings, or you can also check documentation for the feature here:

https://quarto.org/docs/websites/website-listings.html#overview

src/resources/language/_language-zh.yml
```diff
@@ -51,3 +51,22 @@ crossref-eq-prefix: "方程式"
 crossref-lof-title: "插图目录"
 crossref-lot-title: "列表目录"
 crossref-lol-title: "列表目录"
+
+listing-page-order-by: "排序方式"
+listing-page-order-by-default: "缺省"
+listing-page-order-by-date-asc: "最老"
```
yitao-li, line 57:
```suggestion
listing-page-order-by-date-asc: "按日期升序排列"
```

src/resources/language/_language-zh.yml
```diff
@@ -51,3 +51,22 @@ crossref-eq-prefix: "方程式"
...
+listing-page-order-by-date-desc: "最新"
```
yitao-li, line 58:
```suggestion
listing-page-order-by-date-desc: "按日期降序排列"
```

src/resources/language/_language-zh.yml
```diff
@@ -51,3 +51,22 @@ crossref-eq-prefix: "方程式"
...
+listing-page-order-by-number-desc: "降序"
```
yitao-li, line 59: assuming "number" means page number here then the translation would be this:

```suggestion
listing-page-order-by-number-desc: "页码降序"
```

src/resources/language/_language-zh.yml
```diff
@@ -51,3 +51,22 @@ crossref-eq-prefix: "方程式"
...
+listing-page-order-by-number-asc: "升序"
```
yitao-li, line 60: (same as above)
```suggestion
listing-page-order-by-number-asc: "页码升序"
```

src/resources/language/_language-zh.yml
dragonstyle: This one is a little confusing - these don't representing ordering by page numbers, but instead more generally ordering by a numeric value. 

So for example, we currently say

'High to Low' in the en translation, but it could also be 'Largest to smallest' or 'Descending'. I very much appreciate your help and would love a recommendation given this new context. Thank you!!

yitao-li: OK. Then I think the original version is fine.

dragonstyle: Thank you very much with your help with this- we really appreciate it!

PR #340, opened 2022-02-15
Write tlmgr command in a tempfile on Windows and execute the file

### LINKED ISSUE #336: missing font not correctly detected on Windows with latexmk
Test document:

````markdown
---
title: "example"
format:
  pdf:
    pdf-engine: lualatex
    mainfont: Open Sans
    sansfont: Open Sans
    keep-tex: true
---

# content

````

**OpenSans** fonts is not correctly auto-installed. Log in Job pane is: 

````
updating existing packages
finding package for OpenSans(-(Bold|Italic|Regular).*)?[.](tfm|afm|mf|otf|ttf)

compilation failed- no matching packages
Package fontspec Error: The font "OpenSans" cannot be found.
````

With R Markdown and **tinytex** it works ok: 

<details>
<summary>With this sample file</summary>

````markdown
---
title: "example"
output:
  pdf_document: 
    latex_engine: lualatex
    keep_tex: true
mainfont: Open Sans
sansfont: Open Sans
---

Content
````

</details>

This took me a while to debug this and here is where I am at. 

Quarto command seems to correctly searching for `OpenSans(-(Bold|Italic|Regular).*)?[.](tfm|afm|mf|otf|ttf)` using `tlmgr search`. 
However, running in debug mode, I find that the issue is a quoting issue in Windows command line because the result of 
```typescript
const result = await tlmgrCommand(
        "search",
        [...args, ...(opts || []), searchTerm],
        true,
      );
```
is an error 
````
code:255
stderr:''Italic' n’est pas reconnu en tant que commande interne\r\nou externe, un programme exécutable ou un fichier de commandes.\r\n'
stdout:''
success:false
````

The `stderr` means 
> 'Italic' is not recognized as a internal or external command, an executable program or a command file. 

I believe this is because `|` is not seems as part of an argument to `tlmgr search` the command being something like 
````
tlmgr search --file --global OpenSans(-(Bold|Italic|Regular).*)?[.](tfm|afm|mf|otf|ttf)
````
as I am not sure `Deno.Run()` is quoting by default. 

It seems like adding double quote will help on Windows

* No quotes = it will error 
````powershell
> await Deno.run({cmd: ["cmd.exe", "/c", "tlmgr", "search", "--file", "--global", "OpenSans(-(Bold|Italic|Regular).*)?[.
](tfm|afm|mf|otf|ttf)"]}).status()
'Italic' n’est pas reconnu en tant que commande interne
ou externe, un programme exécutable ou un fichier de commandes.
````
* With quotes = no error 
````powershell
> await Deno.run({cmd: ["cmd.exe", "/c", "tlmgr", "search", "--file", "--global", "\"OpenSans(-(Bold|Italic|Regular).*)?
[.](tfm|afm|mf|otf|ttf)\""]}).status()
tlmgr.pl: package repository https://ctan.mines-albi.fr/systems/texlive/tlnet (verified)
{ success: true, code: 0 }
````

So this is a first potential issue. However, even by adding quotes in Quarto code around the arg, I do not get the correct output. No output is returned by the command in `result.stdout`. 

I have found this issue in Deno (https://github.com/denoland/deno/issues/8852) but not sure that is directly related to why no result.

So I am still looking into this.

### PR DESCRIPTION
**Pull request by cderv**
This PR fixes #336

Due to  some quoting issue with `Deno.Run()` , we have an error with `tlmgr` when passed to `cmd /c`.
Writing to a file prevent any quoting difficulty.

With this PR 

* It builds the command and write it to a file on Windows
* It will run `cmd /c tempfile.bat` instead of the command passed as a string.

### DIFF
--- a/src/command/render/latexmk/texlive.ts
+++ b/src/command/render/latexmk/texlive.ts
@@ -7,7 +7,7 @@
 import { info } from "log/mod.ts";
 import * as ld from "../../../core/lodash.ts";
 
-import { execProcess } from "../../../core/process.ts";
+import { execProcess, ProcessResult } from "../../../core/process.ts";
 import { kLatexHeaderMessageOptions } from "./types.ts";
 import { lines } from "../../../core/text.ts";
 import { removeIfExists } from "../../../core/path.ts";
@@ -273,41 +273,46 @@ async function verifyPackageInstalled(
   return result.stdout?.trim() === pkg;
 }
 
-async function safeTlmgrExec(
+// Execute tlmgr on Windows by writing command to a tempfile
+// and exectute the file with CMD
+async function safeTlmgrWindowsExec(
   cmd: string,
   args: string[],
-  _quiet?: boolean,
+  fnExec: (cmdExec: string[]) => Promise<ProcessResult>,
 ) {
-  // Function to execute tlmgr command
-  const execTlmgr = (tlmgrCmd: string[]) => {
-    const result = execProcess(
-      {
-        cmd: tlmgrCmd,
-        stdout: "piped",
-        stderr: "piped",
-      },
-    );
-    return result;
-  };
-
-  if (Deno.build.os === "windows") {
-    // On Windows writing the command to a file to avoid quoting issue with Deno
-    // https://github.com/quarto-dev/quarto-cli/issues/336
-    // Quoting every argument for CMD
-    args = args.map((a) => `"${a}"`);
-    const lines = ["tlmgr", cmd, ...args];
-    const tempFile = Deno.makeTempFileSync(
-      { prefix: "tlmgr-cmd", suffix: ".bat" },
-    );
-    try {
-      Deno.writeTextFileSync(tempFile, lines.join(" ") + "\n");
-      return await execTlmgr(["cmd", "/c", tempFile]);
-    } finally {
-      removeIfExists(tempFile);
-    }
-  } else {
-    return execTlmgr(["tlmgr", cmd, ...args]);
+  const lines = ["tlmgr", cmd, ...args];
+  const tempFile = Deno.makeTempFileSync(
+    { prefix: "tlmgr-cmd", suffix: ".bat" },
+  );
+  try {
+    Deno.writeTextFileSync(tempFile, lines.join(" ") + "\n");
+    return await fnExec(["cmd", "/c", tempFile]);
+  } finally {
+    removeIfExists(tempFile);
   }
+}
+
+function needQuoting(
+  args: string[],
+) {
+  let needQuoting = false;
+  if (Deno.build.os === "windows") {
+    // On Windows, we need to check if arguments may need quoting to avoid issue with Deno.Run()
+    // https://github.com/quarto-dev/quarto-cli/issues/336
+    const shellCharReg = new RegExp("[ <>()|\\:&;#?*']");
+    args = args.map((a) => {
+      if (shellCharReg.test(a)) {
+        needQuoting = true;
+        return `"${a}"`;
+      } else {
+        return a;
+      }
+    });
+  }
+  return {
+    status: needQuoting,
+    args: args,
+  };
 }
 
 function tlmgrCommand(
@@ -315,9 +320,27 @@ function tlmgrCommand(
   args: string[],
   _quiet?: boolean,
 ) {
+  // Function to execute tlmgr command
+  const execTlmgr = (tlmgrCmd: string[]) => {
+    return execProcess(
+      {
+        cmd: tlmgrCmd,
+        stdout: "piped",
+        stderr: "piped",
+      },
+    );
+  };
+  // tlmgr command dependending on OS
+  const tlmgr = Deno.build.os === "windows"
+    ? ["cmd", "/c", "tlmgr"]
+    : ["tlmgr"];
+
+  // Is safe execution needed because of quoting issue ?
+  const safeExecNeeded = needQuoting(args);
   try {
-    const result = safeTlmgrExec(cmd, args, _quiet);
-    return result;
+    return safeExecNeeded.status
+      ? safeTlmgrWindowsExec(cmd, safeExecNeeded.args, execTlmgr)
+      : execTlmgr([...tlmgr, cmd, ...args]);
   } catch {
     return Promise.reject();
   }
--- a/src/command/render/latexmk/texlive.ts
+++ b/src/command/render/latexmk/texlive.ts
@@ -54,7 +54,7 @@ export async function findPackages(
     } else {
       const result = await tlmgrCommand(
         "search",
-        [...args, ...(opts || []), `"${searchTerm}"`],
+        [...args, ...(opts || []), searchTerm],
         true,
       );
 
@@ -289,6 +289,8 @@ async function tlmgrCommand(
   let cmdExec: string[];
   let tempFile: string | undefined;
   if (Deno.build.os === "windows") {
+    // quote args on Windows
+    args = args.map((a) => `"${a}"`);
     // writing to a file to exectute to avoid quoting issue with Deno
     // https://github.com/quarto-dev/quarto-cli/issues/336
     const lines = ["tlmgr", cmd, ...args];
--- a/src/command/render/latexmk/texlive.ts
+++ b/src/command/render/latexmk/texlive.ts
@@ -10,6 +10,7 @@ import * as ld from "../../../core/lodash.ts";
 import { execProcess } from "../../../core/process.ts";
 import { kLatexHeaderMessageOptions } from "./types.ts";
 import { lines } from "../../../core/text.ts";
+import { removeIfExists } from "../../../core/path.ts";
 
 const tlmgr = Deno.build.os === "windows"
   ? ["cmd.exe", "/c", "tlmgr"]
@@ -53,7 +54,7 @@ export async function findPackages(
     } else {
       const result = await tlmgrCommand(
         "search",
-        [...args, ...(opts || []), searchTerm],
+        [...args, ...(opts || []), `"${searchTerm}"`],
         true,
       );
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/command/render/latexmk/texlive.ts [hunk 1]
```diff
@@ -53,7 +54,7 @@ export async function findPackages(
     } else {
       const result = await tlmgrCommand(
         "search",
-        [...args, ...(opts || []), searchTerm],
+        [...args, ...(opts || []), `"${searchTerm}"`],
```
cderv, line 57: I think those quotes were part of the problem also and need to be added for CMD to work

src/command/render/latexmk/texlive.ts
```diff
@@ -285,14 +286,33 @@ function tlmgrCommand(
...
   try {
     const result = execProcess(
       {
-        cmd: [...tlmgr, cmd, ...args],
+        cmd: cmdExec,
         stdout: "piped",
         stderr: "piped",
       },
     );
+    result.finally(() => {
```
jjallaire, line 311: Our typical idiom if we are going to wait on the promise would be to use `await` (and then you can make the finally a normal try/finally statement)

src/command/render/latexmk/texlive.ts
```diff
@@ -10,6 +10,7 @@ import * as ld from "../../../core/lodash.ts";
 import { execProcess } from "../../../core/process.ts";
 import { kLatexHeaderMessageOptions } from "./types.ts";
 import { lines } from "../../../core/text.ts";
+import { removeIfExists } from "../../../core/path.ts";
 
 const tlmgr = Deno.build.os === "windows"
```
cderv, line 15: I did not touch this function but it is now only used by the following `hasTexLive()` which can probably be rewritten with 

````ts
await tlmgrCommand({
  cmd: ["--version"]
})
````

but it was not part of the issue so I did not touched.

Not sure about the policy about trying to leave as much as possible for tiny changes, or trying to refactor a bit on the go when we modify code base like that. You'll tell me.

File: src/command/render/latexmk/texlive.ts [hunk 4]
```diff
@@ -285,14 +286,33 @@ function tlmgrCommand(
...
+  let cmdExec: string[];
+  let tempFile: string | undefined;
+  if (Deno.build.os === "windows") {
+    // writing to a file to exectute to avoid quoting issue with Deno
+    // https://github.com/quarto-dev/quarto-cli/issues/336
+    const lines = ["tlmgr", cmd, ...args];
+    tempFile = Deno.makeTempFileSync(
+      { prefix: "tlmgr-cmd", suffix: ".bat" },
+    );
+    Deno.writeTextFileSync(tempFile, lines.join(" ") + "\n");
```
cderv, line 298: I looked into `temp.ts` but was not sure any function should have been used here. Temp file is written in default temp dir. Maybe a special quarto-session dir should be used instead ? Not sure. 
I used default Deno temp file then.

src/command/render/latexmk/texlive.ts
```diff
@@ -285,14 +286,33 @@ function tlmgrCommand(
...
+        cmd: cmdExec,
         stdout: "piped",
         stderr: "piped",
       },
     );
+    result.finally(() => {
+      if (Deno.build.os === "windows" && tempFile) {
+        removeIfExists(tempFile);
+      }
+    });
```
cderv, line 315: At first I tried adding a `finally` in the `tryCatch` but I think that is the right way to go as `execProcess()` is a promise and we want to remove the file if success or if it fails.

src/command/render/latexmk/texlive.ts
cderv: Oh I was commenting that below. You were quick on the review.

I tried something like that considering the current try / catch there, but the await is outside as the this function is async for now.  If I add an await it won't be anymore right ? 

This is only to remove the tempfile we are creating, maybe it should be an argument, or maybe it should be written in a place that Quarto is already cleaning up. What the usual workflow with tempfile ? I have looked but not easy to see something useful for this small task.

jjallaire:
right, so you can do this:

```typescript
try {
    return execProcess(
      {
        cmd: cmdExec,
        stdout: "piped",
        stderr: "piped",
      },
    );
  } finally {
     if (tempFile) {
        removeIfExists(tempFile);
      }
  }
```

Note that you don't need to check for windows again in the finally since you only populate tempFile on windows.

The `catch` then `return Promise.reject()` also fully eats the exception details (you actually want it to be thrown and handled above).

cderv:
Ok I'll try this. I was under the impression the `try { } finally { } ` was deleting my temp file too quickly, not waiting for the execution. I'll try again this await / async and try / catch / finally way. thanks !

cderv:
@dragonstyle quoting all args works and don't mess the tests

src/command/render/latexmk/texlive.ts
cderv: We quote every argument now for windows

dragonstyle (commented):
Looks great, I made a couple suggestions for the actual implementation. Thanks for doing this!

src/command/render/latexmk/texlive.ts
cderv: ok I let it as is. thanks

src/command/render/latexmk/texlive.ts
```diff
@@ -280,21 +281,39 @@ async function verifyPackageInstalled(
...
 ) {
+  let cmdExec: string[];
+  let tempFile: string | undefined;
+  if (Deno.build.os === "windows") {
+    // quote args on Windows
+    args = args.map((a) => `"${a}"`);
+    // writing to a file to exectute to avoid quoting issue with Deno
+    // https://github.com/quarto-dev/quarto-cli/issues/336
+    const lines = ["tlmgr", cmd, ...args];
+    tempFile = Deno.makeTempFileSync(
```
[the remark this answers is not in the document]
cderv: I understand. I'll modify accordingly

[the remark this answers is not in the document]
dragonstyle: Totally agree re function argument, this is nicer!

You could make that `let tempFile` be a const if you move the creation of the temp file right there. Also means your finally wouldn't have to be conditional...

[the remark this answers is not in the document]
cderv: > You could make that let tempFile be a const if you move the creation of the temp file right there. Also means your finally wouldn't have to be conditional...

I wanted to do that but that is the part that I don't get 😅 
If I do that, I have warning from Typescript in VSCODE
image

and then error 
````
error: TS2304 [ERROR]: Cannot find name 'tempFile'.
      removeIfExists(tempFile);
                     ~~~~~~~~
    at file:///C:/Users/chris/DOCUME~1/DEV_R/QUARTO~2/src/command/render/latexmk/texlive.ts:306:22
````

That is why I did defined it outside of the try catch. 

am I missing a typescript special syntax ?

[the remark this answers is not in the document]
Comment (dragonstyle) [reply src/command/render/latexmk/texlive.ts]: I think you want:

```typescript
if (Deno.build.os === "windows") {
     // On Windows writing the command to a file to avoid quoting issue with Deno
     // https://github.com/quarto-dev/quarto-cli/issues/336
     const tempFile = Deno.makeTempFileSync(
         { prefix: "tlmgr-cmd", suffix: ".bat" },
       );
     try {
       // Quoting every argument for CMD
       args = args.map((a) => `"${a}"`);
       const lines = ["tlmgr", cmd, ...args];
```

that should make it visible to finally (and if creating the temp file throws, no need to clean it up so safe outside the try)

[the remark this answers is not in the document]
cderv: Oh I see. I thought it would need to be in the try, but you're right. No need to clean, and the other try in the other function  will catch that.

src/command/render/latexmk/texlive.ts
cderv: Done

src/command/render/latexmk/texlive.ts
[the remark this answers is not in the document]
cderv: I have put out the variables in [62ceb48](https://github.com/quarto-dev/quarto-cli/pull/340/commits/62ceb4856953d890ecce17f27df9e1b26eeb03e3)

dragonstyle:
This all looks very good to me (and I'm good with this if you're happy with it). 

One other random thought that I had was perhaps we should only the route the request through a tempfile in the event that one or more arg requires quoting (not 100% if this is well defined, but perhaps if it contains one of `& \ < > ^ |`? (from https://ss64.com/nt/syntax-esc.html).

cderv:
> One other random thought that I had was perhaps we should only the route the request through a tempfile in the event that one or more arg requires quoting 

I have thought of that at first, but I wasn't sure which list we should target. But we can definitely start with a list we think is relevant and adjust later. In **rmarkdown** we have this as a regex
````
# Regular expression representing characters likely to be considered special by
# the shell (require quoting/escaping)
.shell_chars_regex <- '[ <>()|\\:&;#?*\']'
````

I can tweak and do the addition for this if you prefer ?

dragonstyle:
That would be great, thank you. The rmarkdown list seems like a better place to start than the proposal I made in the previous comment.

cderv:
So I have integrated your feedback and also took the opportunity to reorganize a bit by making more generic. Not sure if it is necessary or a good idea, but it seemed like one 😅 

* There is now a `safeWindowsExec()` in `process.ts` that will execute a command (like `<program> <args>`) by writing to a temp file to run `CMD /c tempfile`
* `tlmgrCommand()` uses this function if it detects some special args to quote, and quote them before running this function. Otherwise, it will use the usual `execProcess` function. 

It seems clean enough to do it like that. 

If we want to generalize a bit more quoting and safe execution could in fact go into `execProcess()` directly. Would it be clever on the long run ? 
it seems like this is used for texlive only for now. `execProcess()` is used in several other places on Windows, but it does not seem like it is causing issues like with `tlmgr`.

dragonstyle (approved):

cderv:
Thanks for the help and review!

PR #363, opened 2022-02-24
Document Citation Support

dragonstyle:
- Add support for including `citation` in the front matter of documents. This allows you to specify the CSL-YAML that represents the citation information for this document (it will inherit the document metadata such as title when not specified). 

- Adds support for Google Scholar Metadata
Disabled by default, add `google-scholar` true to enable. Add additional metadata to configure properties. Generates references from bibliography.

- Adds support for citations in appendix
Includes a citation for the document in the appendix (bibtex and a formatted reference using the document CSL)

```yaml
---
title: "Test Document"
bibliography: references.bib
author: 
  - "Charles Teague"
  - J.J. Allaire
date: 1/23/2022
citation:
  type: article-journal
  container-title: The International Journal of Scientific Writing
  number: 7
  page: 12-15
  DOI: 10/223.2342432.2
google-scholar: true
---
```

Results in:
Screen Shot 2022-03-01 at 2 11 19 PM

depends upon #361


jjallaire:
I think it would be fine for the `citation-url` to trigger the appendix (rather than `appendix-attribution` being required).

PR #364, opened 2022-02-25
Fix hostname addr type convertation and update info output

lqhuang:
## Information for help:

- `quarto-cli` version: 0.9.39 (download from github release page)

## What I want to do:

I'm using `quarto` to create a project in my **remote server** and write my documents. Then I hope to use `quarto` to host and preview generated website. I try to specify a public address as hostname to preview arguments, which I could browse site from my **local laptop**. But I found it does't work and show the wrong url path.

```
$ quarto preview --port 4485 --host 0.0.0.0 --no-navigate --no-browser
Preparing to preview

Watching files for changes
Browse at http://localhost:4485/
```

## What I found:

After my debugging, I found `quarto` preview subcommod didn't [recognize host correctly](https://github.com/quarto-dev/quarto-cli/blob/12abf43f89b39ffa36404842ad09a04b8cba5e16/src/command/preview/cmd.ts#L162). And url output `printBrowsePreviewMessage` has no hostname passed.

## What I did:

1. Replace `parseInt` with `String` when parsing arguments
2. Add a new arg `host` to function `printBrowsePreviewMessage()` and update all codes called `printBrowsePreviewMessage()`
3. [Modified but reverted] Update `http://localhost:${options.port}/` to `http://${options.host}:${options.port}/`, but later I realized this is used for opening local browser, so I reverted back again.

That's all, thx

### DIFF
--- a/src/command/preview/cmd.ts
+++ b/src/command/preview/cmd.ts
@@ -159,7 +159,7 @@ export const previewCommand = new Command()
     }
     const hostPos = args.indexOf("--host");
     if (hostPos !== -1) {
-      options.host = parseInt(args[hostPos + 1]);
+      options.host = String(args[hostPos + 1]);
       args.splice(hostPos, 2);
     }
     const renderPos = args.indexOf("--render");
--- a/src/command/preview/preview.ts
+++ b/src/command/preview/preview.ts
@@ -141,13 +141,13 @@ export async function preview(
       relative(projectOutputDir(project), result.outputFile),
     )
     : "";
-  const url = `http://localhost:${options.port}/${initialPath}`;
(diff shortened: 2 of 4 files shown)


jjallaire:
Thank you! Could you send a signed copy of our contributor agreement as described here: https://quarto.org/about.html#pull-requests

lqhuang:
@jjallaire Ok, why not. I have sent you the email with signed copy. Please check. Thanks!

jjallaire:
FYI we had to revert the printing part of this as it broke Preview in RStudio: https://github.com/quarto-dev/quarto-cli/commit/751e82caf8a27ae1f5bca829263cf1f3adfa490a

lqhuang:
OK, no problem. Thank you for your work :)

PR #370, opened 2022-02-28
Update instructions in README

gvelasq:
Thank you for the excellent tools and documentation. For users not familiar with how Git and bash scripts work, it would be useful to give explicit instructions that the configure script should be rerun after each instance of `git pull` or `git pull --rebase`. I've added this explicit instruction and sample bash code in the README. Many thanks for considering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,12 @@ cd quarto-cli
 ./configure-macos.sh
 ```
 
-To update to the latest development version, just `git pull` from the local repo directory:
+To update to the latest development version, just `git pull` from the local repo directory and rerun the configure script:
 
 ``` bash
 cd quarto-cli
 git pull
+./configure-macos.sh
 ```
 
 ## License


jjallaire:
We actually actively detect when the configure script needs to be re-run and automatically re-run it (this happens the first time you try to execute a quarto command). You would _not_ have seen this behavior if you updated recently and hadn't pulled in a couple of months.

It may be that there are some scenarios where this fails, but we are definitely trying to make a simple `git pull` work. Are you seeing a situation where the reconfigure doesn't happen?

gvelasq:
Thank you for the explanation. My apologies, I think I figured out what happened, ultimately I think that this pull request can be closed since it's unlikely to be an issue with the README instructions on updating Quarto.

I was trying to help someone update their Quarto installation, and this is my best understanding of what happened. The user first installed Quarto on 2022-02-17 using the recommended steps. Then they deleted the local `quarto-cli` folder, but report that Quarto continued to work without issue. Today, since they no longer had a Git-tracked folder to run `git pull` from, they ran the following into the same location as the original deleted `quarto-cli` folder:

```bash
git clone https://github.com/quarto-dev/quarto-cli
cd quarto-cli
git pull
```

`git pull` did not bring in new changes immediately after `git clone`, which cloned the latest development version. The Quarto installation worked as before and didn't appear to reconfigure when rerunning a Quarto command, because it did not have expected new features implemented since 2022-02-17. After reinstalling by running `./configure-macos.sh`, Quarto did have the expected new features. My interpretation is that deleting the original Git-tracked `quarto-cli` folder likely severed the link with the original installation and reconfigure detection, so in a way the user reinstalled Quarto rather than updating it with `git pull`.

jjallaire:
Okay, great. That all makes sense! If you do see any situations where a reconfigure isn't prompted for do let us know and we'll patch it up. When we started off we required manual reconfigure each time but we found that even experienced git/bash users would forget this step, much better to automate!

dragonstyle:
Might be worth checking whether the user has ever run the Quarto installer- from your description above (great detail, thank you!) I think there is a reasonable chance that the user ran an installer after cloning the repo. This would reset the global `quarto` symlink to point at the installed version (rather than the locally cloned version). When the user re-cloned the repo the symlink remained pointed to the installed version (rather than the cloned version) until `configure-*.sh`  was run, at which time, the symlink was reset to point to the cloned version... That's a reasonable theory, I think...

You can know for sure which version of quarto you're using by running `quarto --version` which will always return `99.9.9` for a locally cloned version...

gvelasq:
Thanks @dragonstyle for following up. After further discussion your hypothesis seems entirely possible. The user confirmed having installed Quarto using the command line on one machine, and an installer on another machine, and is not clear on which machine was the one that ran an installer. Many thanks to both of you for your thoughtful and rapid answers! Glad to see this project gaining momentum.

Issue #375, opened 2022-03-01
behavior of `render` and `preview` of `revealjs` files that are listed in an index html file


juliantao:
When a listed (child?) page is re-rendered, the listing (parent?) page is also re-rendered automatically.  It seems that this is necessary since the `data-listing-file-modified-sort` field needs to be updated on the parent page. 

However, I encountered a problem when I listed a number of `revealjs` files on an index html page. When I re-render or preview an updated `revealjs` file, the index page is also re-rendered, but into a `revealjs` format. Right now, I need to re-render the index page again to recover it back to a conventional `html` format.

dragonstyle:
This is a great find- thank you for reporting it! It should be fixed now as of:

https://github.com/quarto-dev/quarto-cli/releases/tag/v0.9.57

juliantao:
Thank you! I can confirm that it works as expected.

PR #377, opened 2022-03-02
Supplemental Files Can Filter Formats

dragonstyle:
When a supplemental file is contributed to a render, it can now includes foramts that go along with it (so that it won’t be forced to render in an unsupported format). For example, when a listing gets added to the render of a revealjs document, it still needs to render as an html document rather than as a revealjs presentation.

https://github.com/quarto-dev/quarto-cli/issues/375


jjallaire:
LGTM! (and I really did review every line :-)

PR #386, opened 2022-03-05
Add content conditional on the output formats

jmbuhr:
Discussion: see #373

This PR adds a a lua-filter which allows the following syntax:

````
:::web-only
Hello from the digital version.
:::

:::print-only
Hello from the print version.
:::
````
to include content for only a specific output format. This comes in handy when e.g. embedding video directly in html / web formats, which could then be replaced by a link or description in a print format.

The filter runs before other quarto filters so that we can use e.g. shortcodes in conditional blocks:

````
:::web-only
Hello from the digital version.
You are reading {{< meta title >}}.
:::
````

### DIFF
--- a/src/resources/filters/quarto-pre/conditional-content.lua
+++ b/src/resources/filters/quarto-pre/conditional-content.lua
@@ -1,0 +1,29 @@
+-- conditional-content.lua
+--
+
+local function matchAny(str, patternList)
+    for _, pattern in ipairs(patternList) do
+        local w = str:match(pattern)
+        if w then return w end
+    end
+end
+
+local webFormats = {'html','revealjs'}
+local printFormats = {'latex','docx','pdf','opendocument','odt','beamer'}
+
+function conditionalContent()
+  return {
+
+    -- discard the content of a div if the output format doesn't match
+    Div = function(div)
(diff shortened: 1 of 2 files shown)


jmbuhr:
Looks like some of my assumptions about the order of filters were wrong, So right now this fails in some cases.

jjallaire:
Thanks for this! There would be a few more things to consider in terms of formats to make this really work well. For example, here are the aliases we use to define various classes of format: https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/schema/format-aliases.yml

I also might choose to use an attribute rather than a class (as an attribute would have more of the flavor of a directive vs. styling which is what class invokes). 

The other thing to consider is computations -- I think that this might also want to restrict what blocks are actually run through knitr, Jupyter (e.g. in the case where one block produces a PDF and one produces a PNG).

All that to say there is more than meets the eye here! We are going to work on an include feature soon which will run as a global preprocessor (so you can do computational includes). At that point I think we should implement this feature as well.

jmbuhr:
Closing this for cleanup, after the addition of https://quarto.org/docs/authoring/conditional.html

PR #464, opened 2022-03-25
[revealjs] do not move speaker notes as other asides

### LINKED ISSUE #459: `revealjs` speaker notes, aside notes and footnotes
No description provided.

cderv:
Fix #459

Pandoc support fenced divs of class notes in revealjs to create  speaker notes 
https://pandoc.org/MANUAL.html#speaker-notes

They are translated to HTML as `<aside class="notes">` for revealjs to pick them up (https://revealjs.com/speaker-view/), so we need to ignore them in our post process and not move those aside. 

I am suprised that we want to move more than `div.aside` element created using 
````markdown
::: aside
Some additional commentary of more peripheral interest.
:::
````

but maybe I am missing something, so fix here is to just ignore `aside` with class `notes`. 

I'll try to add test for this example so that we check it does not break in the future

### DIFF
--- a/src/format/reveal/format-reveal.ts
+++ b/src/format/reveal/format-reveal.ts
@@ -583,7 +583,7 @@ function revealHtmlPostprocessor(format: Format) {
     const slides = doc.querySelectorAll("section.slide");
     for (const slide of slides) {
       const slideEl = slide as Element;
-      const asides = slideEl.querySelectorAll("aside");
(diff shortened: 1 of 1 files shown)


jjallaire:
Yeah, I think I was just being "complete" by moving the `<aside>` tag as well. Ideally since there is an HTML tag for just this purpose we could _only_ support `<aside>` but the div syntax is definitely more natural to write in markdown. I think we are okay to support both with the change you made.

PR #465, opened 2022-03-25
[revealjs] tests for aside, footnotes and speaker notes

cderv:
This complement previous PR #464 by adding some smoke tests. 

First one I wrote so I hope I got right. I don't know if I am testing to precise selectors maybe.

@dragonstyle your the test guy on the team I believe, so I'll ask you for review. 

It passes on my side, hopefully it will in CI.

### DIFF
--- a/tests/smoke/render/render-reveal.test.ts
+++ b/tests/smoke/render/render-reveal.test.ts
@@ -5,7 +5,25 @@
 *
 */
 
-import { docs } from "../../utils.ts";
+import { docs, outputForInput  } from "../../utils.ts";
+import { ensureHtmlElements } from "../../verify.ts";
 import { testRender } from "./render.ts";
 
+// demo file renders ok
 testRender(docs("reveal/index.qmd"), "revealjs", false);
+
+// aside and footnotes correctly moved with speaker notes ignored
+const input = docs("reveal/aside-footnotes.qmd");
+const output = outputForInput(input, "revealjs");
+testRender(input, "revealjs", false, [
+  ensureHtmlElements(output.outputPath, [
+    // speaker notes are left not moved
+    "section#slide-with-speaker-notes > aside.notes:last-child",
+    // div of class aside are moved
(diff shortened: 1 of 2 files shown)


tests/smoke/render/render-reveal.test.ts
```diff
@@ -5,7 +5,25 @@
...
+testRender(input, "revealjs", false, [
+  ensureHtmlElements(output.outputPath, [
+    // speaker notes are left not moved
+    "section#slide-with-speaker-notes > aside.notes:last-child",
+    // div of class aside are moved
+    "section#slide-with-footnotes > aside:last-child > div > p",
+    // footnotes are put in aside
+    "section#slide-with-footnotes > aside:last-child > ol.aside-footnotes > li",
+    // footnotes back are removed
+    "section#slide-with-footnotes > aside:last-child > ol.aside-footnotes > li > :not(.footnote-back)",
```
dragonstyle, line 27: Just FYI if you're trying to validate the absence of elements in the DOM (for example, there are no `.footnote-back` elements), you can also pass an array of no-match selectors after the match selectors (the verification will check that those selectors do not return elements). 

```
 ensureHtmlElements(output.outputPath, [
     // speaker notes are left not moved
     "section#slide-with-speaker-notes > aside.notes:last-child",
     // div of class aside are moved
     "section#slide-with-footnotes > aside:last-child > div > p",
     // footnotes are put in aside
     "section#slide-with-footnotes > aside:last-child > ol.aside-footnotes > li"],
     [     
       // footnotes back are removed
       "section#slide-with-footnotes > aside:last-child > ol.aside-footnotes > li > .footnote-back"])
```

dragonstyle:
> I don't know if I am testing to precise selectors maybe.

I don't think we have a systematic approach to selectors, though in general I've tried to be minimal in my selectors just to make the tests less fragile, especially regarding the complete structure of the document. From looking at your selectors, they seem like they express pretty essential structures, so seem ok...

PR #545, opened 2022-04-06
Move binary dependencies into subfolder

dragonstyle:
Moves binary dependencies into a `tools` subdirectory in order to prevent them from being on the path (in windows).


jjallaire:
Reviewed and all looks good to me! This is a great thing to get behind us.

PR #563, opened 2022-04-07
Update _language-de.yml

mchlbckr:
Adjust translations to German

PR #565, opened 2022-04-07
Revert "Update _language-de.yml"

jjallaire:
Reverts quarto-dev/quarto-cli#563


jjallaire:
Actually, since I'm not a German speaker I sort of automatically assumed that these edits are strictly better. Then I recalled that our translations were in fact done by a native speaker (@skeydan). I'm going to re-open the PR and ask for review.

PR #566, opened 2022-04-07
Update _language-de.yml

mchlbckr:


skeydan:
@jjallaire @mchlbckr  I agree these sound more idiomatic. I guess I don't really visit many German-language websites ...

jjallaire:
Thanks @skeydan !

PR #576, opened 2022-04-07
add `iso` alias for `date-format`

cderv:
@dragonstyle Following discussion with @sellorm , here is a PR to add `date-format: iso` as an alias for `date-format: YYYY-MM-DD` 

````markdown
---
title: "Test"
date: 2022-04-10
format: html
---

# test

````

image

````markdown
---
title: "Test"
date: 2022-04-10
date-format: iso
format: html
---

# test
````
image

* Do you need tests for that ? I don't think there is any for the date features. Could add some if you want. 
* YAML validation does not have yet support for `date-format` it seems. There are some descriptions, but no value for autocompletion yet. We could probably add some ?
* Another option would be to use [`toISOString()`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) method but it will return the full isodatetime format. I think using `YYYY-MM-DD` is easier and better. 
* Seems like this is the only place to tweak - when (if) merged, I'll update the doc site.

### DIFF
--- a/src/core/date.ts
+++ b/src/core/date.ts
@@ -19,7 +19,7 @@ export const kToday = "today";
 export const rSysDate = "`r Sys.time()`";
 export const kNow = "now";
 
-export type DateFormat = "full" | "long" | "medium" | "short" | string;
+export type DateFormat = "full" | "long" | "medium" | "short" | "iso" | string;
 export type TimeFormat = "full" | "long" | "medium" | "short";
 
 export function today(): Date {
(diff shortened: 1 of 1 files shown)


dragonstyle:
This looks great - I think your choice re: ISO dates versus full date and time is smart, given the common use. If users really need it, we can always add `iso-datetime` in the future if need be...

I don't think tests are strictly necessary, but there is a `render-date.test.ts` that you could piggyback upon if you'd like (it currently only checked the today / last-modified date aliases, but don't test formatting at all...

PR #579, opened 2022-04-07
de translations for environments

jjallaire:
@mchlbckr and @skeydan Could you review?


src/resources/language/_language-de.yml
```diff
@@ -64,6 +64,10 @@ crossref-lof-title: "Abbildungsverzeichnis"
 crossref-lot-title: "Tabellenverzeichnis"
 crossref-lol-title: "Listingverzeichnis"
 
+environment-proof-title: "Nachweisen"
```
skeydan, line 67: I think this should be "Beweis"

src/resources/language/_language-de.yml
```diff
@@ -64,6 +64,10 @@ crossref-lof-title: "Abbildungsverzeichnis"
...
+environment-remark-title: "Bemerkung"
```
skeydan, line 68: I think "Anmerkung" would be a better fit

src/resources/language/_language-de.yml
mchlbckr: Yes indeed!

src/resources/language/_language-de.yml
mchlbckr: "Bemerkung" is perfect!

src/resources/language/_language-de.yml
```diff
@@ -64,6 +64,10 @@ crossref-lof-title: "Abbildungsverzeichnis"
...
+environment-solution-title: "Lösung"
+
 listing-page-order-by: "Sortieren nach"
 listing-page-order-by-default: "Voreinstellung"
 listing-page-order-by-date-asc: "Älteste"
```
mchlbckr, line 73: Instead of "Älteste" I would say "Datum (aufsteigend)"

src/resources/language/_language-de.yml
```diff
@@ -64,13 +64,13 @@ crossref-lof-title: "Abbildungsverzeichnis"
 crossref-lot-title: "Tabellenverzeichnis"
 crossref-lol-title: "Listingverzeichnis"
 
-environment-proof-title: "Nachweisen"
-environment-remark-title: "Bemerkung"
+environment-proof-title: "Beweis"
+environment-remark-title: "Anmerkung"
 environment-solution-title: "Lösung"
```
mchlbckr, line 69: I would keep "Bemerkung"

jjallaire: Okay I made that change. @skeydan and @mchlbckr if you continue discussing I'll track the discussion and change to whatever consensus emerges from the discussion.

skeydan:
I would translate "remark" as "Bemerkung" if used in some - preferredly, verbal - context; as in: "Then, she made a remark that ...".

If used standalone, however, and in written text, I would use "Anmerkung". A prototypical example would be a footnote starting : "Anmerkung: ..." 

The way I'm imagining this being used in quarto would be more like the second one - correct me if I'm wrong.
That said, "Bemerkung" seems acceptable as well.

jjallaire:
Yes, this is in the context of a LaTeX "Remark" environment (which is much like a footnote). Based on this I'll change it back to Anmerkung

mchlbckr:
Yes, with that context, this seems to be the best translation!

PR #597, opened 2022-04-09
Fix #593: recognize position specifier in latexTablePattern

### LINKED ISSUE #593: In TeX, table caption is inserted between \begin{table} and [placement options]
This relates to https://github.com/hughjonesd/huxtable/issues/223. Huxtable produces TeX tables programmatically from within R. The table starts like 

```
\begin{table}[ht]
...
```

In a Quarto code block with a label and/or caption set via a chunk option, a TeX caption is produced:

```
\begin{table}
\caption{\label{tbl-iris}Iris Data }
[ht]
```

The problem is that the `\caption` statement is being placed between `\begin{table}` and the table placement options `[ht]`.
Instead, the caption should be placed after the `[ht]`.

Test case:

    ---
    title: "test"
    format: pdf
    editor: visual
    ---

    ```{r}
    #| include: false
    library(huxtable)
    ```

    @tbl-iris shows up and can be cross-referenced, but it has an "\[ht\]" in front of it.

    ```{r}
    #| label: tbl-iris
    #| tbl-cap: "Iris Data"

    hux(head(iris, n = 1))
    ```

hughjonesd:
This is just a guess!

You may also need to make a similar change to the LongTablePattern. CTAN documentation says this has some options too, either [c], [l] or [r].

### DIFF
--- a/src/resources/filters/quarto-pre/table-captions.lua
+++ b/src/resources/filters/quarto-pre/table-captions.lua
@@ -4,7 +4,7 @@
 kTblCap = "tbl-cap"
 kTblSubCap = "tbl-subcap"
 
-local latexTablePattern = "(\\begin{table}(\\[.*?\\])?)(.*)(\\end{table})"
+local latexTablePattern = "(\\begin{table}(\[.*?\])?)(.*)(\\end{table})"
 local latexLongtablePattern = "(\\begin{longtable})(.*)(\\end{longtable})"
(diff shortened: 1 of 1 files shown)


jjallaire:
Thanks for this! Believe it or not those aren't regexes but "lua patterns". As a result the cleanest way to look for this is a separate pattern (Lua patterns don't have the same conditional matching syntax as regexes). I've don that here: https://github.com/quarto-dev/quarto-cli/commit/0a47ba0648ba59ee09966a33414275e83dc7e6a2

PR #614, opened 2022-04-12
Add a11y syntax highlighting themes

batpigandme:
* Updates one color in `arrow-light` to pass WCAG AA contrast ratio
* Adds KDE ports of `a11y-dark` and `a11y-light` adapted from https://github.com/ericwbailey/a11y-syntax-highlighting


cderv:
Thanks @batpigandme , I updated the `arrow.theme` in **distill** and **rmarkdown** too.

PR #663, opened 2022-04-16
Improve Listings

dragonstyle:
- Fix rendering issue with markdown (html blocks in default seem to cure it)
- Fix glob resolution so we always use yaml or inputs
- fix up column matching for title banner


dragonstyle:
(For testing / bundling)

PR #668, opened 2022-04-17
Japanese translation brush-up

eitsupi:
Improve the changes made in #658 by @kozo2.

1. I don't think that the conditions for quoting a copyrighted work are generally referred to as "再利用".
  I think the general expression is "二次利用". (e.g. <https://creativecommons.jp/faq/#i14>)
  However, judging from the context, I think that "ライセンス" is generally the most appropriate.
2. I think that it is common to indicate with "()" when indicating descending or ascending order like "日付（昇順）".
3. I think "ファイル変更" is redundant to indicate the modification date of a file.

And, I correct what I consider inappropriate when I apply the Japanese language to some files I have created.


src/resources/language/_language-ja.yml
```diff
@@ -70,16 +70,16 @@ environment-solution-title: "解答"
...
+listing-page-order-by-date-desc: "日付（降順）"
+listing-page-order-by-number-desc: "ページ番号（昇順）"
+listing-page-order-by-number-asc: "ページ番号（昇順）"
 listing-page-field-date: "日付"
 listing-page-field-title: "題名"
 listing-page-field-description: "説明"
 listing-page-field-author: "作者"
 listing-page-field-filename: "ファイル名"
-listing-page-field-filemodified: "ファイル変更"
+listing-page-field-filemodified: "更新"
```
atusy, line 82: I would suggest `更新日`.
In English, `Modified` is preferred maybe because it is fairly shorter than `Modified data`.
OTH, adding a letter in Japanese is small enough.
Note that Spanish version is more like `更新日` (`Fecha de modificación`).

eitsupi: Thanks for your suggestion!
That make sense, this seems a "date" type.

https://github.com/quarto-dev/quarto-cli/blob/e8dd7e1448cbce1c52c04274222a8e731780bc6c/src/project/types/website/listing/website-listing-read.ts#L127

eitsupi:
I've waited a week and there seems to be no other feedback so I think we can merge this.

PR #707, opened 2022-04-21
Use curl to get deno instead of wget

mbaynton:
This is selfishly motivated by the need to run this configuration script
in an environment outside our immediate control whose wget is not linked
against a TLS library, but whose curl is.

### DIFF
--- a/configure-linux.sh
+++ b/configure-linux.sh
@@ -2,6 +2,17 @@
 set -e
 source configuration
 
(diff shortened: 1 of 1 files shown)


jjallaire:
I don’t think curl is installed by default on many Linux systems so at a minimum this needs to check for curl first and fall back to wget

mbaynton:
@jjallaire updated to use curl if present and advertising https support, otherwise fallback to wget

Issue #724, opened 2022-04-24
Fig captions not rendering in pdf output


jazzbuck:
Thanks for all the work on quarto - I've been really enjoying the flow of it.

I'm using quarto to render my jupyter notebook to pdf file.  

I expect to see standard behaviour of cross-referencing between markdown cells to cells containing my tables. This works when rendering to html, but fails when rendering to pdf.

Example cells:

Cell 1:
```markdown
In @tbl-summary-gen we see...
```

```python
#| echo: false
#| label: tbl-summary-gen
#| tbl-cap: "summary of hypothesis generation datasets"

df = pd.DataFrame(
    {
    "a":["2,3,4,5"],
    "b":["3,4,5,6"],
    }
)
df
```

jjallaire:
The problem is likely that when you executed the notebook in Jupyter (or VS Code) that HTML output (rather than markdown) was produced. Then, when rendering to PDF you get LaTeX attempting to process the HTML. If you render with execution like this:

```bash
quarto render notebook.ipynb --to pdf --execute
```

Then you'll get the expected output.

PR #735, opened 2022-04-26
support raw html tables in all formats

jjallaire:
If we have a raw html table in a format that doesn't handle raw_html then have pandoc parse the table into a proper AST table block.

The motivation for this is to resolve this issue: https://github.com/quarto-dev/quarto-cli/issues/724. 

ipynb files are not executed by default when rendered (as the output is already embedded and it may not even be possible to re-execute given the rendering environment, availability of dependencies, etc.). This means that we need to rely on the representation created in the interactive notebook. For tables, this will often be HTML. Without attempting to interpret the HTML, this makes ipynb files with tables un-renderable to non-HTML formats (e.g. pdf, docx, ec.)

This PR has two main changes:

1) When marshaling from` ipynb` to `md` prior to rendering, include raw HTML output if it includes a `<table>` tag.

2) For formats that can't render HTML tables natively (e.g. pdf, docx, rtf, odt, etc.), read raw HTML tables into the pandoc AST (which allows them to be rendered in their native format)

### DIFF
--- a/src/config/format.ts
+++ b/src/config/format.ts
@@ -119,7 +119,16 @@ export function isIpynbOutput(format: FormatPandoc) {
 
 export function isMarkdownOutput(
   format: FormatPandoc,
-  flavors = ["markdown", "gfm", "commonmark"],
+  flavors = [
+    "markdown",
+    "markdown_github",
+    "markdown_mmd",
+    "markdown_phpextra",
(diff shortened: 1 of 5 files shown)


alexander-newton:
Thanks for tackling this so swiftly.

PR #783, opened 2022-05-02
Improve support of specific knitr engines for source output 

cderv:
it should not be pass through so that it does not end up into the div when pass as a knitr option. 

This option is new from knitr 1.38 (https://github.com/yihui/knitr/blob/master/NEWS.md#changes-in-knitr-version-138)

I don't think this option `lang` has any effect in Quarto yet. So we have a difference in behavior between **knitr** in R Markdown and in Quarto. 

With **knitr**, `lang` is used in various place to change the language that will be applied as class to the code chunk in Markdown. Default is the engine. Sometimes it needs to be tweaked.

Example: 

````markdown
```{verbatim, lang = "js"}
console.log('test')
```
````

This is used in **knitr** to output a code chunk of class `js`

````markdown
```{bash, lang = "default"}
pwd
```
````

This is used to execute a chunk as bash but highlight as a default block. 

So this PR only makes sure that the option is correctly identified a knitr option. I'll make another one (or just continue in this one if you prefer) to make the recent engine (`verbatim`, `embed`, `comment`, ...) works correctly. They are not really working as expected I believe (mainly because we overwrite the source hook I think).

### DIFF
--- a/src/resources/rmd/hooks.R
+++ b/src/resources/rmd/hooks.R
@@ -339,7 +339,7 @@ knitr_hooks <- function(format, resourceDir, handledLanguages) {
     x <- knitr:::one_string(c('', x))
     
     # leave verbatim alone
-    if (identical(options[["engine"]], "verbatim")) {
+    if (options[["engine"]] %in% c("verbatim", "embed")) {
       return(paste0('\n\n````', options[["lang"]] %||% 'default', x, '\n````', '\n\n'))
     }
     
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cderv** [comment]
I added support for 

````markdown
```{embed, file = "special-highlight.lua"}
```
````

in addition of the `verbatim` engine. 

There is still the `cat` engine that has a specific behavior depending on the source hook (that quarto replace). I am still onto this. 

@yihui maybe we need to reorganize and refactor in **knitr** so that we could maybe not rewrite things too much and so that Quarto can directly use some knitr function.

yihui (approved):
> @yihui maybe we need to reorganize and refactor in **knitr** so that we could maybe not rewrite things too much and so that Quarto can directly use some knitr function.

Sure! We should definitely do that.

PR #800, opened 2022-05-04
Bring PDF viz support to Julia

jeremiahpslewis:
Enable vector plot output for quarto PDF docs for Julia libraries CairoMakie.jl and Plots.jl

### DIFF
--- a/src/resources/jupyter/lang/julia/setup.jl
+++ b/src/resources/jupyter/lang/julia/setup.jl
@@ -19,6 +19,7 @@ fig_dpi = {3}
 if fig_format == :retina
   fig_format = :svg
 elseif fig_format == :pdf
+  fig_dpi = 96
   # Enable PDF support for IJulia
   IJulia.register_mime(MIME("application/pdf"))
 end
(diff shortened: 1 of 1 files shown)


jeremiahpslewis:
- Tested locally, looks reasonable for Plots and CairoMakie.
- Shifted from `using Plots` to `import Plots` to handle namespace conflicts, user can add a `using Plots` as needed

src/resources/jupyter/lang/julia/setup.jl
```diff
@@ -10,30 +10,49 @@ import IJulia
...
+fig_width = {0}
+fig_height = {1}
+fig_format = :{2}
+fig_dpi = {3}
+
+# no retina format type, use svg for high quality type/marks
+if fig_format == :retina
+  fig_format = :svg
+elseif fig_format == :pdf
+  fig_dpi = 96
```
jjallaire, line 22: Do we want the DPI to be left alone here? (we default it to 300 for PDF, we defaulted it back to 96 only b/c we were re-routing to PNG)

jjallaire: I see now that we need the dpi to be 96 for the pixel calculation to be correct

jjallaire:
I'll go ahead and merge this notwithstanding the comments about dpi (I'll assume you'll come back with a different PR if you think some adjustment in how we handle dpi for PDF is warranted)

jjallaire:
Hi there @jeremiahpslewis, posting back here because I'm giving a talk on Quarto at JuliaCon next month and was hoping to find some good examples of using Julia w/ Quarto. If you are able to share any of your own (or are aware of others) please LMK over on this discussion thread: https://github.com/quarto-dev/quarto-cli/discussions/1261. Thanks!

PR #837, opened 2022-05-09
Add tweet shortcodes

### LINKED ISSUE #765: Shortcodes for embedding tweets
On https://quarto.org/docs/authoring/shortcodes.html I see that tweet shortcodes are on the todo list. I just want to say that I will use them as soon as they are available!

If you can point me in the right direction I can also start on a PR for this.

alexpghayes:
Closes #765.


jjallaire:
Thank you! This looks really cool. We've had some internal discussion about a tweet shortcode and a few other things to consider are:

1) Whether we can still build a nice looking tweet for print media (e.g. by composing return values from the endpoint into something that looks like a tweet)

2) Whether we should do some caching of the http response so that the render doesn't always have to hit the internet.

My suggestion is that we keep this shortcode external for now as we'll want to consider more general ways of doing the above before proceeding. We are going to create an "extensions" mechanism soon that will include shortcodes, and perhaps once your shortcode is packaged as an extension we can point to it as an example.

alexpghayes:
Sounds good! Excited to give the extensions mechanism a try!

dragonstyle:
We have the extensions mechanism ready and documented here:

https://quarto.org/docs/extensions/shortcodes.html#overview

Would appreciate if you have the time to take a look!

Issue #817, opened 2022-05-06
In PDF, long callouts overflow the page.


postylem:
Currently it seems that (at least by default), a callout with a lot of text in it will end up overflowing the page in PDF.

That is, 
```markdown
---
title: A title
format: pdf
---

Here is an example of a long box:

:::{.callout}
...
(many lines of text here)
...
::: 
```
will lead to the contents of the callout overflowing past the bottom of the page.

Is there a way to configure this not to happen?  

If not, since callouts are rendered in latex with the `tcolorbox` package, I imagine it would be easy to make these automatically page-break, by adding, just passing the `breakable` option when making the box

```latex
\begin{tcolorbox}[breakable]
...
\end{tcolorbox}
```
perhaps here: https://github.com/quarto-dev/quarto-cli/blob/9167b7bbc5fe5ed95357459bb80288fa0ae3fa0b/src/resources/filters/quarto-pre/callout.lua#L323

Is there a reason not to do this?

jjallaire:
It seems like we should indeed do this. We'll experiment and hopefully have a fix for this soon! Thanks for reporting and for the precise diagnosis of the solution.

postylem:
Actually, I see that setting `breakable` is already used by default here:  https://github.com/quarto-dev/quarto-cli/blob/5cfe5046f1df1e55c611056f5ca5d51273b52bf0/src/resources/filters/layout/meta.lua#L44

So, I think it should be quite simple to extend this to be the default in general.  

I made a PR https://github.com/quarto-dev/quarto-cli/pull/886 with this simple change.

PR #886, opened 2022-05-15
Allow pagebreaks in callout boxes in PDF

postylem:
Following changes to the options passed to `tcolorbox`:
- Add `breakable` option so that boxes will break to span across pagebreaks
- Change `standard jigsaw` -> `enhanced jigsaw`, so that partial boxes do not close.

To address issue https://github.com/quarto-dev/quarto-cli/issues/817


jjallaire:
@dragonstyle could you review this?

dragonstyle (approved):

PR #908, opened 2022-05-17
Correctly query registry key on Windows for finding chrome

### LINKED ISSUE #868: Chrome is not found when trying to render document with Graph and it errors
I tried example at https://quarto.org/docs/authoring/diagrams.html

````markdown
---
title: "My simple title"
format: html
---

# Head

```{mermaid}
flowchart LR
  A[Hard edge] --> B(Round edge)
  B --> C{Decision}
  C --> D[Result one]
  C --> E[Result two]
```
````

and I get this when `quarto render`

````
WARNING: Capturing of embedded web content disabled (chromium not installed)
ERROR: TypeError: Cannot read properties of undefined (reading '0')

TypeError: Cannot read properties of undefined (reading '0')
    at Object.cell (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/core/handlers/mermaid.ts:84:69)
    at async Promise.all (index 0)
    at async Object.document (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/core/handlers/base.ts:367:20)
    at async handleLanguageCells (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/core/handlers/base.ts:330:34)
    at async renderFiles (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-files.ts:384:39)
    at async render (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-shared.ts:91:18)
    at async Command.fn (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/cmd.ts:183:26)
    at async Command.execute (https://deno.land/x/cliffy@v0.19.3/command/command.ts:997:7)
    at async Command.parse (https://deno.land/x/cliffy@v0.19.3/command/command.ts:918:16)
    at async Command.parse (https://deno.land/x/cliffy@v0.19.3/command/command.ts:870:16)
````

I am using last dev version (946b31a703bda35218c26323d1c2349f0cca08a5) on windows 10

cderv:
Fix #868

There was 2 issues: 
1. Quoting issue with `Deno.run()` as we encountered with `tlmgr` in the past
2. Localized output value of `reg.exe query` which made our regex matching failed

For the first, I refactored the fix for `tlmgr` by making the wrapper `safeWindowsExec` available in `windows.ts` so that we can reuse it. Basically the trick is to quote the arguments that needs it (`requireQuoting()`), then write the command to execute into a `.bat` file using `Deno.run()`

For the second, I understand we don't need to search for `Default` in the returned key when `/ve` is used on the query as only the default will be returned. So when `(Default)` keyname issued, we ignore it in the regex. 

This solve the issue on Windows. Hopefully, it is not breaking anything elsewhere. 

Side notes: 

I took some time before submitting the PR as we discussed with @dragonstyle making the quoting on Windows more generic as this is a known issue with `Deno.run()`. Idea was to make `execProcess()` supports this. However, it is a bit more complex that I though and this seems to lead to change the way `execProcess()` is called, and we use that in a lot of places. 

That is only the second time we are encountering this and we now have a pattern using `requireQuoting()` and then `safeExecWindows()` when needed. Maybe that is enough for now ? 
Anyway, I could still continue this rewrite, but it would probably be good in another PR.

### DIFF
--- a/src/command/render/latexmk/texlive.ts
+++ b/src/command/render/latexmk/texlive.ts
@@ -9,10 +9,11 @@ import { existsSync } from "fs/mod.ts";
 
 import * as ld from "../../../core/lodash.ts";
 
-import { execProcess, safeWindowsExec } from "../../../core/process.ts";
+import { execProcess } from "../../../core/process.ts";
 import { kLatexHeaderMessageOptions } from "./types.ts";
 import { lines } from "../../../core/text.ts";
 import { tinyTexInstallDir } from "../../tools/tools/tinytex-info.ts";
+import { requireQuoting, safeWindowsExec } from "../../../core/windows.ts";
 
 // Determines whether TexLive is installed and callable on this system
 export async function hasTexLive(): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


jjallaire:
Okay, agreed that we should keep this fix scoped rather than rewriting `execProcess()` right now. Our hope is that in advance of the v1.0 release we will substantially improve test coverage which will allows us to more confidently refactor lower level routines like `execProcess()`

PR #947, opened 2022-05-22
use host in preview and serve

rcannood:
The documentation website of [Viash](https://viash.io) urgently needs a makeover, and I was looking into using quarto to create it. However, I ran into the issue of the `--host` parameter not being passed along inside of the code. This means I can't access the web server from inside a container.

## Demonstration of the issue

I set up a new website. For this example, I'll be using Viash to make the Docker container build a little bit easier:

```bash
quarto create-project mysite --type website
cd mysite

# download viash
mkdir bin
wget https://github.com/viash-io/viash/releases/download/0.5.11/viash -O bin/viash
chmod +x viash
```
I created a [Viash component](https://github.com/rcannood/quarto-cli/blob/viash_configs/src/viash_configs/quarto_preview/config.vsh.yaml) (which is a script + a description of the CLI + a Docker container spec) to make running quarto preview easier.

When I run the Viash component, I get:

```bash
$ bin/viash run https://raw.githubusercontent.com/rcannood/quarto-cli/viash_configs/src/viash_configs/quarto_preview/config.vsh.yaml
```

    [notice] Checking if Docker image is available at 'quarto_preview:latest'
    [warning] Could not pull from 'quarto_preview:latest'. Docker image doesn't exist or is not accessible.
    [notice] Building container 'quarto_preview:latest' with Dockerfile
    
    Preparing to preview
    
    Watching files for changes
    Browse at http://localhost:4444/

When I try to browse `http://localhost:4444/`, it fails because the web server is bound to 127.0.0.1.

## Demonstration of the solution

This [Viash component](https://github.com/rcannood/quarto-cli/blob/viash_configs/src/viash_configs/quarto_preview_fix/config.vsh.yaml) uses my fork of quarto.

This time, I will build the component explicitly into an executable. The Docker container build will take a bit longer because it's building quarto from source.

```bash
viash build https://raw.githubusercontent.com/rcannood/quarto-cli/viash_configs/src/viash_configs/quarto_preview_fixed/config.vsh.yaml -o bin --setup cachedbuild
```
    [notice] Building container 'quarto_preview_fix:latest' with Dockerfile

When I run the Viash component, I get:
```bash
# need to clean up previous preview first
bin/quarto_preview_fix
```

    Terminating existing preview server....DONE

```bash
bin/quarto_preview_fix
```
    Preparing to preview
    
    Watching files for changes
    Browse at http://172.17.0.2:4444/

Browsing this link now works.

Sidenote: The executable generated by Viash now works as a standalone executable which uses a Docker container in the backend.

### DIFF
--- a/src/command/preview/preview.ts
+++ b/src/command/preview/preview.ts
@@ -178,7 +178,7 @@ export async function preview(
       relative(projectOutputDir(project), result.outputFile),
     )
     : "";
-  const url = `http://localhost:${options.port}/${initialPath}`;
+  const url = `http://${options.host}:${options.port}/${initialPath}`;
   if (
     options.browser &&
     !isRStudioServer() && !isRStudioWorkbench() && !isJupyterHubServer() &&
@@ -188,7 +188,7 @@ export async function preview(
   }
 
   // print status
-  await printBrowsePreviewMessage(options.port!, initialPath);
+  await printBrowsePreviewMessage(options.host!, options.port!, initialPath);
 
   // watch for src changes in dev mode
   monitorQuartoSrcChanges(stopServer);
--- a/src/command/render/render-shared.ts
+++ b/src/command/render/render-shared.ts
@@ -139,7 +139,7 @@ export function printWatchingForChangesMessage() {
   info("Watching files for changes", { format: colors.green });
 }
 
-export async function printBrowsePreviewMessage(port: number, path: string) {
+export async function printBrowsePreviewMessage(host: string, port: number, path: string) {
   if (isJupyterHubServer()) {
     const httpReferrer = `${
       jupyterHubHttpReferrer() || "<jupyterhub-server-url>/"
@@ -154,10 +154,10 @@ export async function printBrowsePreviewMessage(port: number, path: string) {
     (isJupyterServer() || isVSCodeTerminal()) && isRStudioWorkbench()
   ) {
     const url = await rswURL(port, path);
-    info(`\nPreview server: http://localhost:${port}/`);
+    info(`\nPreview server: http://${host}:${port}/`);
     info(`\nBrowse at ${url}`, { format: colors.green });
   } else {
-    const url = `http://localhost:${port}/${path}`;
+    const url = `http://${host}:${port}/${path}`;
     if (!isRStudioServer()) {
       info(`Browse at `, {
         newline: false,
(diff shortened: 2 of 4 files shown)


jjallaire:
We have to be a little careful about making this change as the current shipping versions of RStudio IDE and the Quarto VSCode extension look explicitly for the token "localhost" when attempting to determine the port that the preview server is running on. I believe that the bug here is the terminal display (the actual host is indeed bound to e.g. 127.17.0.2). Is that what you are seeing?

jjallaire:
It would probably suffice to just make sure we continue printing “localhost” for the default case rather than 127.0.0.1

rcannood:
Thanks for your input! I can make the necessary changes to ensure that it prints localhost by default. I do think that the PR creates functional changes, as the `url` in L181 is passed to openUrl on L187.

jjallaire:
Yes, that `openUrl()` call should be fine (but should also use `localhost` by default if no `host` is specified)

rcannood:
I see, it's indeed only the message that's wrong. I updated the PR based on your feedback:

```
$ quarto preview --host 192.168.178.10
Preparing to preview

Watching files for changes
Browse at http://192.168.178.10:7210/
GET: /

$ quarto preview
Preparing to preview

Watching files for changes
Browse at http://localhost:6470/
GET: /

jjallaire:
Thanks, one more thing though. I don't know for certain if `localhost` will always work correctly in the places we currently use `127.0.0.1` (e.g. calls to listen). I am probably just being paranoid but I don't want to push a change that breaks in some configurations. So if we could have the `kLocalhost` constant still be "localhost" and then have a helper function for embedding it in URLs that converts "127.0.0.1" to "localhost" that would be great.

rcannood:
> So if we could have the kLocalhost constant still be "localhost" and then have a helper function for embedding it in URLs that converts "127.0.0.1" to "localhost" that would be great.

I made some changes based on your suggestions. If you see things differently, feel free to modify my PR as you see fit :)

> Also if you could email a contributor agreement as described here that would be great: https://quarto.org/about.html#pull-requests

Will do!

rcannood:
Hi @jjallaire !

Did you receive the contributor agreement I sent you?

Do you have any further remarks regarding the content of this PR?

Robrecht

jjallaire:
I haven't seen a contributor agreement yet. Could you re-send?

My only remaining comment on the PR is it would be nice if the url passed to `openUrl` also has "localhost" in it rather than "127.0.0.1".

PR #977, opened 2022-05-25
Delete old dist directory on windows too

cderv:
@dragonstyle I think this is missing. It seems I was having since a few weeks some files that should not be there, and that messed up my dev configuration. 
https://github.com/quarto-dev/quarto-cli/blob/9d390e964aa0c7bc3ae5e7676164a41b1d447c70/configure-linux.sh#L20-L21
https://github.com/quarto-dev/quarto-cli/blob/9d390e964aa0c7bc3ae5e7676164a41b1d447c70/configure-macos.sh#L9-L10

Do you think there is other part of the process that should be adapted ?


dragonstyle (approved):
👍

PR #1022, opened 2022-05-30
Add link to possible CSL item types

salim-b:


jjallaire:
Reference data is always edited in YAML (JSON is generated). You can change this line of code here: https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/schema/definitions.yml#L1772

salim-b:
Ah, sorry! Now I get it🙈

PR #1051, opened 2022-06-03
[Windows] Improve R finding for special installation

### LINKED ISSUE #995: Failed to locate `Rscript.exe` if R was installed via scoop in RStudio
I was using [scoop](https://scoop.sh/) to install R (via the main bucket), RStudio and Quarto (both via [cderv/r-bucket](https://github.com/cderv/r-bucket)). However, I found a wired issue that RStudio errored when rendering quarto documents saying that:

```
ERROR: Error executing 'C:\Users\hongy\scoop\apps\r\current\bin\Rscript': 系统找不到指定的文件。 (os error 2)

```

But directly running `quarto render test.qmd` from command line works.

After some search though the codebase, I found that `rBinaryPath("Rscript")` was called to run `Rscript` with given arguments. `rBinaryPath()` internally uses `QUARTO_R`, `R_HOME` and Windows Registry entry to locate the `bin` directory under the R installation folder. Then use `bin/Rscript.exe` as the target path. 

https://github.com/quarto-dev/quarto-cli/blob/af5c051e8251d3a0535f2e8dcf62d011259c8fb2/src/core/run/r.ts#L14-L33

However, this did not work if R was installed via scoop in RStudio. There are several related issues here:

If R was installed via scoop:

1. There is no Windows Register key written
2. The `Rscipt.exe` was only at `bin\x64|x32\Rscript.exe`, and scoop did not create links for `Rscript.exe` under the `bin` folder. See: https://github.com/ScoopInstaller/Main/blob/master/bucket/r.json#L22-L40
3. Shims for `Rscript.exe`, `R.exe`, `Rterm.exe` and etc. were created under `~/scoop/shims`. This folder was added to the user PATH so that every other program can access them. This is why directly running `quarto render` from the command line works. `rBinaryPath("Rscript")` will use the path that `which("Rscript")` returns.

However, when quarto was called in R, since the `R_HOME` environment variable was automatically created. `rBinaryPath("Rscript")` will use `R_HOME/bin/Rscript.exe` which does not exist.

One solution maybe is instead of directly returning `R_HOME/bin/Rscript.exe`:
1. Check if the path exists
2. If not, try testing `R_HOME/bin/{arch}/Rscript.exe`

cderv:
This fixes #995

After some tests with R installation with Scoop and rig on Windows, here are some ajustement in what we are doing.
With this change:  
* Scoop or rig installation will be found when in terminal for PATH or registry version
* They will also be found inside RStudio where `R_HOME` is set and takes precedence. 

For `scoop`, as described in #995, the installation will not add a `Rscript.exe` file in `R_HOME/bin`, they will be in `R_HOME/bin/x64` only. I kept the adjustment simple for now by adding directly the path to `x64` folder if none is found in root `bin` folder. This is really scoped to the specific issue. If we want more generic approach, it could be clever to search for the `Rscript.exe` within the `R_HOME` folder and take the first one found. Could be done by refactoring the search in program files 
https://github.com/quarto-dev/quarto-cli/blob/abb46a31904665e3774319520e7d8473c869290d/src/core/resources.ts#L97-L106
If needed, I'll add that. 

For `rig`, it took me more time to test because I found a few issues that messed up my system; They are reported in https://github.com/r-lib/rig.  It helps though identified one thing: `which()` uses `CMD /C WHERE <binary>`. This command on windows will return all the files found, and not just the only one like `which` on Linux.  I added a handling of this: we are keeping only the first path returned. 

Regarding search in registry, , I noticed a few things that I was not sure : 

* Is there a reason to no read the version install path in two steps ? 
https://github.com/quarto-dev/quarto-cli/blob/abb46a31904665e3774319520e7d8473c869290d/src/core/resources.ts#L76-L93
`Software\\R-core\\R` as a `InstallPath` value directly. So I was wondering by curiosity. I did not change anything as this is working fine currently. 

* Slightly unrelated, we have two sets of registry reading functions: in `registry.ts` and in `windows.ts`. Should we try refactor into one set for easier upkeep in the future ? Asking because I fixed one issue in the past in `windows.ts` for chrome finding, but not the other that I wasn't aware of.

### DIFF
--- a/src/core/path.ts
+++ b/src/core/path.ts
@@ -95,7 +95,10 @@ export async function which(cmd: string) {
     { cmd: args, stderr: "piped", stdout: "piped" },
   );
   if (result.code === 0) {
-    return result.stdout?.trim();
+    return Deno.build.os === "windows"
+      // WHERE return all files found, only first is kept
+      ? result.stdout?.split("\n")[0].trim()
+      : result.stdout?.trim();
   } else {
     return undefined;
   }
--- a/src/core/resources.ts
+++ b/src/core/resources.ts
@@ -8,7 +8,7 @@
 import { existsSync, walkSync } from "fs/mod.ts";
 import { dirname, join } from "path/mod.ts";
 import { warnOnce } from "./log.ts";
-import { which } from "./path.ts";
+import { safeExistsSync, which } from "./path.ts";
 import { quartoConfig } from "./quarto.ts";
 import {
   kHKeyCurrentUser,
@@ -63,7 +63,13 @@ export async function rBinaryPath(binary: string): Promise<string> {
   // if there is an R_HOME then respect that
   const rHome = Deno.env.get("R_HOME");
   if (rHome) {
-    return join(rHome, "bin", binary);
+    let rHomeBin = join(rHome, "bin", binary);
+    if (safeExistsSync(rHomeBin)) return rHomeBin;
+    if (Deno.build.os === "windows") {
+      // Some installation have binaries in the sub folder only
+      rHomeBin = join(rHome, "bin", "x64", binary);
+      if (safeExistsSync(rHomeBin)) return rHomeBin;
+    }
   }
 
   // then check the path


jjallaire:
I'll let @dragonstyle comment on the two-step registry reading (but I agree that leaving it alone is best for now in any case).

As far as refactoring registry functions, let's do this on v1.1 (trying to minimize changes that could lead to breakage on v1.0 right now).

PR #1053, opened 2022-06-03
[reveal] Apply a fix for fenced div starting with a header

### LINKED ISSUE #976: Putting h(3) in a fragment causes a loop in reveal.js
Example:

```
---
format: revealjs
---
# Text 1

## Text 2

Text.

::: {.fragment}
### Text 3

Text.
:::

## Final Text

Text.
```

You can't see the final slide as it loops back to the starts after Text 3.

cderv:
To overcome a current Pandoc limitation which makes fenced divs followed by a header a section, even if header below `--slide-level`

This closes #976 and as discussed there, it has been fixed in current Pandoc dev version, but we won't probably make an update of Pandoc version before conference. This PR adds the workaround described in https://github.com/quarto-dev/quarto-cli/issues/976#issuecomment-1144792316 

It adds a comment line right after the Div block:  Do you think this could cause an issue with our other filters ? 
`````markdown
:::: {.fragment}
<!-- -->
### Header

Content 
:::
`````

Other solution is to add an `id` on the fragment div 
````markdown
:::: {.fragment #q-frag-1}
### Header

Content 
:::
````

but I felt it was more complex to generate some random id than add comment line. 

Regarding the where for this filter, I have put that in the quarto post part really so that it applies latest and avoid issues. It could have been even considered a fix workaround to run last in quarto filter fininalizer but not sure this was design for this. 

Also, I did add the new small filter in the current `reveal.lua` file to avoid creating another one for reveal.
I just add some struggle to see how to apply two filter function on Div within one quarto filter file that will be combine together. cc @dragonstyle I did not manage to catch you live this week, so I couldn't ask directly. 

Usually a filter can return a list of list for that. Something like 
````lua
return {
  { Div = do_something, Code = do_something()},
  { Div = do_another_thing_after_first}
}
````
but this is not working within our internal Lua filter that will be combined by Quarto using `combineFilters()`

Hopefully, my own usage of `combineFilters` this way is ok.

### DIFF
--- a/src/resources/filters/quarto-post/reveal.lua
+++ b/src/resources/filters/quarto-post/reveal.lua
@@ -3,10 +3,15 @@
 
 function reveal()
   if _quarto.format.isRevealJsOutput() then
-    return {
-      Div = applyPosition,
-      Span = applyPosition,
-      Image = applyPosition
+    return combineFilters{
+      {
+        Div = applyPosition,
+        Span = applyPosition,
+        Image = applyPosition
+      },
+      {
+        Div = fencedDivFix
+      }
     }
   else
     return {}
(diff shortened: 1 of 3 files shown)


jjallaire:
Usage of combineFilters is 100% fine.

PR #1102, opened 2022-06-08
Change the issue template to issue form

eitsupi:
Reference: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

This would be easier for the user to enter checkboxes, etc.
image

As a first step, I have ported the current Markdown template almost as is.
Note that you can, for example, create a new input field for system information or set the quarto version as a required field to be written.


jjallaire:
Thanks for this, looks great!

PR #1114, opened 2022-06-09
update docs for eval and echo

jmbuhr:
originally at: <https://github.com/quarto-dev/quarto-web/pull/213>

### DIFF
--- a/src/resources/schema/cell-codeoutput.yml
+++ b/src/resources/schema/cell-codeoutput.yml
@@ -4,7 +4,12 @@
     execute-only: true
   schema: boolean
   default: true
-  description: "Evaluate code cells (if `false` just echos the code into output)."
+  description: |
+    Evaluate code cells (if `false` just echos the code into output).
+
+    - `true` (default): evaluate code cell
+    - `false`: don't evaluate code cell
+    - `[...]`: A list of positive or negative line numbers to selectively include or exclude lines. Excluded lines are commented out with `## `
 
 - name: echo
   tags:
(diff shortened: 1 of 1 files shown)


src/resources/schema/cell-codeoutput.yml
```diff
@@ -4,7 +4,12 @@
     execute-only: true
   schema: boolean
   default: true
-  description: "Evaluate code cells (if `false` just echos the code into output)."
+  description: |
+    Evaluate code cells (if `false` just echos the code into output).
+
+    - `true` (default): evaluate code cell
+    - `false`: don't evaluate code cell
+    - `[...]`: A list of positive or negative line numbers to selectively include or exclude lines. Excluded lines are commented out with `## `
```
jjallaire, line 12: Is this a knitr-only feature? (I don't think we implemented this for Jupyter). If that's the case I might prefer we keep it out of these docs (as it won't apply to not just jupyter but also observable, diagrams, etc.)

jmbuhr: Yes, I don't think there is an equivalent option for jupyter. I didn't realize this was coupled in the documentation. The original PR only modified `docs/reference/cells/cells-knitr.json`, which is repsonsible for https://quarto.org/docs/reference/cells/cells-knitr.html, but if this is programatically generated from a common source for jupyter and knitr it doesn't work.

jjallaire: Okay, for now let's leave this out of the reference.

jjallaire: I'm actually going include it but note that its knitr only

PR #1291, opened 2022-07-05
Support for Extension Embedding

dragonstyle:
Add support for the ability to embed extensions within other extensions. To install an extension into another extension, use:

`quarto install extension <ext-to-install> --embed <ext-to-embed-within>`

Embedding is only supported for format extensions (e.g. formats can embed the filters / shortcodes from other extensions).

The `quarto update` and `quarto remove` command also accept an optional `--embed` argument.


dragonstyle:
@jjallaire this is good to review- appreciate it.

jjallaire (approved):

PR #1298, opened 2022-07-06
Correctly support multiple classes on title

### LINKED ISSUE #1297: Adding more that one class on header in book is not parsed correctly
### Bug description

Create a book project
````
quarto create-project test-quarto-book --type book
````

Add classes une header like this in `index.qmd`

````
# Preface {.unnumbered .unlisted}
````

Render to PDF -> the classes are not correctly parsed by Pandoc and the curly braces part is taken as is. 

image

in tex 
````tex
\chapter*{Preface \{.unnumbered,.unlisted\}}\
````

I am still trying to figure out where this is can happen in our filters.

It only happens in book project. I can reproduce within a single document

This is using last dev version on Windows 11

### Checklist

- [X] [formatted your issue](https://yihui.org/issue/#please-format-your-issue-correctly) so it is easier for us to read?
- [X] included a minimal, self-contained, and reproducible example?
- [X] documented the quarto version you're running, by providing the output produced by `quarto check` in a terminal in your issue?
- [ ] documented the RStudio IDE version you're running (if applicable), by providing the value displayed in the "About RStudio" main menu dialog?
- [X] documented which operating system you're running? If on Linux, please provide the specific distribution as well.
- [X] upgraded to the latest version, including your versions of R, the RStudio IDE, and relevant R packages?

### PR DESCRIPTION
**Pull request by cderv**
fixes #1297

When multiple classes are provided on heading e.g
````
# Preface {.unnumbered .unlisted}
````

we need to correctly rewrite the markdown from the parsed file to create valid Pandoc markdown. 

By default in string interpolation, `array.join(',')` will be used. So we apply joining with no separator in the step before

### DIFF
--- a/src/project/types/book/book-render.ts
+++ b/src/project/types/book/book-render.ts
@@ -365,7 +365,7 @@ async function mergeExecutedFiles(
                 const keyValueStr = attr.keyvalue.map((kv) => {
                   const escapedValue = kv[1].replaceAll(/"/gm, '\\"');
                   return `${kv[0]}="${escapedValue}" `;
-                });
+                }).join("");
                 const attrContents = `${idStr}${clzStr}${keyValueStr}`.trim();
                 attrStr = `{${attrContents}}`;
               }
--- a/src/project/types/book/book-render.ts
+++ b/src/project/types/book/book-render.ts
@@ -361,7 +361,7 @@ async function mergeExecutedFiles(
                 const idStr = attr.id !== "" ? `#${attr.id} ` : "";
                 const clzStr = attr.classes.map((clz) => {
                   return `.${clz} `;
-                });
+                }).join("");
                 const keyValueStr = attr.keyvalue.map((kv) => {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/project/types/book/book-render.ts [hunk 1]
```diff
@@ -361,7 +361,7 @@ async function mergeExecutedFiles(
                 const idStr = attr.id !== "" ? `#${attr.id} ` : "";
                 const clzStr = attr.classes.map((clz) => {
                   return `.${clz} `;
-                });
+                }).join("");
                 const keyValueStr = attr.keyvalue.map((kv) => {
```
dragonstyle, line 365: Great find! I think this map has the exact same problem and should be fixed the same way.

cderv: I always miss the fact that I should get a generic fix from a specific issue I found. I'll improve! 

Thanks for the review.

cderv: Don in [16c5496](https://github.com/quarto-dev/quarto-cli/pull/1298/commits/16c54961376e13286679441643d13d1af759058d)

PR #1308, opened 2022-07-06
fix(language): `title-block-affiliation-*` and `code-tools-*-all-code`

mcanouil:
This PR changes the following translations:

- `title-block-affiliation-*` 
   Affiliation is also the french word for affiliation.
  "Association" does not have the same meaning and thus is not suitable here.

- `code-tools-*-all-code`  
   "Montrer tout le code" instead of "Montrer tous les codes", because even if the latter is correct, when we talk about a global entity we use the former.
   Here, the option is to show/hide the code everywhere, this is not exactly the same as "show/hide every pieces of code".


mcanouil:
I can add inclusive translation if needed #1309

cderv (approved):
Thanks. I agree that the suggestion better suits the context.

PR #1376, opened 2022-07-13
`quarto tools` help page improvment

cderv:
* `quarto tools list` is the way to list all status following change in d3a6418
* `quarto tools info` was not documented (I looked for it in one of the github discussion we had to find the syntax)

### DIFF
--- a/src/command/tools/cmd.ts
+++ b/src/command/tools/cmd.ts
@@ -34,7 +34,7 @@ export const toolsCommand = new Command()
     uninstall
     update
     
-Use 'quarto tools' with no arguments to show the status of all tools.`,
+`,
   )
   .example(
     "Install TinyTex",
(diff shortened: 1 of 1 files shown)


jjallaire:
@dragonstyle didn't we deprecate "quarto tools" in favor verb based variants?

dragonstyle:
We have deprecated that command (though it is still functional). In particular, I haven't replaced `quarto tools info <toolname>`, which prints a json summary of the tools (install status, paths, etc...). I wasn't sure what global command would make sense for this (e.g. `quarto info tool tinytex`, `quarto summarize tool tinytex`, 'quarto status tool tinytex`). 

I like `quarto info` I think but curious if you have a reaction / idea. The function for printing the tool is very straightforward so I can easily add the command and update the docs.

dragonstyle:
FYI @cderv I'm going to close this since we have the tracking issues to deal with this post 1.0 - want to push as much as possible to post 1.0 wherever we can.

cderv:
Thanks ! This all makes sense.

PR #1663, opened 2022-08-03
Improve `toc-depth` in html output

dragonstyle:
The `toc-depth` option will now control how deep html table of contents will display (previously, we always displayed to TOC to a depth of 3). This will control what depth of items are show when the `toc-location` is body. It will control what depth of items will be opened / tracked when scrolling when `toc-location` is left or right.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,5 @@
 # Changelog
 
 ## Unreleased
+
+- respect `toc-depth` in the HTML format (bootstrap) rather than always acting as if depth is 3.
--- a/src/format/html/format-html-bootstrap.ts
+++ b/src/format/html/format-html-bootstrap.ts
@@ -18,6 +18,7 @@ import {
   kLinkCitations,
   kQuartoTemplateParams,
   kSectionDivs,
+  kTocDepth,
   kTocLocation,
 } from "../../config/constants.ts";
 import {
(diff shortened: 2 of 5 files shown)


src/format/html/format-html-bootstrap.ts
```diff
@@ -275,11 +279,17 @@ function bootstrapHtmlPostprocessor(
       }
 
       // default collapse non-top level TOC nodes
-      const nestedUls = toc.querySelectorAll("ul ul");
-      for (let i = 0; i < nestedUls.length; i++) {
-        const ul = nestedUls[i] as Element;
-        ul.classList.add("collapse");
+      const tocDepth = format.pandoc[kTocDepth] || 3;
```
dragonstyle, line 282: Is there a better place for me to be defaulting this?

changelog.md
```diff
@@ -1,3 +1,5 @@
 # Changelog
 
 ## Unreleased
+
+- respect `toc-depth` in the HTML format (bootstrap) rather than always acting as if depth is 3.
```
jjallaire, line 5: Changelog should be a separate commit (in case we want to cherry pick back onto v1.0)

src/format/html/format-html-bootstrap.ts
jjallaire: That seems fine to me

changelog.md
```diff
@@ -1,3 +1,5 @@
 # Changelog
 
 ## Unreleased
```
jjallaire, line 3: Let's remove the changelog entry here

PR #1687, opened 2022-08-04
conda plumbing

msarahan:
This is a work in progress. It adds a development conda recipe that can be used to build a conda package using the Quarto source in the working directory. I've been using it to develop changes to quarto that will ease the eventual conda-forge recipe. Most of the changes here are to make paths more absolute, and to avoid downloading/vendoring binaries that are available on conda as separate packages.

I'm working on Linux first.


msarahan:
Right now, things are crashing, likely at the point where deno-dom is loaded:

```
Building Intelligence Resources
In file included from <string>:1:
/usr/include/stdint.h:26: error: include file 'bits/libc-header-start.h' not found

============================================================
Deno has panicked. This is a bug in Deno. Please report this
at https://github.com/denoland/deno/issues/new.
If you can reliably reproduce this panic, include the
reproduction steps and re-run with the RUST_BACKTRACE=1 env
var set and include the backtrace in your report.

Platform: linux x86_64
Version: 1.24.0
Args: ["/home/msarahan/mambaforge/conda-bld/quarto-cli_1659638627749/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/bin/deno", "run", "--unstable", "--allow-env", "--allow-read", "--allow-write", "--allow-run", "--allow-net", "--allow-ffi", "--importmap=/home/msarahan/mambaforge/conda-bld/quarto-cli_1659638627749/work/src/dev_import_map.json", "/home/msarahan/mambaforge/conda-bld/quarto-cli_1659638627749/work/package/src/bld.ts", "prepare-dist"]

thread 'main' panicked at 'gen_trampoline: ()', /home/conda/.cargo/registry/src/github.com-1ecc6299db9ec823/deno_ffi-0.49.0/lib.rs:735:51
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/panicking.rs:142:14
   2: core::result::unwrap_failed
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/result.rs:1785:5
   3: deno_ffi::make_sync_fn
   4: deno_ffi::op_ffi_load::v8_func
   5: <extern "C" fn(A0) .> R as v8::support::CFnFrom<F>>::mapping::c_fn
   6: _ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
             at ./../../../../v8/src/api/api-arguments-inl.h:147:3
   7: _ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEESA_NS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EENS0_16BuiltinArgumentsE
             at ./../../../../v8/src/builtins/builtins-api.cc:114:36
   8: _ZN2v88internal21Builtin_HandleApiCallEiPmPNS0_7IsolateE
             at ./../../../../v8/src/builtins/builtins-api.cc:132:1
   9: Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit
```

.github/workflows/create-release.yml
```diff
@@ -16,7 +16,7 @@ on:
         description: 'Draft'
         required: false
         default: 'false'
-
```
msarahan, line 19: This file has only whitespace changes. I'm fine to revert it to simplify this PR.

package/src/common/configure.ts
```diff
@@ -53,19 +53,21 @@ export async function configure(
...
+  info("Copied Quarto script to bin directory");
+
   // record dev config
   const devConfig = createDevConfig(
-    Deno.env.get("DENO") || "",
-    Deno.env.get("DENO_DOM") || "",
-    Deno.env.get("PANDOC") || "",
-    Deno.env.get("DARTSASS") || "",
-    Deno.env.get("ESBUILD") || "",
+    Deno.env.get("QUARTO_DENO_PATH") || "",
```
msarahan, line 60: These would be set with a conda activation script if providing external executables. The activation script would probably be part of the quarto conda package, as opposed to an activation script in each of the other packages. Generally, it's better to rely on PATH, but using these here lets us manually specify the potentially vendored paths.

package/src/common/update-html-dependencies.ts
```diff
@@ -487,7 +487,7 @@ async function updatePdfJs(
     "pdf",
     "pdfjs",
   );
-  moveSync(from, to, { overwrite: true });
+  copySync(from, to, { overwrite: true });
```
msarahan, line 490: copySync is necessary because from/to may be on different partitions (temp fs to a disk). Moving/renaming doesn't work under those conditions.

package/src/macos/installer.ts
```diff
@@ -78,7 +78,7 @@ export async function makeInstallerMac(config: Configuration) {
     // Sign these executable / binary files
     // and include our entitlements declaration
     const signWithEntitlements = [
-      join(config.directoryInfo.bin, "tools", "deno"),
+      Deno.env.get("QUARTO_DENO_PATH"),
```
msarahan, line 81: probably don't need to sign the conda binary. This file is probably fine to revert.

msarahan:
I think this is ready for review. Here's an overview of the kinds of changes that are made here:

* consolidating the configure scripts for mac/linux: https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R28
* Allow overrides of the binary dependency versions from `configuration` - the conda packages plug in numbers according to what binary versions are present at build time. My hope is that subsequent Javascript build steps account for these version differences. https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-b616abe240ebf422563d30b460a4262aaf223ba4d0170e7ee4591b7d992f01a1R14-R21 and https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-0ba17b1855b0f97042c31b2b2c2fd2b075fa355a840b152f1959daf2a755266eR5-R12
* Add env var to determine whether binary vendoring happens: https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-0ba17b1855b0f97042c31b2b2c2fd2b075fa355a840b152f1959daf2a755266eR30
* Development [conda recipe](https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-509a47a718bd7e8905b1c18a6af970d2c9d557ff53896759823a11f2b085ffb9), with [win script](https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-68bd70ec78b286ab6a1248622e79fb564343099b34c663f641837d55a5fd9ef5) and [mac/linux script](https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-9be544e0205a4d591a94bd58047fe5d869aab7093f780c1bfb2bb33b47dd5a43). This recipe is basically what will be submitted to conda-forge once these changes are merged and a new release is tagged. The source section will change to point at that new github tag instead of the local source folder.
* Moving the quarto version file to a more descriptive name, since it lives in a much more shared location in a conda environment. https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-31a537782688da9d3fb3a8ec52abb45e9a21dbd666a6b0075d13f3210a78adc8R83
* quote lines that have trailing spaces so that linting on saving doesn't screw up batch scripts - https://github.com/quarto-dev/quarto-cli/pull/1687/files#diff-8945c4925fb67d944d253ab16ef903a1d794674c66db1f4eddcd604046869ceeL104

.vscode/launch.json
```diff
@@ -21,7 +21,7 @@
       ],
       "env": {
         "QUARTO_BIN_PATH": "${workspaceFolder}/package/dist/bin",
-        "QUARTO_SHARE_PATH": "${workspaceFolder}/src/resources",
```
jjallaire, line 24: We have code in the field that relies on `QUARTO_SHARE_PATH` (IDE uses it for crossref indexing). It's fine to change but we should just make sure that the old environment variable is still respected.

package/conda-recipe/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+source:
+  path: ../..
+
+build:
+  number: 0
+
+requirements:
+  host:
+    - deno
```
jjallaire, line 16: Our source code is definitely sensitive to these version requirements:

- Deno must be v1.22.0 (higher versions will have compile errors)
- Pandoc must be 2.19 (previous versions will fail w/ unsupported command line options)

Generally speaking our Deno and Pandoc requirements will remain _very specific_ as both projects frequently introduce breaking changes.

package/scripts/common/quarto
```diff
@@ -27,7 +27,7 @@ if [ -f "$DEV_PATH" ]; then
     QUARTO_TARGET=$DEV_PATH
   fi
   export QUARTO_BIN_PATH=$SCRIPT_PATH
-  export QUARTO_SHARE_PATH="`cd "$SCRIPT_PATH/../../../src/resources/";pwd`"
+  export QUARTO_SHARE_DIR="`cd "$SCRIPT_PATH/../../../src/resources/";pwd`"
```
jjallaire, line 30: See previous note about continuing to respect the previous `QUARTO_SHARE_PATH` variable for the sake of the IDE.

package/scripts/common/quarto
```diff
@@ -71,26 +67,28 @@ else
...
+    elif [[ $SCRIPT_PATH == *"/usr/local/bin/quarto" ]]; then
+      export QUARTO_SHARE_DIR="`cd "$SCRIPT_PATH/../../share/quarto";pwd`"
+    else
+      export QUARTO_SHARE_DIR="`cd "$SCRIPT_PATH/../share";pwd`"
+    fi
   fi
 
   if [ "$1" == "--version" ] || [ "$1" == "-v" ]; then
-    echo `cat "$QUARTO_SHARE_PATH/version"`
+    echo `cat "$QUARTO_SHARE_DIR/quarto_version"`
```
jjallaire, line 83: Sadly the RStudio IDE also relies on the presence of the `version` file so we can't readily rename this.

package/scripts/deno_std/download.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-export DENO_DIR=src/resources/deno_std/cache 
-package/dist/bin/tools/deno cache --unstable --lock src/resources/deno_std/deno_std.lock "$@" package/scripts/deno_std/deno_std.ts
-
+if [ -z "$QUARTO_DENO" ]; then
```
jjallaire, line 3: Currently we run the lock script (which alongside this) to periodically lock the embedded version of the deno std library. This script calls download.sh so currently will fail if run from the top level.

package/src/common/config.ts
```diff
@@ -35,14 +35,15 @@ export function readConfiguration(
...
-  const out = join(pkg, getEnv("QUARTO_OUT_DIR"));
-  const bin = join(dist, getEnv("QUARTO_BIN_DIR"));
+  const root = getEnv("QUARTO_ROOT");
+  const src = getEnv("QUARTO_SRC_DIR");
+
+  const pkg = getEnv("QUARTO_PACKAGE_DIR");
+  const out = getEnv("QUARTO_OUT_DIR");
+
+  const dist = getEnv("QUARTO_DIST_DIR");
+  const share = getEnv("QUARTO_SHARE_DIR");
```
jjallaire, line 45: Here is at least one place where we need to respect `QUARTO_SHARE_PATH` (there may be others)

src/project/serve/serve.ts
```diff
@@ -554,6 +554,13 @@ export async function serveProject(
   }
 }
 
+// https://deno.com/blog/v1.23#remove-unstable-denosleepsync-api
```
jjallaire, line 557: It looks like you got things to work w/ Deno 1.23 w/ this change. That's fine, but as an FYI there are more changes required for 1.24 (and we currently have an issue whereby 1.24 crashes on M1 macs so we can't update to it).

src/resources/js/graphviz-wasm.js
```diff
@@ -28,7 +28,7 @@ export function graphvizSync() {
 }
 
 export function wasmPath() {
-  return `${Deno.env.get("QUARTO_SHARE_PATH")}/wasm/graphvizlib.wasm`;
```
jjallaire, line 31: Respect `QUARTO_SHARE_PATH` here.

src/resources/pandoc/datadir/init.lua
```diff
@@ -1261,7 +1261,7 @@ modify_lua_functions{
 }
 
 -- Bootstrap our common libraries by adding our filter pandoc to the lib path
-local sharePath = os.getenv("QUARTO_SHARE_PATH");
```
jjallaire, line 1264: Respect `QUARTO_SHARE_PATH` here

jjallaire (changes requested):
I reviewed but would also like to see a review by @dragonstyle before we merge (as he wrote nearly all the code that is being changed -- note that that he's on vacation until Monday). I'd also like to see this merged for our v1.1 release as it would be great to not have to wait on v1.2 for Conda)

package/conda-recipe/meta.yaml
msarahan: yes, I'm wary of that. I would like to get some kind of test suite running as part of this package build. I don't have the older versions of deno/deno-dom built on conda-forge yet. I will get them added.

.vscode/launch.json
msarahan: Is QUARTO_SHARE_DIR also in use? I have no preference, I just didn't like that we had two variables for the same thing.

package/scripts/common/quarto
msarahan: We'll need another place then, perhaps with a different env var. We can't just drop a file like that into a shared folder where it's likely to have conflicts. Or maybe I'm misinterpreting the share dir and in the conda package it should be something like `QUARTO_SHARE_DIR=$PREFIX/share/quarto`, and this `version` file would live in there.

src/project/serve/serve.ts
msarahan: Is there an issue you can point me to about this M1 problem? I'd like to test my packages against it, and get it working if I can.

.vscode/launch.json
jjallaire: Perhaps we use `QUARTO_SHARE_PATH` in the product and `QUARTO_SHARE_DIR` in some configure scripts? I don't know. @dragonstyle Could you provide clarity here?

package/scripts/common/quarto
jjallaire: Yes exactly, the share dir is a quarto specific share dir NOT the global one

src/project/serve/serve.ts
jjallaire: This is a PR that has the upgrade to 1.24.1: https://github.com/quarto-dev/quarto-cli/pull/1945

Everything works great except if you render on an newer M1 Mac you see sporadic crashing (at the Rust runtime level). We speculated that it could be FFI issues however we disabled deno dom native and still saw the crash (so perhaps in WASM loading which we do in a number of places).

jjallaire: Note that we don't plan on trying to merge this until v1.2 so your current PR should stay based on Deno 1.23 if want this to land in v1.1

.vscode/launch.json
dragonstyle: `QUARTO_SHARE_DIR` was used in configuration to provide the simple name of the share directory (e.g. `share`). `QUARTO_SHARE_PATH` was used in the bash script to contain the complete path to the share directory.

We followed this general pattern (`*_DIR` = just a simple name, `*_PATH` = an actual path) throughout configuration (or at least we generally should've).

package/src/common/configure.ts
```diff
@@ -55,17 +55,17 @@ export async function configure(
...
+    Deno.env.get("QUARTO_DART_SASS") || "",
+    Deno.env.get("QUARTO_ESBUILD") || "",
     config.directoryInfo.bin,
   );
   writeDevConfig(devConfig, config.directoryInfo.bin);
-  info("");
+  info("Wrote dev config to bin directory");
 
-  if (Deno.build.os !== "windows") {
+  if (Deno.build.os !== "windows" && Deno.env.get("QUARTO_NO_SYMLINK") !== undefined) {
```
dragonstyle, line 68: The `QUARTO_NO_SYMLINK` behavior may be reversed here unless I'm misunderstanding it.

dragonstyle (commented):
Overall looks good to me - I'd be in favor of keeping `QUARTO_SHARE_PATH` to avoid indirect dependency issues!

package/src/common/configure.ts
jjallaire: In that case since they have different uses I think we should keep `QUARTO_SHARE_DIR` and `QUARTO_SHARE_PATH` exactly as they are

configure-windows.cmd
```diff
@@ -9,101 +9,125 @@ if "%ERRORLEVEL%"=="0" goto :denoRunning
 REM Reads the configuration file line by line
```
jjallaire, line 9: Could we now rename this file `configure.cmd` ?

jjallaire:
@msarahan This seems like it might be ready to merge (after the conflicts are resolved). Is that correct?

msarahan:
I'm taking one more pass at very carefully examining the _DIR vs _PATH stuff, trying to make it all consistent with what @dragonstyle said. I will ping you when I think it is all set.

jjallaire:
Okay, great. On the last pass all the checks are failing w/ inability to download deno. This could be transient or could be related to the changes?

jjallaire:
There is no "unknown-linux" w/ a .tar.gz extension. There is however a .zip one:

Screen Shot 2022-08-18 at 9 22 05 AM

msarahan:
That seems like something I messed up somehow with the consolidation of the configure scripts. I'll try to figure out how I did that.

msarahan:
Ah ha! It was github copilot being helpful and messing up!

msarahan:
Looks like I’ve trashed the batch file stuff by trying to factor out the
path settings. I’ll keep working at it on my windows machine.

jjallaire:
Okay, CI should be enabled (LMK on Slack if it isn't).

FYI we just updated to Pandoc 2.19.1 (which came out today).

.github/workflows/actions/quarto-dev/action.yml
```diff
@@ -3,23 +3,21 @@ description: 'Configures the image for Quarto Development'
...
           echo "$HOME/bin" >> $GITHUB_PATH
-      
+
       - name: Configure Quarto
         shell: bash
         run: |
-          ./configure-linux.sh
-          
+          ./configure.sh
+          package/src/quarto-bld prepare-dist
```
msarahan, line 16: I'm not sure yet what happened here. I don't think this should be necessary, but without it the vendor folder is not copied over as part of the configure script, so lots of stuff is broken.

configure.sh
```diff
@@ -0,0 +1,108 @@
...
+fi
+
+# Run the quarto command with 'reload', which will force the import_map dependencies
+# to be reloaded
+if ! quarto_loc="$(type -p quarto)" || [[ -z $quarto_loc ]]; then
+  echo "Quarto symlink doesn't appear to be configured."
+else
+  export QUARTO_DENO_EXTRA_OPTIONS="--reload"
+  mkdir -p $QUARTO_SHARE_PATH
+  echo "$QUARTO_VERSION" > $QUARTO_SHARE_PATH/version
```
msarahan, line 106: From what I've seen, the version file is only generated with prepare-dist. It doesn't exist the first time this script is run. I don't know how this test could succeed without adding this file. Is that supposed to happen?

package/scripts/common/quarto
```diff
@@ -2,66 +2,78 @@
 
 # Determine the path to this script (we'll use this to figure out relative positions of other files)
 SOURCE="${BASH_SOURCE[0]}"
-while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+if [ -h "$SOURCE" ]; then
```
msarahan, line 5: I changed this because I found that this misbehaved on conda, where there is no symlink at all

package/scripts/windows/quarto.cmd
```diff
@@ -79,34 +79,22 @@ IF "%1"=="--paths" (
...
 )
 
 SET QUARTO_DENO_OPTIONS=--unstable --no-config --cached-only --allow-read --allow-write --allow-run --allow-env --allow-net --allow-ffi %QUARTO_DENO_OPTIONS%
 
 %QUARTO_DENO% %QUARTO_ACTION% %QUARTO_DENO_OPTIONS% %QUARTO_DENO_EXTRA_OPTIONS% %QUARTO_IMPORT_ARGMAP% %QUARTO_TARGET% %*
 
-
 GOTO :end
-
-REM Reads each line of a file and converts any exports into SETs
```
msarahan, line 101: This is moved to the package/src/store_win_configuration.bat file. It outputs a static file of variables that other scripts can use.

src/core/devconfig.ts
```diff
@@ -93,20 +94,13 @@ export function readSourceDevConfig(): DevConfig {
...
-    isWindows() ? "windows" : "common",
-  );
   return createDevConfig(
     readConfig("DENO"),
     readConfig("DENO_DOM"),
     readConfig("PANDOC"),
     readConfig("DARTSASS"),
     readConfig("ESBUILD"),
-    srcScriptPath,
+    quartoConfig.binPath(),
```
msarahan, line 103: These paths are pretty confusing to me. It seems like quartoConfig is meant to only expose paths that are in the dist folder (bin and share). If it's OK for other folders to have accessor methods, I'll add them and use them instead of these funky relative paths.

configure.sh
msarahan: never mind, I see that it should never get to looking for that file. Something else isn't quite right.

src/core/devconfig.ts
jjallaire: I honestly don't remember enough about these paths to say whether a change is a good idea or likely to introduce other problems. I'd say maybe keep them as-is for now (just so we can close down v1.1) then revisit for v1.2. Does that sound okay?

jjallaire:
Do we need to change the versions in the conda config to pin them?

src/core/devconfig.ts
msarahan: Makes sense!

jjallaire:
@dragonstyle Before we merge this I'd like to try out creating installers and testing them for each platform -- I don't want to mess up the existing pre-release flow though. What's the easiest way to do this?

msarahan:
Regarding the Conda recipe, I was omitting the pins because I don't want to create an ongoing maintenance burden. I will add them in here, but I will also work on creating a way to have conda-build read the versions in from the `configuration` file, so that the recipe stays in sync.

dragonstyle:
I think the easiest thing to do would be run the `Build Installers` action on the branch without publishing a release. Once the job is complete, the artifacts (installers) created by the action should be available when you view the completed action's detail page.

jjallaire:
And they won't become the new pre-release?

dragonstyle:
There is an option to not even publish a release, which won't create the release or pre-release. You can also select 'draft' to go with belt and suspenders as drafts won't reach pre-release status.

msarahan:
I added in the  conda pins. They are parsed from the `configuration` file, so there shouldn't be any maintenance associated with the recipe. I will carry this over to the conda-forge recipe as well. People there may not like such tight pinning, but until we can prove stability with other versions I agree that it is necessary.

jjallaire:
It looks like I can't build the installers until we merge. I'll do that as a draft after we merge. I'm also going to squash when merging just so we can back the whole thing out cleanly if we need to (e.g. if there are installer problems)

msarahan:
sure thing. I did just squash it, so that should be taken care of.

.github/workflows/actions/quarto-dev/action.yml
jjallaire: @dragonstyle Is there a better or alternate way to do this?

Issue #1609, opened 2022-07-31
Zotero metadata for automatic parsing of citation


dmi3kno:
In the [Creating Citeable Articles](https://quarto.org/docs/authoring/create-citeable-articles.html#google-scholar) the metadata compatible with Google Scholar is listed. The set of metadata tags accepted by Google Scholar is a subset of a more general open standard parsed by [Zotero](https://www.zotero.org/support/dev/exposing_metadata) based on Highwire vocabulary.  If metadata is done properly, Zotero browser plug-in will be able to generate proper record from the Quarto article.

I suggest we expand the list of metadata included in the `google-scholar` partial or add a new partial called `zotero`, which can include more fields (beyond the minimally required by Google Scholar).

### Google Scholar

| Tag | YAML source | Comment | 
| --- | --- | --- |
| citation_title | title | Document title |
| citation_author | author:name | Should be used for single author documents |
| citation_online_date | date | Online publication date |
| citation_fulltext_html_url | citation:URL | URL of the document |
|  citation_publication_date | date | Publication date - same as online date above? |
|  citation_journal_title | citation:container-title | Journal title | 
|  citation_reference | bibliography (parsed) | One entry per reference in bibliography | 

### Zotero
Everything above plus:

| Tag | Suggested YAML source | Comment | 
| --- | --- | --- |
| citation_editor | citation:editor | One entry per editor name, can be simplified to a single semicolon-separated string passed to citation_editors  |
| citation_date| citation:date | Date for the citation (not necessarily when it appeared online) |
| citation_cover_date| citation:publication_date | Date of publication (date of journal issue). Used by [Springer](https://forums.zotero.org/discussion/59197/springer-useing-first-online-instead-of-real-publication-date)  |
| citation_year | - | Can be automatically computed from one of the date fields |
| citation_journal_abbrev | citation:container-abbr | Correct abbreviation for the journal name |
| citation_book_title | citation:book-title | Name of the book where the paper appears |
| citation_inbook_title | citation:inbook-title | Name of the book where the paper appears |
| citation_firstpage | citation:first-page | First page number for the paper (in the journal issue) |
| citation_lastpage | citation:last-page | Last page number for the paper (in the journal issue) |
| citation_volume | citation:volume | Journal volume number |
| citation_issue | citation:issue | Journal issue number |
| citation_series_title | citation:series-title | Title of journal/book series |
| citation_publisher | citation:publisher | Name of the publisher |
[...]

PR #1701, opened 2022-08-05
Improve the creation of citations

dragonstyle:
Bug #1609 has a number of great suggestions for improving citation metadata for documents. This addresses most of those suggestions, generating additional `citation_` meta tags based upon document citation data as well as making our handling of citation data more consistent.

We will newly output:
`citation_abstract`
`citation_doi`
`citation_pmid`
`citation_language`
`citation_keywords`
`citation_pdf_url`
`citation_public_url`
`citation_abstract_url`
`citation_fulltext_url`
`citation_journal_abbrev`
`citation_publisher`

Based upon the `type` specified:
`citation_conference_title`/`citation_conference`
`citation_dissertation_institution`
`citation_book_title`

General approach is to allow specification of citation data with names and structure that corresponds with CSL schema and to map those values onto the appropriate meta names (rather than matching the meta names directly).

### DIFF
--- a/src/core/csl.ts
+++ b/src/core/csl.ts
@@ -5,7 +5,20 @@
 *
 */
 
-import { parsePandocDate } from "./date.ts";
+import { formatDate, parsePandocDate } from "./date.ts";
+
+export const kPdfUrl = "pdf-url";
+export const kAbstractUrl = "abstract-url";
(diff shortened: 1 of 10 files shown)


dragonstyle:
I need to make a few more changes before we merge this.

PR #2015, opened 2022-08-20
Restore conda with dev mode fix

msarahan:
@jjallaire identified an issue where a dev install wasn't working. This was caused by some incorrect path assignments in the quarto script. This PR restores the reversion from https://github.com/quarto-dev/quarto-cli/commit/f2dee12d868020b5719c43bfb1e32e70b801e8b6 and adds a fix for the incorrect path assignments. It reverts the call to prepare-dist in dev mode setup for the CI, and adds some sanity checks to validate that the dev mode setup is working as expected.


msarahan:
Installer build for this branch is at https://github.com/quarto-dev/quarto-cli/actions/runs/2894955674

msarahan:
Mac has an outdated bash version that doesn't support the -v operation. I'll find another way...

jjallaire (commented):
The wholesale reformatting of the YAML isn't great (as late in a release we'd ideally be able to look at these diffs and reason about exactly what changed and with the reformat there is no way to do that). One way I've handled this in the past is to do a standalone reformat commit first and then make the change. 

We also need to be careful to not have dueling reformatters (I know Charles wrote these actions initially and I'm not sure whether any formatter was in play). In our repo I think we currently reformat typescript but it's probably best that we not reformat YAML (unless we are going to make that a standard part of the workspace and reformat all of it at the outset w/ the standard tool).

jjallaire:
BTW I'm fine with this commit as-is (as I realize it could be a hassle to get the minimal diff from here) just wanted to make that comment for future reference.

msarahan:
Agreed, completely. I can revert it and try to get VS Code to leave it alone. The formatting change was not intentional.

msarahan:
I’m done for the night, but I still need to sort out path issues with the tests on windows and Mac. Back at it tomorrow.

jjallaire:
Okay, so much appreciate your effort to get all of this working. It's going to be a massive win for users to be able to install from Conda!

msarahan:
The installer tests I added are working now - https://github.com/quarto-dev/quarto-cli/actions/runs/2899172357

I will squash what I have here, and then do the reformat-only commit for the installer generation action, followed by the additions there.

msarahan:
Hopefully the added tests here will also make future development less error-prone. Thanks for all your effort in helping to find the problems.

msarahan:
@jjallaire i think this is all set. Let me know if you see anything else.

jjallaire:
Indeeds, look great! Merging now, thank you again :-)

PR #2092, opened 2022-08-24
Add colors for tokens with null text color in a11y-dark theme

batpigandme:
Previously, when tokens in the highlighting theme had `"text-color": null`, the CSS would render such that the equivalent `code span` class would be empty, and use the default highlighting color for the theme. In rendering with the latest dev version of quarto (though I'm not sure when/where the change was introduced), the `code span` now uses the color set by the default theme in quarto for that token.

In this PR, I've added appropriate colors for previously `null` classes in a11y-dark, since the color from quarto (`#008800`) has very low contrast with the a11y-dark theme's background color.

Here's a screenshot of HTML rendered with the latest version of quarto with the theme _without_ the changes made in this PR:
image

### DIFF
--- a/src/resources/pandoc/highlight-styles/a11y-dark.theme
+++ b/src/resources/pandoc/highlight-styles/a11y-dark.theme
@@ -4,7 +4,7 @@
             "line-number-color": "#97947a",
                 "line-number-background-color": null,
                     "_comments": [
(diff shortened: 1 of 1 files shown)


jjallaire:
v1.1 is going to get pushed out a bit for a pandoc update and some schema validation work so I'm merging this now.

Issue #2114, opened 2022-08-25
Need test for relative symlinks to Quarto


msarahan:
Issue https://github.com/quarto-dev/quarto-cli/issues/2113 is a problem that I caused by not considering that symlinks to the quarto binary may be relative. Our dev configuration makes absolute path symlinks, and we don't otherwise have a check for this use case. It seems like we should add something to the CI process to test this behavior.

PR #2124, opened 2022-08-25
consolidate configure, download, quarto-bld to shell scripts

msarahan:
Re-opening #2105 here with a branch on the main repo so that I can trigger the installer tests.

Implements https://github.com/quarto-dev/quarto-cli/issues/2071 to make maintaining configure stuff more uniform and easier. I have tested this with git bash and with msys2. I also took the opportunity here to get symlinks working on windows. There's a setting you must change yourself on the OS, but the warning messages here tell you where to go to learn about that setting ("Developer mode").

This is not intended for the 1.1 release cycle - obviously way too late and not the right thing to focus on right now.

One further thing that we might consider is to remove the ".sh" extension from the configure script (for the sake of muscle memory for those of us who have done a lot of `./configure && make && make install`), and possibly rename the `configuration` file so that autocomplete for `./configure` "just works"


.github/workflows/create-release.yml
```diff
@@ -146,12 +146,36 @@ jobs:
...
+      - name: Test directly running quarto script
+        run: |
+          quarto check
+          quarto --paths
+          quarto --version
+      - name: set up bin dir for symlink tests
+        run: |
+          mkdir $GITHUB_WORKSPACE/bin
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+      - name: Test relative symlinks to quarto script
```
msarahan, line 158: New tests here, intended to address #2114

package/scripts/windows/quarto.cmd
```diff
@@ -21,34 +28,44 @@ IF EXIST "!QUARTO_TS_PATH!" (
...
 			echo !DENO!>"!DENO_VERSION_FILE!"
-
-			cd !QUARTO_ROOT!
-			call configure.cmd
-      echo
-			echo Quarto required reconfiguration to install Deno !DENO!. Please try command again.
-			GOTO end
+      echo "Quarto needs reconfiguration to install the correct Deno version (!DENO!)."
+      echo "    (Found Deno version: !DENO_INSTALLED_VERSION!)"
+      echo "Please run the configure.sh script in either git bash or MSYS2."
```
msarahan, line 67: I found that calling "bash configure.sh" ended up having a hard time with windows line endings and failing pretty horribly. Exiting and telling the developer to use MSYS2 or git bash seemed like a better option than trying to get that to work.

msarahan:
This is ready for review whenever you have time. @cderv might especially be interested.

Installers pass OK: https://github.com/quarto-dev/quarto-cli/actions/runs/2927907560

package/scripts/windows/quarto.cmd
cderv: Did not looked at the context for this, but just for reference, when you have WSL installed, calling `bash` will mean running in WSL not in a Windows bash emulator. Which makes things pretty different. Anyway, I'll test this.

msarahan: Yes, that's exactly what happens for me, and that's where the line ending stuff arises. WSL doesn't compensate for it, but the MSYS2/git bash stuff does. Or at least that's my working theory...

cderv: With WSL the script will be running on Ubuntu. Not Windows, so it is not just the Line ending that change. All the downloads and installation won't work I think in that case. We are calling powershell I think to download on Windows even from the sh script, so it definitely should not run under bash for WSL.

msarahan: It wasn't my intent to get that working. I consider WSL to be truly "linux" and it shouldn't be taking the windows path in these scripts at all. Perhaps I messed up that case. I was developing only with git bash and msys2. Is it enough to have it documented as it is now, or should we try to either support it or fail more gracefully in the case where WSL bash gets used?

cderv (commented):
I was not so much acquainted with the whole logic, as I was mainly using it, but never write anything. So maybe I missed some important thing. I think @dragonstyle would be the one to review all this. 

A few question in addition of the comments: 

Is the `configure.sh` only call when we manually ask for a configuration or can it be called when running Quarto itself ? 
Context for my question: I believe there is currently a feature where only `git pull` update and then re-running quarto would reconfigure on the fly. Is that calling the whole configure script ? Do we lost that ? 

Because on Windows, if the `quarto` installed using this methods is called outside of bash Msys windows or equivalent, I don't think executing a `configure.sh` will work (well). 

Regarding the new `quarto` file created on Windows, I don't think having a `sh` file in PATH is a good idea. It duplicates the `quarto.cmd` and it could only be run inside a bash environment. I get it is useful but we maybe open to some errors depending on the bash environment or the config. 
I get the the `quarto` which is a `sh` file will solves our issues when quarto is called in Git Bash or equivalent, but why not calling `quarto.cmd` for it so that we keep one way of running on Windows ? 

Something like 
````
#!/bin/sh
MSYS2_ARG_CONV_EXCL=/C cmd.exe /C  "<path to `quarto.cmd` here>"  "$@"
````

Here are first comments for discussion.

File: src/core/devconfig.ts [hunk 3]
```diff
@@ -119,16 +128,12 @@ export async function reconfigureQuarto(
...
-
   const quartoDir = Deno.realPathSync(
     join(quartoConfig.sharePath(), "..", ".."),
   );
 
   const process = Deno.run({
-    cmd: [configureScript],
+    cmd: ["bash", "configure.sh"],
     cwd: quartoDir,
   });
```
cderv, line 138: I wonder is this would cause an issue when reconfigure would be called outside of the context of Git bash or MSYS. 

`reconfigureQuarto()` is called in `quarto.ts` which I think could be called from `quarto.cmd` at some point. I did not manage to trigger this to test, but still trying to get a sense of if this could cause issue. 

It is just that calling `bash` in Powershell or CMD when WSL is installed will open WSL bash window - so run in Ubuntu. Not what we want I think

src/tools/impl/tinytex.ts
```diff
@@ -335,9 +335,8 @@ This will instruct TeX Live to create symlinks that it needs in <bin_dir_on_path
     const envPath = Deno.env.get("QUARTO_TEXLIVE_BINPATH");
     if (envPath) {
       paths.push(envPath);
-    } else if (Deno.build.os !== "windows") {
-      paths.push(...suggestUserBinPaths());
     } else {
+      paths.push(... await suggestUserBinPaths());
       paths.push(tlmgrPath);
     }
```
cderv, line 341: With this change, `binPathMessage` should be updated also below I think ? 

TBH I am not just why this need to be changed for Windows and if it works effectively.

We may end up with a different configuration when `quarto tools install tinytex` is run from Powershell or CMD, or run from Git bash. In the latter `~/bin` will be in PATH I think, but in the former it won't. So `suggestUserBinPaths()` will filter out from possible paths.

package/scripts/windows/quarto.cmd
msarahan: Yes, this is completely true, and we should not ever call bash from a .cmd script. You comment elsewhere about the reconfiguration stuff, and I'll answer more there. Suffice to say, this is a code path that is not handled very well in this PR, and yes, it is possible as it stands to call bash from cmd or powershell, resulting in a confusing failure to run the .sh script.

configure.cmd
```diff
@@ -1,103 +0,0 @@
...
-SETLOCAL ENABLEDELAYEDEXPANSION
-
-REM Batch script reminder: variables between % % are evaluated as parse time, not at runtime.
-REM    Variables set dynamically (including being read from a file) should be evaluated
-REM    using ! ! instead of % %. However, this only allows one level of expansion. Do not
-REM    try to create a variable that derives from a derived variable. It will be empty.
-
-REM First Check that Deno isn't running since this causes weird and confusing errors
-tasklist /fi "ImageName eq deno.exe" /fo csv 2>NUL | find /I "deno.exe">NUL
-if "%ERRORLEVEL%"=="0" goto :denoRunning
```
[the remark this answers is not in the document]
msarahan: Unfortunately, it probably is worthwhile. Thanks for catching that. The issue is that on Windows, you can't overwrite a file that is currently open. This interacts strongly with the whole "reconfiguration" scheme, and I'm not sure how that ever worked. As I understand it, if the deno program called back into the shell/batch script as a subprocess, then deno would always be running, and the reconfiguration would never have worked on windows. I need to go back and test that to understand how it worked (or didn't).

msarahan:
> Regarding the new quarto file created on Windows, I don't think having a sh file in PATH is a good idea. It duplicates the quarto.cmd and it could only be run inside a bash environment. I get it is useful but we maybe open to some errors depending on the bash environment or the config. I get the the quarto which is a sh file will solves our issues when quarto is called in Git Bash or equivalent, but why not calling quarto.cmd for it so that we keep one way of running on Windows ?

This is a really interesting thought. I see it as competing interests of uniformity. I was only considering uniformity across all platforms running one file, but there is also something to be said for uniformity of different files on one platform. I have no strong opinion on which should take priority. From a code perspective, I prefer uniformity across platforms, but I can see how uniformity on a platform could be more important from a user perspective.

package/conda-recipe/bld.bat
```diff
@@ -9,11 +9,8 @@ SET QUARTO_PACKAGE_PATH=%SRC_DIR%\package
 SET QUARTO_DIST_PATH=%LIBRARY_PREFIX%
 SET QUARTO_SHARE_PATH=%LIBRARY_PREFIX%\share\quarto
 
-call %SRC_DIR%\package\src\set_package_paths.bat
-
-call configure.cmd
-call package\src\quarto-bld.cmd update-html-dependencies
-call package\src\quarto-bld.cmd prepare-dist
+bash configure.sh
+bash package/src/quarto-bld.cmd prepare-dist
```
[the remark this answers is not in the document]
msarahan: yes. Apologies, I have been neglecting the conda recipe for the most part. I will fix it. Part of the reason for the neglect has been that the windows configuration stuff required some gross patch work, and I was going to circle around and fix that once this unix configuration stuff was in place on Windows.

src/tools/impl/tinytex.ts
msarahan: I'm fine to revert this. I think there's a larger issue at play here of Quarto acting as a package manager of sorts. It is a huge convenience, but perhaps also a quagmire. That's a discussion for a separate issue, and I got sidetracked here by finding this connection with suggestUserBinPaths.

src/core/devconfig.ts
msarahan: I would call this the Achilles heel of this PR, and this must be solved before this work can be viable. Thanks for raising this issue.

configure.cmd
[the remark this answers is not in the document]
cderv: I can confirm that the reconfiguring worked well when used in RStudio IDE where I am mostly using Quarto. 
So I did not encountered the issue with Deno LSP before the check was introduced. I can't say. 

I think it is worth adding back, but it means indeed that the reconfiguring won't work when using quarto inside VSCODE

package/conda-recipe/bld.bat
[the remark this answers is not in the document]
cderv: No worries. I just stumbled upon that while reading. I did not run and check the conda recipe at all. So probably / maybe other problems in there.

src/core/devconfig.ts
cderv: As we are trying to target dev mode only mostly (as only developer would run the configure.sh after a `git clone` and not use pre-release only for testing), maybe a simple workaround would be to ask for more configuration.

We could ask the Windows developer to set in its environment an env var to the Windows bash to use, and check this before trying a reconfiguration ? 

This is quite manual and would be part of prerequisite to opt-in the feature, but it would allow the script to call the right shell with the script. 

Does it make sense ? 

I am not use on Windows to run bash script really - I usually stay in BATCH or powershell.

configure.cmd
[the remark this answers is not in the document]
dragonstyle: Right, though I think the idea was that before this check, reconfiguring when working in VSCode on quarto didn't work anyway (since the Deno LSP would be running, making deno.exe `open`), but threw a confusing message without much help as to how to resolve. This at least tries to provide some clue (I added this as I only worked in windows occasionally and I would always get bitten by this and spend 15 minutes trying to debug why the error was happening before I realized I should just close vscode).

cderv:
> I was only considering uniformity across all platforms running one file, but there is also something to be said for uniformity of different files on one platform

I think I am considering that even when using a git bash or other MSYS bash on Windows, the quarto script for Windows should be run , which would be `quarto.cmd`. I feel than running the `quarto` bash script which is designed for Linux only because we are in a  bash emulation is not safe enough. I don't have a strong example of failure, but I know I add in the past some challenge using `git` in Git Bash or outside. So by experience, I am cautious 😅 

However, having a `quarto` file executing on bash is I think the missing piece to solve our issue on Git Bash found during conference (cc @dragonstyle ). We should have that in addition to `quarto.cmd`. That is why I am proposing the alternative of calling `quarto.cmd` from this file.

msarahan:
closing in light of the superior approach being done in https://github.com/quarto-dev/quarto-cli/tree/feature/rust-loader

cderv:
Oh ok... It goes quick ! Thanks for the reference. I'll follow there.

PR #2136, opened 2022-08-25
make quarto script POSIX-compatible. Add GH Action to check.

msarahan:
This is a follow-up to #2128. It makes the quarto script that we ship POSIX compatible and adds a Github Action to verify that it remains POSIX compatible.

I left in the paths to other shell scripts, but commented them. We may still want to run them through shellcheck, but with a less restrictive target shell (perhaps Bash).

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,48 @@
+name: Shellcheck
+
+on:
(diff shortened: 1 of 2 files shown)


msarahan:
closing in light of the superior approach being done in https://github.com/quarto-dev/quarto-cli/tree/feature/rust-loader

PR #2149, opened 2022-08-26
Upload tinitex artifacts

yihui:
so that we don't need to rebuild them in the rstudio/tinytex repo.

### DIFF
--- a/.github/workflows/test-quarto-latexmk.yml
+++ b/.github/workflows/test-quarto-latexmk.yml
@@ -59,3 +59,10 @@ jobs:
           ./quarto-bld compile-quarto-latexmk --target x86_64-unknown-linux-gnu --name tinitex
           ls -R ../dist/bin
           ../dist/bin/tinitex/x86_64-unknown-linux-gnu/tinitex --help
+
+      - name: Upload binary
+        uses: actions/upload-artifact@main
+        with:
+          name: tinitex
+          path: |
+            package/dist/bin/tinitex/*/tinitex*


cderv:
That is an excellent solution !  thanks ! 

We could then just download them and surface them in `tinytex-releases` Github release page. 

Would it make sense otherwise to add the `tinitex` binaries to each pre-release and release here on Github ? Probably better in their own  release at `tinytex-releases` repo though.

yihui:
I still consider `tinitex` experimental (my main concern is that the file size is too big), so I hope to "release" it in stealth.

I'm trying to build and upload it in this repo because I finally figured out why the build was failing in `rstudio/tinytex` in the past few days. It was due to the `setup-r` action, which might have changed `PATH`, causing a wrong `find` being used. I'm not sure what the exact problem is, but building `tinitex` *before* installing R worked okay.

Anyway, I don't have write access here. If you do, please feel free to merge it. Thanks!

cderv (approved):
> It was due to the setup-r action, which might have changed PATH, causing a wrong find being used. I'm not sure what the exact problem is, but building tinitex before installing R worked okay.

Interesting 🤔  nothing come to mind of what could have changed. 

> Anyway, I don't have write access here. If you do, please feel free to merge it. Thanks!

I have but usually only do PR and rely on @dragonstyle to merge things regarding github actions. I'll let him do the merge, especially because I don't yet fully comprehend where they merge stuff exactly at which step of the release process. (which they are fully in right now). 

@dragonstyle you can merge when you're done.

dragonstyle:
Looks great!

yihui:
@dragonstyle Thanks!

@cderv Could you update https://github.com/rstudio/tinytex/blob/main/.github/workflows/build.yaml accordingly when you have time? Thanks!

cderv:
Sure !

PR #2274, opened 2022-09-02
ensure external links opened in separate window when document rendered in iframe

kevinushey:
This PR ensures that, for Quarto documents previewed to HTML which are hosted within an iframe, that external links are opened with an external browser, rather than having an attempted navigation occur within the iframe.

Tested with https://github.com/jjallaire/hopr.

### DIFF
--- a/src/command/preview/preview.ts
+++ b/src/command/preview/preview.ts
@@ -592,11 +592,11 @@ function htmlFileRequestHandlerOptions(
           file = format.formatPreviewFile(file, format);
         }
         const fileContents = await Deno.readFile(file);
-        return reloader.injectClient(fileContents, inputFile);
+        return reloader.injectClient(req, fileContents, inputFile);
       } else if (isTextContent(file)) {
         const html = await textPreviewHtml(file, req);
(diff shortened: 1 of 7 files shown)


jjallaire:
I think we should talk over the `origin` and `search` variables realtime just so I understand 100% and so that there aren't use cases you aren't aware of that might make these not work. Let's plan on doing this on Tuesday.

kevinushey:
For context, the intention here is to capture the URL used by the embedding application itself when serving the Quarto document / site, since (as users navigate around the rendered site) the URL will change and become a "plain" localhost URL (without, for example, the `capabilities=` adornment that RStudio provides) and so we'd fail to inject the devserver scripts once the user starts navigating.

That said, it may be cleaner to handle this as part of the `devhost-init` contract. Perhaps the `event.origin` here would already have what we need, and we could store those values as part of the `quartoDevhost` object?

https://github.com/quarto-dev/quarto-cli/blob/058643e2e9520616bdcdb3c06364ce526867370b/src/resources/editor/devserver/devserver-viewer.html#L6-L16

PR #2298, opened 2022-09-05
Configure the Github Issue chooser

cderv:
To add some more choices when a user go to the issue workflow

Doc about this Github feature in https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

The proposal is to have something like in **rmarkdown** (https://github.com/rstudio/rmarkdown/issues/new/choose) 
image

or in RStudio IDE (https://github.com/rstudio/rstudio/issues/new/choose)
image

Proposal: 

* Issues are opened in the repo following the template
* Link to issue guide is shared from the chooser so that user know where to find it even before opening an issue
* Questions should be opened to https://github.com/quarto-dev/quarto-cli/discussions/categories/q-a so we share a link to the category directly
* Any type of feedback can be shared in our Dicussion board, and the last link point to Discussion. 

The two last could be merged in one link.

`blank_issues_enabled: false` means it is not possible to open a blank issue - I think it is valid as we allow discussion board for anything so that if someone does not want to open a Bug, then it will be a new discussion (we could convert to issue). 

This is a first proposal to have a more welcoming page when someone click on "New Issue" in this repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,5 +1,5 @@
 name: Bug report
-description: Report an error or unexpected behavior you saw while using this package
+description: Report an error or unexpected behavior
 labels: [bug]
 
 body:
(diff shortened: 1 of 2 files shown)


jjallaire:
Awesome, thank you!!!!

PR #2552, opened 2022-09-22
Typo in `self-contained` YAML description

### LINKED ISSUE #406: Spelling Error: decrecated
image

https://quarto.org/docs/reference/formats/html.html#rendering

cderv:
I believe this is the source where all comes from. 

Especially this should close quarto-dev/quarto-web#406 once we regenerate all the files. 

I am not sure of the all workflow though. 

Is this correctly the only file to modify ? Are the update to JS files with reference description automatic  ? 

Also, are the files in quarto-web correctly generated from this YAML ? 

Hopefully I got it right.

### DIFF
--- a/src/resources/schema/document-render.yml
+++ b/src/resources/schema/document-render.yml
@@ -72,7 +72,7 @@
     short: "Produce a standalone HTML file with no external dependencies"
     long: |
       Produce a standalone HTML file with no external dependencies. Note that
-      this option has been decrecated in favor of `embed-resources`.
+      this option has been deprecated in favor of `embed-resources`.
 
 - name: self-contained-math
   tags:


dragonstyle:
Since this is an update to our YAML schema (which is used to generate the reference docs, help, and so on), when you make changes to these files you need to run `quarto build-js` from the `quarto-cli` directory to build the js from the updated schema. (those updated files should be included in this PR).

To rebuild the reference docs in `quarto-web` I think that you need to run `quarto run tools/reference.ts` from the `quarto-web` directory, which will regenerate the reference doc pages from the revised schema... happy to jump on a zoom to help if you want!

ghost (approved):

cderv:
> when you make changes to these files you need to run `quarto build-js` from the quarto-cli

I was missing this. @allenmanning did you run it already ? I don't see it. 

> rebuild the reference docs in quarto-web I think that you need to run quarto run tools/reference.ts from the quarto-web

And something new I learn ! Seems it was not run either ? 

Do you plan to do that @allenmanning  ? 

Just to clarify as you merged already.

cderv:
@dragonstyle  this is indeed `quarto run ./tools/reference.ts` run, but this cannot be done right now as it is out of date with current schemas. I opened https://github.com/quarto-dev/quarto-web/issues/407 to track.

PR #2954, opened 2022-10-20
Fix regression caused by listing warnings

### LINKED ISSUE #367: Listing of non-qmd files
The listing function is really powerful! I wonder if it is possible to make `listing` work also for non-qmd files.

This is helpful to create a list of links for downloadable files, such as pdf, csv, or image files. I tried with a list a csv files by using the `filename` as the field name. It did not work out of the box. 

I can imagine that this use case is quite common.

dragonstyle:
This re-fixes #367

My implementation of warning regarding empty qmd files was way too aggressive, scale it back.

### DIFF
--- a/src/project/types/website/listing/website-listing-read.ts
+++ b/src/project/types/website/listing/website-listing-read.ts
@@ -6,7 +6,7 @@
 *
 */
 import { debug, warning } from "log/mod.ts";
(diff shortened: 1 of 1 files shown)


cscheid (approved):
LGTM.

PR #3085, opened 2022-10-28
Fix table top border rendering for html

### LINKED ISSUE #3084: Top border table does not get rendered without `colgroup`
### Bug description

When tables get rendered without a `colgroup` element (so whenever columns do not get a specific size), the border at the top of the table doesn't show. For example, in a new website with this `index.qmd`:

``` md
---
title: "Test"
---

| a | a | a |
| - | - | - |
| a | a | a |

| aaaaaaaaa  | aaaa   | aaaaaaaaaaa |
| -          | -      | ---         |
| aaaaaaaaaaaaa | aaaaaaaaaaaa | aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa |

| aaaaaaaaa  | aaaa   | aaaaaaaaaaa |
| -          | -      | ---         |
| aaaaaaaaaaaaa | aaaaaaaaaaaaaaaa | aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa |
```

only the third table's top border is rendered. This is because the top border CSS rule is defined as `.table>:not(:first-child)` and the two first tables do not have the `colgroup` element (I think this is Pandoc?).

I'm using Quarto v1.2.242.

### Checklist

tomvaneyck:
As explained in issue #3084, the top border of tables does not always render correctly in html because of the way the html is generated by Pandoc and the css rules defined for tables. This pull request fixes #3084 and simplifies the border rules.

I don't know if this bug fix is large enough to require a signed contributor agreement. Please let me know if I have to file one!

### DIFF
--- a/src/resources/formats/html/bootstrap/_bootstrap-rules.scss
+++ b/src/resources/formats/html/bootstrap/_bootstrap-rules.scss
@@ -1060,11 +1060,9 @@ div.hanging-indent {
   word-break: break-word;
 }
 
-.table {
-  border-top: 1px solid $table-border-color;
-}
-
 .table > thead {
+  border-top-width: 1px;
+  border-top-color: $table-border-color;
   border-bottom: 1px solid $table-group-separator-color;
 }
 
--- a/src/resources/formats/html/bootstrap/_bootstrap-rules.scss
+++ b/src/resources/formats/html/bootstrap/_bootstrap-rules.scss
@@ -1060,17 +1060,12 @@ div.hanging-indent {
   word-break: break-word;
 }
 
-.table > :not(:first-child) {
-  border-top-width: 1px;
-  border-top-color: $table-border-color;
+.table {
+  border-top: 1px solid $table-border-color;
 }
 
 .table > thead {
   border-bottom: 1px solid $table-group-separator-color;
-}
-
-.table > tbody {
-  border-top: 1px solid $table-border-color;
 }
 
 @if $code-block-border-left {


cscheid:
Thanks for the contribution!

We'll take a closer look at this soon. The situation is not as simple as you'd expect because we need to make sure your suggestion works across a number of possible ways to generate tables. (Right now, the quarto table situation is not as clean as we'd prefer.)

For example, we have tables coming from:

- raw html
- simple markdown tables like you have
- grid tables
- from computational output:
  - `r`: tables from knitr::kable, flextable, dt, gt
  - `python`: tables from jupyter, pandas
 
We need to check that this change doesn't introduce regressions. In 1.3 we're hoping to create an API that table libraries will target, which will make things easier.

cscheid:
Hi, thank you for you contribution again and sorry for the delay. I've created a test document for all the tables which we'd like to look consistent (if we're going to touch this styling code).

You can see the file here: https://github.com/quarto-dev/quarto-cli/blob/main/tests/docs/smoke-all/2023/02/01/3085-r.qmd

I've annotated the document with `good!` and `bad!` subsections that illustrate our desired behavior on `main`.

Unfortunately, it looks like the changes in your PR change the style more than we'd like.

## On `main`

image

## On this PR:

image

Note a couple of things:

- the rendering is still inconsistent: on `main` we are missing the light top border in the `bad` tables, but on this PR we get a thick dark top border in the `good` tables.
- the style of the rendering changed, and now some of the borders are thicker. We'd like that not to change.

Once again, sorry for the delay in responding this. If you want to go ahead and try to correct this, I'll be happy to take a look (and we're working on tables right now, so my turnaround time should be much quicker!)

tomvaneyck:
I've had another go, and I got your example file to render correctly. However, there is some weird behavior with the old rules (for some reason in the final css file, there were two `.table > :not(:first-child)` rules, although I could only find one that had any influence in the source files. But then again, I would think that these rules were there for a reason.

In any way, I fixed it by defining the border based on the `thead` element alone, so hopefully this is covers all cases.

cscheid:
Thanks for the update! @dragonstyle I think you authored our css here, so do you mind giving this a quick review?

dragonstyle (approved):

PR #3429, opened 2022-11-22
Feature/notebooks

dragonstyle:


src/format/formats.ts
```diff
@@ -208,12 +261,15 @@ export function defaultWriterFormat(to: string): Format {
...
+  // Set the originating to
+  writerFormat.identifier[kTargetFormat] = to;
+
   // return the createFormat
   return writerFormat;
 }
 
 function powerpointFormat(): Format {
-  return createFormat("pptx", {
+  return createFormat("pptx", "Powerpoint", {
```
cderv, line 272: @dragonstyle I believe this should be in the other order right ? 
This now creates file with Powerpoint extension... not good 😓 

Found that try to build a pptx with dev version, and it led me to this PR.

cderv: I opened #3691 to fix this

PR #3540, opened 2022-12-01
[chores] Add id to invalidate cache easily in smoke test

cderv:
This add an id to invalidate the cache easily. We just need to increase the id when needed. 

We could also add a the OS version info into the cache key, but this would be one more step in the workflow to compute the information, and then use it in the key. If we want to do that, then this action can be useful (https://github.com/marketplace/actions/github-actions-runner-os-system-information) to use or for inspiration on how to retrieve.

### DIFF
--- a/.github/workflows/test-smokes.yml
+++ b/.github/workflows/test-smokes.yml
@@ -48,9 +48,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.RENV_PATHS_ROOT }}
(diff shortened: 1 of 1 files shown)


cderv:
So this is enough to fix the issue. 

If you think that worth it to be more precise and invalidate on the OS specifically, we can do it probably. Otherwise, we will just need to increment the id whenever we need to invalidate the cache manually.

PR #3993, opened 2023-01-17
Feature/site nesting

ghost:
## Description

- Editor v2 for Confluence
- Nested folder support for Confluence publish.
  - Limited support for non-windows https://github.com/rstudio/quarto-confluence-test/issues/36 

## Checklist

I have (if applicable):

- [x] referenced the GitHub issue this PR closes https://github.com/rstudio/quarto-confluence-test/issues/28


src/publish/confluence/confluence-helper.ts
```diff
@@ -33,6 +34,10 @@ export const FILE_FINDER: RegExp = /(?<=href=\')(.*)(?=\.qmd)/;
 const IMAGE_FINDER: RegExp =
   /(?<=ri:attachment ri:filename=["\'])[^"\']+?\.(?:jpe?g|png|gif|m4a|mp3|txt)(?=["\'])/g;
 
+const capitalizeFirstLetter = (value: string): string => {
+  return value[0].toUpperCase() + value.slice(1);
```
dragonstyle, line 38: Do you need protection against empty strings anywhere in the stack (not sure if this already exists when called, so just adding question as sanity check)

src/publish/confluence/confluence-helper.ts
```diff
@@ -420,32 +508,66 @@ export const updateLinks = (
...
+      const linkFullFileName = `${linkFileName}.qmd`;
+
+      let siteFilePath = linkFullFileName;
+      const isAbsolute = siteFilePath.startsWith("/");
+      if (!isAbsolute && docFileNamePathList.length > 1) {
+        const relativePath = docFileNamePathList
+          .slice(0, docFileNamePathList.length - 1)
+          .join("/");
+
+        if (siteFilePath.startsWith("./")) {
```
dragonstyle, line 541: Just FYI, There is a path `join` function you can use to join paths that don't include platform specific separators (perhaps this is a link so doesn't matter).

src/publish/confluence/confluence-helper.ts
```diff
@@ -486,9 +608,65 @@ export const updateImagePaths = (body: ContentBody): ContentBody => {
   return body;
 };
 
-export const findAttachments = (bodyValue: string): string[] => {
+export const findAttachments = (
+  bodyValue: string,
+  publishFiles: string[] = [],
+  filePath: string = ""
+): string[] => {
+  const pathList = filePath.split("/");
```
dragonstyle, line 616: Another file path question - if this were on a windows system would this work? There is a `dirname` which I believe would do this without regard to path separators if that is needed.

dragonstyle: (If this is a change you think should be made, there are other similar patterns that I haven't highlighted since I don't want to make unnecessary noise if it is ok)

src/publish/confluence/confluence-helper.ts
allenmanning: Thank you, I think I need to make that change.

src/publish/confluence/confluence-helper.ts
allenmanning: It may be wise for us to talk about this offline.  I think I can do a better job with my path work here.

src/publish/confluence/confluence-helper.ts
allenmanning: Fixed.  Thank you.

src/publish/confluence/confluence-helper.ts
allenmanning: I need to research Deno support for resolving platform specific paths using the link address.  This is for both links and images.  I've created this to test and fix.  https://github.com/rstudio/quarto-confluence-test/issues/36

src/publish/confluence/confluence-helper.ts
allenmanning: https://github.com/rstudio/quarto-confluence-test/issues/36

PR #4090, opened 2023-01-25
Confluence: 2-Pass update for link resolution in Confluence

ghost:
Confluence: 2-Pass update for link resolution in Confluence

#fixes https://github.com/rstudio/quarto-confluence-test/issues/37

### DIFF
--- a/src/publish/confluence/confluence-helper.ts
+++ b/src/publish/confluence/confluence-helper.ts
@@ -632,6 +632,37 @@ export const updateLinks = (
   return { pass1Changes: updatedChanges, pass2Changes: collectedPass2Changes };
 };
 
+export const convertForSecondPass = (
+  fileMetadataTable: Record<string, SitePage>,
+  spaceChanges: ConfluenceSpaceChange[],
+  server: string,
+  parent: ConfluenceParent
+): ConfluenceSpaceChange[] => {
+  const toUpdatesReducer = (
+    accumulator: ConfluenceSpaceChange[],
+    change: ConfluenceSpaceChange
+  ) => {
+    if (isContentUpdate(change)) {
+      accumulator = [...accumulator, change];
+    }
+
+    if (isContentCreate(change)) {
+      const convertedUpdate = buildContentUpdate(
+        "fake-id-fixme",
+        change.title,
+        change.body, //TODO convert link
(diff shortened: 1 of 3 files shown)


src/publish/confluence/confluence-helper.ts
```diff
@@ -621,7 +629,58 @@ export const updateLinks = (
...
+  const toUpdatesReducer = (
+    accumulator: ConfluenceSpaceChange[],
+    change: ConfluenceSpaceChange
+  ) => {
+    if (isContentUpdate(change)) {
+      accumulator = [...accumulator, change];
+    }
+
+    if (isContentCreate(change)) {
+      const qmdFileName = change?.fileName?.replace(".xml", ".qmd") ?? "";
```
dragonstyle, line 650: Just to play more file system defense here you could consider using base name to prevent a file name like `foo.xml.xml` from messing with you.

src/publish/confluence/confluence-helper.ts
```diff
@@ -597,7 +603,9 @@ export const updateLinks = (
 
         updated = updated.replace(linkFullFileName, pagePath);
       } else {
-        console.warn(`Link not found for ${siteFilePath}`);
+        if (!collectedPass2Changes.includes(changeToProcess)) {
```
dragonstyle, line 606: If you used `collectedPass2Changes.push(changeToProcess);` here, you could get away with making `collectedPass2Changes` a const above.

dragonstyle (approved):

src/publish/confluence/confluence-helper.ts
allenmanning: Thanks for the suggestion, stylistically I prefer clean mutation of lets rather than "hidden" mutation of consts.  I need to think more about this though to be honest, I don't think it is quite there.

allenmanning: After a second look, I think I'm going to keep it as this is updating the let in the closure - pretty non-standard.  Thank you again for the idea.

src/publish/confluence/confluence-helper.ts
allenmanning: Adding test, using `basename` now thanks again.

PR #4175, opened 2023-02-01
Confluence:  task/upgrade to new writer

ghost:
Upgrade writer to the new Pandoc format.


src/publish/confluence/constants.ts
```diff
@@ -1,6 +1,6 @@
 export const DELETE_DISABLED = false;
 
-export const EXIT_ON_ERROR = true;
+export const EXIT_ON_ERROR = false;
```
dragonstyle, line 3: I'm not super well informed about what this means, but I generally vote for failing hard / exit on error by default so we find any unexpected errors quickly / don't proceed with operations once the outcome / state become indeterminant. This might be an expected error or something, so feel free to ignore peanut gallery if not helpful.

src/resources/extensions/quarto/confluence/publish-classic.lua
```diff
@@ -0,0 +1,284 @@
+-- Confluence Storage Format for Pandoc
+-- https://confluence.atlassian.com/doc/confluence-storage-format-790796544.html
+-- https://pandoc.org/MANUAL.html#custom-readers-and-writers
+
+local stringify = (require "pandoc.utils").stringify
+local confluence = require('overrides')
+
+-- From https://stackoverflow.com/questions/9168058/how-to-dump-a-table-to-console
+function dumpObject(o)
```
dragonstyle: Does this have access to the `quarto` namespace? If so could also use `quarto.log.output()` which will print objects or could use the `quarto.utils.dump()` which is older school version of `output()`...

dragonstyle (approved):
Some feedback mainly to be sure you're aware of some possible other options in the codebase but otherwise looks awesome!

src/publish/confluence/confluence.ts
```diff
@@ -562,8 +562,17 @@ async function publish(
       doOperation = async () =>
         (content = await createContent(publishFiles, body));
     }
+    try {
+      await doWithSpinner(message, doOperation);
+    } catch (error: any) {
+      console.info("Error Performing Operation");
```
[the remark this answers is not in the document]
allenmanning: Yes.  I really need to be using that, thank you.

src/publish/confluence/constants.ts
allenmanning: It is worth thinking about.  This is mainly for my testing today.  In production, it seems wiser to fail hard rather than let run-away operations go.  It is a tough call actually.  I need to think more on this.  Thank you for raising it.

PR #4253, opened 2023-02-07
Confluence:  Footnote Support

ghost:
#fixes https://github.com/rstudio/quarto-confluence-test/issues/53

I reviewed the approach with @jjallaire as this could be done in either the Custom Writer or in the publish code.  I decided to go with the publish approach as it seemed like a straight-replace.


src/resources/extensions/quarto/confluence/publish-classic.lua
```diff
@@ -0,0 +1,284 @@
+-- Confluence Storage Format for Pandoc
```
allenmanning, line 1: Added this for historical compare reasons.

jjallaire (approved):

PR #4255, opened 2023-02-07
Do not error on file existence when filtering resources

cderv:
**THIS PR IS AGAINST WINDOWS TESTS BRANCH**

Local resources files won't be found as `images/beach.jpg?bar` is not a file. It errors on Windows (not on Linux). 

Here I tweak test, as I don't think this is something we should really supports. Should we ? 

I mean 
```markdown
![Image](images/beach.jpg?bar)
```
points a no real file and self contained will fail because image detected is not `beach.jpg` but `beach.jpg?bar`

If such local URL needs to be supported somehow, then we should parse resources by removing URL parameters. 

Just want to check that before committing. 

What are your thought @dragonstyle  ?

### DIFF
--- a/src/command/render/resources.ts
+++ b/src/command/render/resources.ts
@@ -6,9 +6,12 @@
 */
 
 import { dirname, join } from "path/mod.ts";
-import { existsSync } from "fs/mod.ts";
-
-import { normalizePath, ResolvedPathGlobs, resolvePathGlobs } from "../../core/path.ts";
+import {
+  normalizePath,
+  ResolvedPathGlobs,
+  resolvePathGlobs,
+  safeExistsSync,
+} from "../../core/path.ts";
 import { engineIgnoreGlobs } from "../../execute/engine.ts";
 import { kQuartoScratch } from "../../project/project-scratch.ts";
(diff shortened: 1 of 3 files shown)


dragonstyle:
It does seem like local paths do not need to support query params in this way- I think it's fine to remove this from the test. This does, to me, seem like a possible place to use safeExistsSync() since our point of view is that this couldn't be a valid file path (not that it should throw). Theoretically, someone could likely form a valid relative path to a file that will be present when deployed on a server which contains a query param which would be properly interpreted in the context of deployment (making the path a valid href, but not a valid local file path)...

cderv:
> Theoretically, someone could likely form a valid relative path to a file that will be present when deployed on a server which contains a query param which would be properly interpreted in the context of deployment (making the path a valid href, but not a valid local file path).

yes that was my thinking too. This would work with Pandoc I guess only with `self-contained: false` right ? 

Otherwise, Pandoc will try to insert something to encode and failed.

cderv:
So I made the modification. I believe this will "solve" also https://github.com/quarto-dev/quarto-cli/issues/4194#issuecomment-1420526610 by hiding the problem there. 

The is the issue with `safeExistsSync` I gues, we'll hide so errors

cderv:
So I will merge this but it will hide the other issue. We just need to remember.

PR #4282, opened 2023-02-08
proivide smarter warning on windows

### LINKED ISSUE #4281: QUARTO_DENO wrongly set internally in `quarto.exe` on Windows ?
On Windows 
````
> quarto --version
1.3.184
> echo 'console.log("hello")' > test.js
> quarto run test.js
WARNING: Specified QUARTO_DENO does not exist, using built in deno
hello
> $ENV:QUARTO_DENO
````

See the warning but `QUARTO_DENO` is not set on the system. Something is off. 

@dragonstyle I need your help here as I don't know exactly how we are bundling the `.exe`. I know we set `QUARTO_DENO` on the `quarto.cmd` file and maybe this is still the case in the `.exe` and path is not correct ? 

Logging here to not forget.

dragonstyle:
Fixes #4281

### DIFF
--- a/src/core/resources.ts
+++ b/src/core/resources.ts
@@ -40,7 +40,19 @@ export function toolsPath(binary: string): string {
   const binaryPath = Deno.env.get(binaryEnvKey);
   if (binaryPath) {
     if (!existsSync(binaryPath)) {
-      displayWarning();
+      // If this is windows, we shouldn't warn if there is an 'exe' version of the path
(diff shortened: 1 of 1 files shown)


cderv (approved):
Yes I believe this should remove this necessary warning. 

I did not check with other tools that are used with this function though, but maybe `exe` is not the only extension. Example: `tlmgr.bat` on Windows but I don't think tlmgr goes through.  

So LGTM. thank you !

PR #4335, opened 2023-02-13
fix [Confluence] Linked Captioned Images not rendering #14

ghost:
#fixes https://github.com/rstudio/quarto-confluence-test/issues/14


src/resources/extensions/quarto/confluence/publish.lua
```diff
@@ -18,6 +18,23 @@ function log(label, object)
   print(label or '' .. ': ', dumpObject(object))
 end
 
+local function inlineOrInlinesToString(inlineOrInlinesValue)
```
allenmanning, line 21: @dragonstyle I'm not sure if this is the correct approach.  I can have an `Inline` or an `Inlines` (list of `Inline`) - I'm manually detecting and rendering out to string.  It works but I'm not sure if this the cleanest technique.

dragonstyle: I could be missing the overall goal here, but a couple things jump to mind for me:

1) I think `link.contents` is always a list of inlines, so you'd probably be safe to only support inlines unless this is used elsewhere.

2) I think this approach will actually return the literal string in the link contents, not actually rendering any markdown that might appear in the contents. Thinking of a link like:

```markdown
[**Hello there Google!**](https://www.google.com)
```

An approach to a similar problem that I used for `asciidoc` was to actually just render the contents before I need them:

```lua
-- render the contents of an admonition to asciidoc so I can use it directly
local admonitionContents = pandoc.write(pandoc.Pandoc(el.content), "asciidoc")
```

I'm not sure what confluence supports as far as links and their contents so this all could be pointless, but just want to throw it out there in case its relevant.

dragonstyle (commented):
I'm not sure if being in a Lua writer is confounding to the suggestion I made for dealing with link contents (haven't really ever done a write myself!) but take a look at suggestion (and especially that example link) and see whether all looks good. If so, lgtm!

src/resources/extensions/quarto/confluence/publish.lua
allenmanning: I knew that I was off the mark here.  Yep, I need to render out the contents first.  Let me run with this a bit and propose something new.  Thank you for the review.

allenmanning:
I have more work to do here.

PR #4574, opened 2023-03-01
reveal: add slide-title-fit-text option

### LINKED ISSUE #4530: If slide title is long, its font should be smaller by default
Suppose you have the following slide

````markdown
---
format: revealjs
---

## This is a veeeeeeeeeeeeeeeeerrrrryyyy loooooooooooooooooooooooooooonnnnngggggg title

```{r}
plot(cars, main = "Title is taking up a huge amount of space so the plot is tiny")
```
````

The result is a slide with a tiny plot and giant title:

Screenshot 2023-02-24 at 6 45 51 PM

Applying `.r-fit-text` to slide title solves the issue as it results in behaviour that is like PowerPoint or Keynote:

````markdown
---
format: revealjs
---

## [This is a veeeeeeeeeeeeeeeeerrrrryyyy loooooooooooooooooooooooooooonnnnngggggg title]{.r-fit-text}

```{r}
plot(cars, main = "Title is taking up a huge amount of space so the plot is tiny")
```
````

Screenshot 2023-02-24 at 6 46 50 PM

It would be nice if `.r-fit-text` behaviour was default, and if so, we should add something to the docs about how to disable it at https://quarto.org/docs/presentations/revealjs/advanced.html#fit-text.

jjallaire:
As discussed in last week's team meeting, this option is enabled by default and automatically fits slide title text onto a single line using the `r-fit-text` class.

### DIFF
--- a/news/changelog-1.3.md
+++ b/news/changelog-1.3.md
@@ -47,6 +47,7 @@
 - Authors on the title slides now correctly object customization of the `$presentation-title-slide-text-align` scss variable ([#3843](https://github.com/quarto-dev/quarto-cli/issues/3843))
 - Properly support `show-notes: separate-page` [#3996](https://github.com/quarto-dev/quarto-cli/issues/3996)
 - Improve footnote / aside layout for centered slides. [#4297](https://github.com/quarto-dev/quarto-cli/issues/4297)
+- Add `slide-title-fit-text` option (enabled by default) which automatically fits slide title text onto a single line
 
 ## Dates
 
--- a/news/changelog-1.3.md
+++ b/news/changelog-1.3.md
@@ -46,8 +46,10 @@
 - Properly support scss imports for RevealJS extensions ([#3414](https://github.com/quarto-dev/quarto-cli/issues/3414))
 - Authors on the title slides now correctly object customization of the `$presentation-title-slide-text-align` scss variable ([#3843](https://github.com/quarto-dev/quarto-cli/issues/3843))
 - Properly support `show-notes: separate-page` [#3996](https://github.com/quarto-dev/quarto-cli/issues/3996)
-- Improve footnote / aside layout for centered slides. [#4297](https://github.com/quarto-dev/quarto-cli/issues/4297)
 - Add `slide-title-fit-text` option (enabled by default) which automatically fits slide title text onto a single line
+- Properly support `show-notes: separate-page` ([#3996](https://github.com/quarto-dev/quarto-cli/issues/3996)).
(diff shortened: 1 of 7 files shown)


cscheid (approved):
I tested this (minimally), and it seems to work great.

I made a trivial change to the changelog, but otherwise this LGTM!

cderv (commented):
I was not there last week for discussion but this will apply to **all** headings, those for title-slide and also those for "normal" slides. 

Don't we want to target only `.title-slide` here ? or really all slides ? 

I ask because using fit text means that we will reduce long title but also make short title a lot bigger ! If there is no much content in the slide it will be huge. 

Example I tried 

````markdown
---
title: test
format: revealjs
---

# Slide title

## Content slide 

Content

# VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY LONG

## Short 

Content

## VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY LONG

Content

# VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY LONG

With sub content on title slide 

## Short 

Content

## VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY VERY LONG

Content

# Short

With sub content 

## Short

Content
```` 

Showing only screenshot for what happens to short title (as for long text it is ok)

* slide title with not long title will be bigger
image

* and also for normal slide with few content, it will be huged

image

Without it the config (setting `slide-title-fit-text: false`), this is what it looks like - seemed fine. Do we really prefer the new behavior ?

* Normal size for slide title
image

* And normal size for the second above 

image

I missed the discussion, but something does not feel right here to me especially for "normal slide" (without `.title-slide` class)

jjallaire:
I think I missed here then. My intention was to only target normal slides (not title slides, which often wrap and have no issue w/ wrapping). Further, I definitely didn't want to make smaller titles bigger! 

I'm thinking we should NOT merge this PR and instead in v1.4 implement our own version of this which never grows the text but only shrinks it (and also is overridable on a per-slide basis with a class).

@cderv Does that sound right to you?

cderv:
Yes it sounds definitely right. reveal fitting behavior does not seems right here - it seems to be made for making stuff as big as possible 
> The r-fit-text class makes text as large as possible without overflowing the slide. This is great when you want BIG text without having to manually find the right font size.

They use [fitty](https://github.com/rikschennink/fitty) and it is possible that we can do our own version for scale down text to big (and don't touch what already fits)

Let's do this for 1.4. Was there already an issue related to this PR so that we can track ?

allenmanning:
Greetings, I'm just checking in on this.  Shall we close this PR?

cderv:
Definitely for 1.4 as we need to come up with another solution. We could probably close this

Issue #4473, opened 2023-02-23
Create Project Scaffold `quarto create project`


allenmanning:
Ok sounds great! A few places to look:

1) The entry point to the `quarto create project` type command is in `src/command/create/cmd.ts`. This supports creating different types of artifacts, so you’ll be most interested in project artifacts.

2) You can see how project artifacts are create in `src/command/create/artifacts/project.ts`.

3) I just made a branch `feature/conf-proj` to show how to scaffold in a new project template for confluence - it was less clean than is ideal since it is mix of the new `quarto create` command and the old infrastructure that supported a legacy `quarto create-project` command.

4) You can basically just head to `src/project/types/project-default.ts` and implement some conditionality to the create function based upon the template (you can see what is happening to produce the default template below).

From JJ:
One other thing in the sample site we should do is have the index page provide a basic hint about workflow (e.g. run quarto preview locally then when you are ready to publish use quarto publish (or something like that, likely with a link to the docs)

allenmanning:
@cwickham We now have a basic project scaffold via `quarto create project confluence`.  It will create a simple example project that you can use to get you started.

PR #4667, opened 2023-03-07
Create Project Scaffold quarto create project #4473

ghost:
#4473


src/project/types/confluence/confluence.ts
```diff
@@ -0,0 +1,83 @@
+import { ProjectScaffoldFile } from "../types.ts";
+
+export const buildConfluenceFiles = (): ProjectScaffoldFile[] => {
+  const file: ProjectScaffoldFile[] = [
+    {
+      name: "project-roadmap",
+      content: `# Welcome
+Welcome to this sample Quarto Confluence Project! Feel free to use this as a way to help get you started in building your own Confluence project.
```
allenmanning, line 8: @jjallaire Would you like to see some changes here?

src/project/types/project-default.ts
```diff
@@ -1,44 +1,58 @@
...
-    const resourceDir = resourcePath(join("projects", "default"));
     return {
       configTemplate: join(resourceDir, "templates", "_quarto.ejs.yml"),
       resourceDir,
       scaffold: (engine: string, kernel?: string, packages?: string[]) => {
-        const file: ProjectScaffoldFile[] = [{
-          name: title,
-          content: `## Quarto
+        if (template === kConfluence) {
+          return buildConfluenceFiles();
```
allenmanning, line 44: @dragonstyle Would you prefer that I refactored this out in a different way?

src/project/types/confluence/confluence.ts
jjallaire: Maybe something more like:

"This project provides provides a simple scaffold for creating a Quarto Confluence Project. You'll almost certainly want to remove the sample files, they are just here as examples. When you add you own documents (including ones in subfolders) they will be automatically added to the site navigation sidebar."

jjallaire: Also, I'm thinking that we'd want this content in the `index.qmd` ? (let's discuss realtime in our 1:1 today)

PR #4676, opened 2023-03-07
Fix Pdf Export mode menu button and keyboard shortcut

### LINKED ISSUE #2988: Reveal.js PDF Export menu item is broken
In reveal.js presentations, the menu item for the PDF Export tool triggers the overview mode rather than a PDF export.

Looking at what I believe is the respective code, it appears to be a simple oversight and probably straightforward to fix:
https://github.com/quarto-dev/quarto-cli/blob/7cc0ced8c9d62a8bc6a3ed8863160f90729b77bf/src/format/reveal/format-reveal-plugin.ts#L438-L442

Weirdly, though, the key binding "e" still works as expected.

cderv:
This fix #2988

We patch RevealJS plugin PdfExport so that the toggle function is exported. This allow us to assign it as handler to our menu logic and keyboard shortcut. 

To make things clean, the patching of external library is done using a Git Patch, so I added support for that in update-html-dependencies logic. 

This PR also adds a prettier configuration file for ignoring some file from the VSCODE auto formatting as we don't want to touch those in our commit tree since they are external. 

A mixed PR but all related to the bug fix.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,8 @@
+src/resources/formats/revealjs/plugins/line-highlight/line-highlight.js
+src/resources/formats/revealjs/plugins/menu/menu.js
+src/resources/formats/revealjs/plugins/menu/quarto-menu.js
+src/resources/formats/revealjs/plugins/multiplex/multiplex.js
+src/resources/formats/revealjs/plugins/multiplex/socket.io.js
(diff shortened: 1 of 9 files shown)


dragonstyle (approved):
Love that you automated the patch process!

PR #4691, opened 2023-03-08
respect QUARTO_PANDOC in update-pandoc

msarahan:
## Description

This is a patch from the conda package. This particular code block does not respect QUARTO_PANDOC, and thus can fail when using pandoc other than the standard distribution. I am submitting this against 1.2.x because that's what the conda package builds right now. If you'd prefer it to be only on newer work, let me know. It's fine to remain a patch on the conda package for as long as needed.

## Checklist

I have (if applicable):

### DIFF
--- a/news/changelog-1.2.md
+++ b/news/changelog-1.2.md
@@ -1,6 +1,9 @@
 ## New In This Patch Release
 
 - Fixes an issue using Quarto with the most recent version of jupyter_client (version 8.0.0 and later)
+- Fix QUARTO_PANDOC variable not being respected in update-pandoc process. This
+  would cause errors where pandoc is not bundled with Quarto, such as in the
+conda package.
 
 ## Fixed In Previous Patch Releases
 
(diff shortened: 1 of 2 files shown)


dragonstyle:
Sorry I'm late getting to this. I think we explicitly do not want this script to respect that env variable - this is explicitly a build tool that updates the Pandoc version used by our build process. In particular, since it uses the freshly archived Pandoc to generate things like templates and lists of variants, we want to be sure that we're using explicitly the Pandoc that we intend to and don't allow other versions to enter the pipeline and generate Pandoc code which isn't consistent with the version that we are archiving and using in our build process.

I'm closing, but if I'm misunderstanding this or you think I'm missing something, please do re-open and we can chat more to be sure we get this right...

msarahan:
That's fine. I think it's still necessary as a patch in the conda recipe, because there is no archived pandoc.

dragonstyle:
Just to make sure I'm understanding - that `update-pandoc.ts` (and contain that private function) is only called from a `quarto-bld` command which is essentially administrative, primarily for automating build and configuration type operations, and not available in distributed versions of Quarto. I'm thinking this isn't something that should be used at all by the Conda recipe, but I might be not understanding... If it is used I'm happy to try to do something better than what we have to make sure Conda is well supported...

msarahan:
It's not so much that it is important in the conda package at runtime, but that it is used when creating the conda package. It gets called during the build process, which for Conda does not include any download/vendoring of external tools.

PR #4735, opened 2023-03-09
Refactor chunk hook to handle verbatim-like engine better

cderv:
> WIP: needs #4732 to be merged before this is then merged in main. The current diff is compared the branch in #4732

This is a draft PR and should not be merged yet. Opened for discussion and sharing work. 

verbatim and embed knitr's engine should not be exception and their lang option needs to be used correctly. 

- refactor handling of block attributes
- create backticks enclosing dynamically according to code content. This is to correctly handle potential block code content using backticks already like `fenced.echo` in Quarto or `verbatim`-like engine in knitr to show markdown content.
- Add more comments to remember why and how it works
- exclude some engine from chunk hook `.cell-code` related processing

Probably require some specific review to decide if this is worth doing in 1.3 or later. 

Warning:  if we don't want  to move **knitr** min requirement to 1.38, we need to copy `create_fence()` for its logic into our R code.


eitsupi:
@cderv Hi, I tried Quarto 1.3 today and noticed that `prql` code blocks by `prqlr` is not rendering as expected. (eitsupi/prqlr#109)
I also noticed that the `lang` option is ignored.

I am planning a new release of `prqlr`, do I need to switch here to use the `engine` option instead of the `lang` option to change the language of the rendered code blocks?

cderv:
Thanks for reaching out @eitsupi. 

Yes quarto is not perfect for custom engine yet. `lang` option support is part of this PR on hold. 
https://github.com/quarto-dev/quarto-cli/blob/5f8841702acf8faf1e189cc67f127597ac6f21e7/src/resources/rmd/hooks.R#L393

This is currently missing in 1.3 where we have 
https://github.com/quarto-dev/quarto-cli/blob/94a815fa510272ed55843cde80ad20cac2e6c2bc/src/resources/rmd/hooks.R#L388

This is not the only issue hence this refactoring PR. 

We are "freezing" changes for now as we are preparing for 1.3 release - that is why this PR is on hold. However, I'll see with the team if I can cherry-pick this change into main before 1.3 release. 

Otherwise, you'll need some workaround in your custom engine yes. Using `options$engine` will be one for now. 

I wonder if there are other issues with custom engine like yours and maybe we need special attributes in **knitr** so that Quarto can know which is the type of engine and how it should behave. 

I'll try your engine and report on your repo.

eitsupi:
@cderv Thank you for your immediate reply!
I also tried the glue sql engine by the `glue` package and that one worked fine. (glue sql uses the `engine` option to convert output to `sql`)

Another problem I found was that the code annotations did not work for `prql` code block (Should be related to #4823).

Issue #4474, opened 2023-02-23
Display warning if user doesn't have rights to change permissions


allenmanning:
Description: 
Display warning if user doesn't have rights to change permissions

Steps to Reproduce:
1. Create a Confluence account where the user doesn't have rights to change permissions - a free account will work for this
2. Publish a quarto document using this account

Expected Result:

Check for rights to change permissions before the publish.  If the account doesn't have rights then:
- Prompt with the warning and ask them to confirm before continuing.
- Show warnings during publish

Actual Result:
No check and publish continues with an error message that doesn't prevent one or more documents from being published.

PR #4804, opened 2023-03-13
Display warning if user doesn't have rights to change permissions #4474

ghost:
Display warning if user doesn't have rights to change permissions #4474

### DIFF
--- a/src/publish/confluence/confluence-verify.ts
+++ b/src/publish/confluence/confluence-verify.ts
@@ -74,7 +74,7 @@ export const verifyOrWarnManagePermissions = async (
 
   if (!canManagePermissions) {
     const confirmed: boolean = await Confirm.prompt(
-      "Published pages will be directly editable within the Confluence web editor.\n\nThis means that if you republish the page from Quarto, you may be overwriting the web edits. We recommend that you establish a clear policy about how this published page will be revised.\n\nAre you sure you want to continue?"
+      "We've detected that your account is not able to manage the permissions for this destination.\n\nPublished pages will be directly editable within the Confluence web editor.\n\nThis means that if you republish the page from Quarto, you may be overwriting the web edits.\nWe recommend that you establish a clear policy about how this published page will be revised.\n\nAre you sure you want to continue?"
     );
     if (!confirmed) {
       throw new Error("");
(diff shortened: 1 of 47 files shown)


tests/configure-test-env.sh
```diff
@@ -17,7 +17,7 @@ python_exists=$(command -v python)
 if [ -z $python_exists ] 
 then 
   python_exists=$(command -v python3)
-  if [ -z python_exists]
+  if [ -z python_exists ]
```
allenmanning, line 20: @cderv Can you please review?  This got my tests running without errors.

src/publish/confluence/confluence-verify.ts
```diff
@@ -58,3 +59,25 @@ export const verifyConfluenceParent = async (
...
+) => {
+  const canManagePermissions = await client.canSetPermissions(
+    parent,
+    space,
+    user
+  );
+
+  if (!canManagePermissions) {
+    const confirmed: boolean = await Confirm.prompt(
+      "Published pages will be directly editable within the Confluence web editor.\n\nThis means that if you republish the page from Quarto, you may be overwriting the web edits. We recommend that you establish a clear policy about how this published page will be revised.\n\nAre you sure you want to continue?"
```
allenmanning, line 77: @cwickham Can you please review this warning prompt message?

tests/configure-test-env.sh
```diff
@@ -53,7 +53,7 @@ fi
 
 # Update tinytex
 echo ">>>> Configuring TinyTeX environment"
-if [ -z $GH_TOKEN ] && [ -n $(command -v gh) ]
+if [ -z $GH_TOKEN ] && [[ -n $(command -v gh) ]]
```
cderv, line 56: I think the safest way if to write this way 
```suggestion
if [[ -z $GH_TOKEN && -n $(command -v gh) ]]
```

So many versions that I was not sure which one to use : http://mywiki.wooledge.org/BashFAQ/031

Hopefully this woks on Mac

cderv (approved):
Yes that is ok for this update on the sh script. If it works on your side with Mac, should be ok.

src/publish/confluence/confluence-verify.ts
allenmanning: That is a great suggestion, thank you.

tests/configure-test-env.sh
allenmanning: I will give that a shot.

PR #4879, opened 2023-03-18
small correction

espinielli:
In Italian book/documents we use "Indice" for "Table of Contents".

you require too much paper work for a small typo, i.e. filling a d sending a pdf form...

if it suffice to say here that you can freely (without me claiming any rights or compensation) use my suggestion, please reuse my proposal as in this PR.


jjallaire:
@espinielli We can definitely take small changes like this without an agreement.

@andrugo Do you agree with this change? It does seem as if the current text does translate literally to "Table of Contents" (at least according to some translation sites). Not sure if the shorter version is enough, let us know your thoughts.

PR #4919, opened 2023-03-22
Add `gfm_auto_identifiers` to reader for correct ids in TOC link

### LINKED ISSUE #4917: `toc: true` with gfm does not produce valid link for Github when special char in headers
This was reported by @mine-cetinkaya-rundel

````markdown
---
title: "Untitled"
format: gfm
toc: true
---

# A header

# With ending dot.

# With a.dot in the middle

````

will render to
````markdown
# Untitled

- [A header](#a-header)
- [With ending dot.](#with-ending-dot.)
- [With a.dot in the middle](#with-a.dot-in-the-middle)

# A header

# With ending dot.

# With a.dot in the middle
````

The dots are not removed from the identifier produced in TOC. Pasting this markdown in Github will have unfunctional TOC

This is because in Pandoc `gfm_auto_identifiers` which remove the punctuation like on Github is only a **reader** extensions - see https://github.com/jgm/pandoc/issues/8709

This means that setting `toc: true` with GFM format does not work really - I think we could add `+gfm_auto_identifiers` to the reader extension `markdown` so that we produce valid TOC

cderv:
due to current limitation of Pandoc https://github.com/jgm/pandoc/issues/8709

closes #4917 

I used the tweak in `resolveFormat()` as we do it elsewhere, but we also do some tweaks like that in `formatExtras()` (see revealsj format) but it seemed to late to check `pandoc[kToc]`

This could be improved by making `pandocFormat()` not add an extension if already there. 

cc @dragonstyle 

For 1.3 or 1.4 - I just wanted to deal with it now. 😅

### DIFF
--- a/src/format/markdown/format-markdown.ts
+++ b/src/format/markdown/format-markdown.ts
@@ -4,8 +4,9 @@
  * Copyright (C) 2020-2022 Posit Software, PBC
  */
 
-import { kOutputDivs, kVariant } from "../../config/constants.ts";
+import { kFrom, kOutputDivs, kToc, kVariant } from "../../config/constants.ts";
 import { Format } from "../../config/types.ts";
+import { pandocFormat } from "../../core/pandoc/pandoc-formats.ts";
 import { createFormat, plaintextFormat } from "../formats-shared.ts";
 import { kGfmCommonmarkVariant } from "./format-markdown-consts.ts";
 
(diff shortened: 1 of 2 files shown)


cderv:
I rebase onto main which includes #4937 but the issue still remains 🤔

cderv:
@cscheid your fix in #4957 for circular dependency solved the issue in the PR with bundling. Thanks !

PR #4986, opened 2023-03-27
Treat Listing Grid Cards Consistently

dragonstyle:
Fixes #4985

This is a regression introduced when I added support for dynamically generated preview images. This  change makes it so the HTML generated in three different cases is consistent and appropriate:

1) When an image is explictly (statically known). This was always correct. 

2) When an image is discovered after rendering. This was omitting the container paragraph. 

3) When a placeholder div was placed. This was omitting the appropriate card class to ensure it gets the card style.


allenmanning:
I'm curious if there is a way we can add a test around this as we have already regressed it once.

dragonstyle:
> I'm curious if there is a way we can add a test around this as we have already regressed it once. It seems like CSS so it doesn't seem easily testable.

This is mostly caused by incorrectly structured HTML output so testing for it is pretty straightforward. I'll update to include a test!

PR #5030, opened 2023-03-30
Don’t number asciidoc cross references

dragonstyle:
Asciidoc toolchain will handle numbers, captions, and so on.


cscheid (approved):
lgtm.

allenmanning:
@dragonstyle Do we have the needed coverage here?

PR #5054, opened 2023-03-31
Stop being clever about breadcrumb titles and just use the pipeline

### LINKED ISSUE #5046: Mobile breadcrumb isn't quite right
### Bug description

Hi there, I just noticed the mobile breadcrumb doesn't show the 2nd tier correctly... This should probably be 

Formats > Presentations > PowerPoint

Looks like it shows the first link in that second menu instead?   

I see it happening on quarto.org as well as my project. Thank you!

mobile-breadcrumb

---

OS: MacOS 3.2.1 (22D68)

quarto check:
      Pandoc version 3.1.1: OK
      Dart Sass version 1.55.0: OK
      Version: 1.3.296

      Version: 3.9.6
      Jupyter: 5.2.0
      Kernels: python3

      Version: 4.1.2
      LibPaths:
        - /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/library
      knitr: 1.38
      rmarkdown: 2.13

### Checklist

dragonstyle:
This fixes #5046 the correct way, avoid conflict and duplication driven by trying to pick the title of the breadcrumb out of the sidebar DOM post render.

### DIFF
--- a/src/project/types/website/website-navigation-md.ts
+++ b/src/project/types/website/website-navigation-md.ts
@@ -8,7 +8,7 @@ import { dirname, extname, isAbsolute, join } from "path/mod.ts";
 import { Document, Element } from "../../../core/deno-dom.ts";
 
 import { Format, Metadata } from "../../../config/types.ts";
-import { NavItem, Sidebar } from "../../types.ts";
+import { NavItem, Sidebar, SidebarItem } from "../../types.ts";
 
 import {
   kBodyFooter,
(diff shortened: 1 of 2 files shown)


src/project/types/website/website-navigation-md.ts
```diff
@@ -192,6 +196,39 @@ const prevPageTitleHandler = (context: NavigationPipelineContext) => {
   };
 };
 
+const safeKey = (key: string) => {
```
allenmanning, line 199: @dragonstyle I'm not sure if refactoring to a Pure Function with unit tests would make sense here, but I thought I would suggest it.  There is growing complexity in this area.

dragonstyle: This overall change looks a lot more complicated than it really is. The root issue here was my decision to try to reuse the rendered sidebar to compute the proper rendered contents of the breadcrumbs by just plucking the contents of the sidebar from the DOM. This is ok in normal cases, but in some cases (for example, sections in the sidebar which are just text with no href), it breaks down. Rather than doing the "clever" thing, this ultimately just does what we do pretty much everywhere else- just render the contents of the breadcrumbs with our markdown pipeline. Most of the diff (especially in website navigation is just removing a few lines of code and moving code).

The growing complexity I think reflected a couple failed attempts to fix the issue with band aids and finally this, the actual fix :)

PR #5059, opened 2023-03-31
Add support for an `atlas.json` file

dragonstyle:
- The atlas file is used by O’Reilly to configure their publishing system.
- Either copy the atlas file in the root into the output directory, or cook up a fresh atlas.json file in the output directory with the correct configuration.

### DIFF
--- a/src/format/asciidoc/format-asciidoc.ts
+++ b/src/format/asciidoc/format-asciidoc.ts
@@ -76,6 +76,7 @@ export function asciidocFormat(): Format {
 
 const kFormatOutputDir = "book-asciidoc";
 const kAsciidocDocType = "asciidoc-doctype";
+const kAtlasConfigFile = "atlas.json";
 
 // Ref target marks the refs div so the post process can inject the bibliography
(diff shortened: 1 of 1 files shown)


src/format/asciidoc/format-asciidoc.ts
```diff
@@ -240,6 +256,51 @@ const asciidocBookExtension = {
   },
 };
 
+function project2Atlas(projDir: string, context: ProjectContext) {
+  // Cook up an atlas.json file that will be used as a placeholder
+  const bookStem = bookOutputStem(projDir, context.config);
+  const bookTitle = bookConfig("title", context.config);
+  const atlasJson = {
```
allenmanning, line 263: Would this be cleaner as an externally loaded JSON file that you then do a few replacements on?  I could go either way, but I thought I would just suggest.

dragonstyle: It would be easy to change to that if needed, but I don't see huge upside. I could see in the long run this being replace with a project scaffold option for books that automatically generates the atlas json with additional options (based upon  user input). We can replace this then!

PR #5118, opened 2023-04-07
Publish Docker Image To GitHub Container Registry (ghcr.io)

hamelsmu:
Adding automation that I discussed with @jjallaire 

To make this work, we will have to:

1.  Create an official `posit` account on DockerHub, and create an empty repo named `quarto`
2. Store the secret token in `secrets.DOCKER_TOKEN` in this repo's secrets

I don't know if this is what you want, but I figured a PR would help out to move it closer to the finish line!


cderv:
Chiming to share some thoughts. 

As I don't think there is an official posit account on dockerhub and everything is still at https://hub.docker.com/u/rstudio. Also I don't know the plan but  it also has been taken already by a [posit docker user](https://hub.docker.com/u/posit) 

One option we could consider is also to store the containers with the `quarto-dev` organisation directly. Github offers [Github Package product](https://github.com/features/packages) and this is available for Open Source project like us. 

It would live at https://github.com/orgs/quarto-dev/packages and I think available at 
````bash
docker pull ghcr.io/quarto-dev/quarto
````
(doc about [container registry feature](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)

Could be a good option IMO to keep everything in here.

**Comment by jjallaire** [comment]
@cderv Yes I think using the GitHub Package product with the quarto-dev org sounds perfect here!

Could you work with @hamelsmu on getting this PR aligned to use that method? (then just merge when ready)

**Comment by hamelsmu** [comment]
I'm more than happy to make the change to point to GitHub package registry.  Let me try and test the changes locally and I'll update this PR.  

Great suggestion by the way !  The GitHub Package Registry has come a long way 😀

**Comment by cderv** [comment]
@jjallaire yes sure ! We'll make that happen. 

@hamelsmu I believe we can use the `GITHUB_TOKEN` to push to the registry. Hopefully it will have enough permissions. Though I don't think it will work from a PR branch coming from a fork. So I may need to take your branch and create a new one to make the Github actions works to publish. (this is a thing with access give to `GITHUB_TOKEN` in external vs internal PR.

I need to go over the docs at https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions to check that we are correctly setup. 

Once you've tested things locally, I can probably take over to integrate in our workflows and repo.

**Comment by hamelsmu** [comment]
@cderv Yes I figured it out!  I updated this PR.  

The trick is to modify the token's permissions with this setting 

```yam
permissions:
    packages: write
```

I've tested this Action on my username so hopefully it works the first time!

docker/action.yml
```diff
@@ -0,0 +1,55 @@
...
+        RUN dpkg -i quarto-linux-amd64.deb && rm quarto-linux-amd64.deb
+        EOF
+
+    - if: ${{ inputs.daily != 'true' && inputs.daily != 'True'  }}
+      shell: bash
+      run: echo "LATEST_TAG=-t ${{ env.BASE_NM }}:latest" >> $GITHUB_ENV
+
+    - if: ${{ inputs.daily && inputs.daily != 'false' && inputs.daily != 'False' }}
+      shell: bash
+      run: echo "DAILY_TAG=-t ${{ env.BASE_NM }}:daily" >> $GITHUB_ENV
```
cderv, line 48: should the tag be `prerelease` instead of daily ? 
This is the term we use - https://quarto.org/docs/download/prerelease.html and we can build several time per day really. 

Currently not a good example as we switched to latest stable vs RC.

docker/action.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Create DockerFile
+      shell: bash
+      run: |
+        cat << 'EOF' >> Dockerfile
+        FROM ubuntu:22.04
+        COPY ${{ inputs.source }} quarto-linux-amd64.deb
+        RUN dpkg -i quarto-linux-amd64.deb && rm quarto-linux-amd64.deb
+        EOF
+
+    - if: ${{ inputs.daily != 'true' && inputs.daily != 'True'  }}
```
cderv, line 42: [`fromJSON()`](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson) should be able to convert an input string `true` to boolean and allow `fromJSON(input.daily) == false` - it would be simpler. 

In composite action, inputs are strings always, but if we fix the `workflow_dispatch` with typed input, this should be ok. 

Non blocking but simpler to understand and maintain probably.

cderv (commented):
General comment in addition to in file comments. This is some thoughts about integration on your work in the overall project. 

I have one concerns: I just want to prevent the publish step to fail because of docker build failure for some reason, or publishing to registry error. 

Now that I can see how this works thanks to your work. I see several solutions 
* we could add for now [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) on the docker step.
* do another job for this using `needs` to depend on the linux one.  The workflow would fail but only because of this jobs. The release would still be created.
* Last options I see is to have this as another workflow to be triggered after the create-release one has been done. This other workflow would get the file required from the release, and then published to the registry. 

This is some thoughts also for later as I'll do a pass on all actions for 1.4 so I can improve at this point.

I can take over and do some modification before merging if you want.

docker/action.yml
hamelsmu: @jjallaire wanted the term to be `daily` [see slack](https://positpbc.slack.com/archives/C01LQETLA7L/p1679620438572129?thread_ts=1679602309.415929&cid=C01LQETLA7L)  I don't mind changing it though if there is a diff opinion.

docker/action.yml
hamelsmu: Yeah I saw some GitHub issues with composite actions not accepting booleans.  That's why I stayed away from this

see [this issue](https://github.com/actions/runner/issues/2238)

docker/action.yml
```diff
@@ -0,0 +1,55 @@
...
+  daily:
+    description: "Apply the `daily` tag to the image."
+    required: false
+    default: true
+
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      run: echo "BASE_NM=ghcr.io/${{ inputs.org }}/${{ inputs.name }}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
hamelsmu: yeah I can change it.  Do you think you might want to use this composite action to push to other registries?

docker/action.yml
hamelsmu: Let me paste his comment in here, that might be easier:

```
You probably want latest, pre-release-latest, and maybe daily ? or perhaps just latest and daily (as pre-release is just a web facing designation we give when the daily is stabilizing)
```

.github/workflows/create-release.yml
```diff
@@ -565,6 +568,16 @@ jobs:
           asset_name: changelog.md
           asset_content_type: text/markdown      
 
+      - uses: quarto-dev/quarto-cli/docker@master
```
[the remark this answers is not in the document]
hamelsmu: Good catch.  This action isn't run on this PR IIUC b/c its only triggered on a workflow dispatch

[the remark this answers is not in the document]
hamelsmu: I moved it out of the root 👍🏽

docker/action.yml
hamelsmu: Oh I see what you mean if it is boolean it will always be lowercase.   Sorry just realized what you were saying.

hamelsmu:
@cderv I went with this suggestion (which I think is very reasonable)

> do another job for this using needs to depend on the linux one. The workflow would fail but only because of this jobs. The release would still be created.

docker/action.yml
hamelsmu: I'll hold off on changing this until we get a comment on changing the inputs to boolean

docker/action.yml
cderv: Perfect ! Let it be daily then. Thanks for reminding me the context !

.github/workflows/create-release.yml
```diff
@@ -17,6 +17,9 @@ on:
         required: false
         default: "false"
```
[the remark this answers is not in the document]
cderv: Discussed with @dragonstyle off github, we can do the change. It will be better to really have those as boolean, and I don't think it will change anything on current workflow as `github.event.inputs.publish-release`. 

We would need to use `${{ inputs.pre-release }}` otherwise to get the boolean, but does not work in the composite anyway

docker/action.yml
cderv: > Oh I see what you mean if it is boolean it will always be lowercase.

yes exactly - still be a string but either 

* `fromJSON(input.daily)` would get you a boolean
* `inputs.daily` could be compared to string `'true'` or `'false'` 

Less case to test. We can do the change with the workflow dispatch

docker/action.yml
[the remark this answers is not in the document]
cderv: You're right we won't use that directly ourself - same as changing the org name, we probably won't either. 

No need to change; I think I wondered if this was planned to be reused by you somehow too. I guess not. 

Nothing to change here - closing

docker/action.yml
hamelsmu: OK I made the change in https://github.com/quarto-dev/quarto-cli/pull/5118/commits/d82b4d660d970ad22cd88dc98806aba19013bba8

eitsupi:
Some questions.

- Why use Ubuntu instead of Debian? In my opinion Debian has the advantage of easy Chromium installation and may be more suitable for Quarto.
- Is there any reason not to do a multi-platform build (amd64 and arm64) here? It would not be difficult to do that as the deb package for arm64 is also currently built every time.

hamelsmu:
> Why use Ubuntu instead of Debian

I've found that data science-y type of things have the best compatibility with Ubuntu.  Feel free to change it if you like!  

> Is there any reason not to do a multi-platform build (amd64 and arm64) here? It would not be difficult to do that as the deb package for arm64 is also currently built every time.

We could certainly do this.  I was trying to scope it to be simple initially but we could expand it to build on all platforms that you build packages for.

hamelsmu:
@cderv  Let me know if should change anything about this PR (wasn't sure so wanted to check)

cderv:
@hamelsmu I am currently testing this in my fork to check everything works as expected. A few adjustment to make. I'll comment in a review and do the change probably. Thanks !

cderv (commented):
Here is what I noticed we need to change for this to work effectively in the repo. I'll push the change in the branch directly

.github/actions/docker/action.yml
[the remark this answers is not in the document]
cderv: In fact those are different type of action. It would require doing a reusable action. I'll so that another time as this could be useful

.github/workflows/create-release.yml
```diff
@@ -7,15 +7,21 @@ on:
...
+        default: true
       draft:
         description: "Draft"
         required: false
-        default: "false"
+        type: boolean
+        default: false
+
+permissions:
+  packages: write
```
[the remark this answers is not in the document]
cderv: Yeah the permissions are really tricky. I spend some times reading and trying to understand. Basically, on PR from fork everything will be deactivated, and then it will depends on the organization or repo configuration. 

The tricky part is that if you start using like this
```yaml
permissions:
  packages: write
```

then you need to define every permission you need as this will reset all the other default permissions to none. Which is an issue here as a new release can't be created. 

I learnt some stuff!

cderv (approved):
I think this is all good now. Everything is working fine and container are published in github Package as expected. Release are still created as we expect. 

I'll share with the team internally in our meeting, and then I'll merge when possible (just taking precaution as we are finalizing the 1.3 release, and things are frozen more than usual because of that). 

Thanks a lot @hamelsmu !

cderv:
@hamelsmu first one is available here: https://github.com/quarto-dev/quarto-cli/pkgs/container/quarto

it should be visible publicly now. We'll monitor the next runs to see if this is working as expected

PR #5154, opened 2023-04-11
JATS Fixes

dragonstyle:
- Pandoc 3 workaround leaves an ID that should be removed, failing JATS validation (added test)
- The template is missing a DTD declaration, failing JATS validation


dragonstyle:
For 1.3 consideration

cscheid (approved):
This looks good to me!, and pretty safe as well, since it's gated inside a JATS format check.

PR #5153, opened 2023-04-11
Ensure case-insentive search of installable tools 

cderv:
Complement #5155 by ensuring to be case insentive in our tool search using the refactored function

**To be merged for 1.4**

PR #5155, opened 2023-04-11
Fix tools update by correctly checking lower case name of the tools

cderv:
otherwise, we enter an infinite loop of updating the tools -> Yes -> not install sorry

Smallest version of the #5153 fix which we will do for 1.4


dragonstyle (approved):
great find

cderv:
@dragonstyle I let you merge when ok - and rebuild a version maybe ?

PR #5157, opened 2023-04-11
Feature/jats 1.4

dragonstyle:
1.4 JATS work, including

- `xmllint` based schema validation tests
- additional metadata cleanup


allenmanning:
Would it make more sense for the XSD 300+ file directory to live in `tools`?

PR #5176, opened 2023-04-13
Use new hosted PDF for contibutor agreeement

cderv:
This PR change the old pdf with the new using Posit brand and hosted on posit.co


cderv:
I guess I can merge this myself then

PR #5241, opened 2023-04-19
Provide a ‘render-file’ lifetime when rendering single file books

dragonstyle:
Jupyter embedding uses this lifetime to cache notebook data (and other code could theoretically use it to if it is assuming it is a part of the render pipeline). Since books takeover single file rendering, we should provide it there as well.


dragonstyle:
Reviewed with @jjallaire and @cscheid good to go.

PR #5248, opened 2023-04-20
Add optional hover to cross references

dragonstyle:
- Allow cross references to show their reference in a tippy
- Controlled via `crossrefs-hover`
- defaults to true
- sections only show the heading + first child


src/resources/formats/html/templates/quarto-html.ejs
```diff
@@ -344,6 +344,36 @@ window.document.addEventListener("DOMContentLoaded", function (event) {
...
+  <% if (hoverXrefs) { %>
+  const xrefs = window.document.querySelectorAll('a.quarto-xref');
+  for (var i=0; i<xrefs.length; i++) {
+    const xref = xrefs[i];
+    tippyHover(xref, function() {
+      let href = xref.getAttribute('href');
+      try { href = new URL(href).hash; } catch {}
+      const id = href.replace(/^#\/?/, "");
+      const note = window.document.getElementById(id);
+      console.log(id);
```
allenmanning, line 356: `console.log`

dragonstyle: > console.log

Thank you!!!!

PR #5325, opened 2023-04-27
feature/toc-body-side

### LINKED ISSUE #3473: Ability to show the TOC both in body and right simulataneously
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/3336

<sup>Originally posted by **AlFontal** November 15, 2022</sup>
As far as I know, it's possible to select a single location to show the Table of Contents, but while the dynamic TOC that always stays on the right is useful, sometimes I also want to show a full TOC just after the header in an article/blog post. 

Is there any mechanism to have it behave like that right now?</div>

ghost:
closes #3473


src/resources/editor/tools/yaml/yaml-intelligence-resources.json
```diff
@@ -12496,6 +12498,8 @@
     "The input document that will serve as the root document for this\nmanuscript",
     "The deployed url for this manuscript",
     "Whether to generate a MECA bundle for this manuscript",
+    "Additional file resources to be copied to output directory",
```
allenmanning, line 12501: @dragonstyle I'm not sure why I picked this up.  Should I remove it?

rich-iannone (approved):
This looks great!

src/resources/editor/tools/yaml/yaml-intelligence-resources.json
dragonstyle: I'm not sure either- I definitely added that but I _thought_ I regenerated the underlying schema and committed that alongside. I must not have! Safe to ignore.

src/resources/schema/document-toc.yml
```diff
@@ -28,7 +28,7 @@
 
 - name: toc-location
   schema:
-    enum: ["body", "left", "right"]
+    enum: ["body", "left", "right", "left-body", "right-body"]
   default: "right"
   tags:
     formats: [$html-doc]
```
dragonstyle, line 34: Could you also update the description below to include the new options? Might consider adding a short & long description and explaining these `-body` options a bit as they probably warrant a sentence to explain....

dragonstyle (approved):

src/resources/editor/tools/yaml/yaml-intelligence-resources.json
allenmanning: This may be my workflow now.  If I see conflicts, I just accept "theirs" in my IDE rather than a manual resolve.  After that I run "build-js".  I'm hoping that my changes are just replaced, then I regenerate my changes back in.  I'm going to try this again and see if that resolves it.  Thanks for reviewing.

allenmanning: This seems to have resolved itself.

src/resources/editor/tools/vs-code.mjs
```diff
@@ -28077,10 +28082,8 @@ function postProcessAnnotation(parse) {
   if (parse.components.length === 1 && parse.start === parse.components[0].start && parse.end === parse.components[0].end) {
     return postProcessAnnotation(parse.components[0]);
   } else {
-    return {
-      ...parse,
-      components: parse.components.map(postProcessAnnotation)
-    };
+    parse.components = parse.components.map(postProcessAnnotation);
```
allenmanning, line 28085: @dragonstyle I'm not sure why in my change the code gen has now been refactored to prefer a non-mutable approach.

dragonstyle: I def can't explain it - those files are generated so perhaps something changed (e.g. a dependency) that resulted in them being generated a slightly different way...

allenmanning: Thanks for reviewing.

PR #5326, opened 2023-04-27
quarto tools with no args throws error #5314

### LINKED ISSUE #5314: `quarto tools` with no args throws error
# Steps To Reproduce

Description: 
`quarto tools` with no args throws error

Steps to Reproduce:
Check help of `quarto tools` and see "Use 'quarto tools' with no arguments to show the status of all tools."
```
❯ quarto tools help
Check file:///Users/amanning/projects/github/quarto/quarto-cli/src/quarto.ts

  Usage:   quarto tools <command> [tool]
  Version: 99.9.9

  Description:

    Installation and update of ancillary tools.

      tools:
        tinytex
        chromium

      commands:
        install
        uninstall
        update

    Use 'quarto tools' with no arguments to show the status of all tools.
 ```  

- Follow doc by running `quarto tools`
    

## Expected Result:

The same results as `quarto tools all` or `quarto tools list`

## Actual Result:
`ERROR: Missing argument(s): command`

### PR DESCRIPTION
**Pull request by ghost**
fixes #5314

### CODE REVIEW

File: src/command/tools/cmd.ts [hunk 1]
```diff
@@ -81,7 +82,6 @@ Use 'quarto tools' with no arguments to show the status of all tools.`,
           );
         }
         break;
-      default:
```
dragonstyle, line 84: I think you should consider handling the `default` case of any unrecognized command (e..g `quarto tools foobar`). Currently that just does nothing and prior to this change that would run the `list` command. Both don't seem right - I think an error is warranted in this case since the user is providing a command that we don't know how to fulfill.

dragonstyle (approved):
This change is good overall but I think you could consider one improvement (ok if you'd like to defer that)

rich-iannone (approved):
Looks good to me (including Charles' suggestion).

PR #5348, opened 2023-04-28
Feature/jats notebook

dragonstyle:


src/resources/filters/quarto-post/book.lua
```diff
@@ -1,6 +1,9 @@
 -- book.lua
 -- Copyright (C) 2020-2022 Posit Software, PBC
 
+local authors = require '../common/authors'
+local license = require '../common/license'
+
```
cscheid, line 6: 🤩

dragonstyle: This has caused me to enter LUA inlining nightmare land. See next commit, then following failure:

```
too many local variables (limit is 200) in main function near '='
[326](https://github.com/quarto-dev/quarto-cli/actions/runs/4833870917/jobs/8614410733?pr=5348#step:5:327)
```

salute

cscheid (approved):
Beast PR, looks good!

(Only one minor question on moving the embed shortcode resolution from pre-engine to post engine. I'm sure there's a good reason for it, but I'm wondering if it'll break user filters that looked at the embed contents.)

src/resources/filters/quarto-post/book.lua
cscheid: That's not terrifying at all

Issue #5495, opened 2023-05-11
Console Errors with `schema-schema.test.ts`


allenmanning:
## Description: 
[run-smokes Deno deprecation warnings](https://github.com/quarto-dev/quarto-cli/issues/5495)

## Steps to Reproduce:
- Create a pull request and check the logs

or

```
./run-tests.sh unit/schema-validation/schema-schema.test.ts
```

### Expected Result:

No warnings in Deno logs

### Actual Result:

![Image](https://github.com/quarto-dev/quarto-cli/assets/674792/809340f5-2884-4f05-a337-6afd7463c593)

```
Annotations
8 warnings
run-smokes (ubuntu-latest)
"files" configuration is deprecated. Please use "include" and "exclude" instead.
run-smokes (ubuntu-latest)
propertyNames and case convention detection are mutually exclusive.
run-smokes (ubuntu-latest)
propertyNames and case convention detection are mutually exclusive.
run-smokes (ubuntu-latest)
propertyNames and case convention detection are mutually exclusive.
run-smokes (windows-latest)
"files" configuration is deprecated. Please use "include" and "exclude" instead.
run-smokes (windows-latest)
propertyNames and case convention detection are mutually exclusive.
run-smokes (windows-latest)
propertyNames and case convention detection are mutually exclusive.
run-smokes (windows-latest)
propertyNames and case convention detection are mutually exclusive.
```

allenmanning:
It seems that `propertyNames and case convention detection are mutually exclusive.` is being triggered by our unit tests.

For example: `./run-tests.sh unit/schema-validation/schema-schema.test.ts`

https://github.com/quarto-dev/quarto-cli/blob/b420134feb547824d4980cf47a560901751abfdf/src/core/lib/yaml-schema/common.ts#L194-L202

allenmanning:
I've broken out the Deno warnings to [a separate issue](https://github.com/quarto-dev/quarto-cli/issues/5760).

PR #5542, opened 2023-05-15
bug/5390-vimeo-newlink

### LINKED ISSUE #5390: Embedding Vimeo video does not work with "unlisted" links
## Description
Unlisted Vimeo videos produce a link that does not work with the Quarto shortcode {{ video }}

## Steps to Reproduce
- Log in with a Standard Account. 
- select a video with privacy settings "unlisted"
- click the share link button
- The link copied to the clipboard is `https://vimeo.com/800138810/fdcc90d662?share=copy`
- Use this link in the video short code `{{< video https://vimeo.com/800138810/fdcc90d662?share=copy >}}`

## Expected Result: 
A video snippet is created with  the url.  The video snippet can load the video.

## Actual Result:
Using this link with the shortcode results in the error message "Invalid embedding-code"

### Workaround
Replace the slash in the privacy hash with `"?h="` and delete `"?share=copy"` so 
`{{< video https://vimeo.com/800138810?h=fdcc90d662 >}}`

Please note: the privacy setting "Hide from Vimeo", that creates links for embedding only, does not (!) contain a privacy hash and works well with the shortcode `{{< video https://vimeo.com/800139587?share=copy >}}` so no problem for "Hide-from-Vimeo" videos as of writing this post, but only for "Unlisted" videos.

ghost:
fixes #5390


src/resources/extensions/quarto/video/video.lua
```diff
@@ -116,10 +127,22 @@ local vimeoBuilder = function(params)
   if not (params and params.src) then return nil end
 
   local VIMEO_STANDARD = 'https://vimeo.com/'
-  local match = checkEndMatch(params.src, VIMEO_STANDARD)
+  local match = checkMatchStart(params.src, VIMEO_STANDARD)
   if not match then return nil end
 
-  params.src = 'https://player.vimeo.com/video/' .. match
+  local internalMatch = string.match(params.src, 'vimeo.com/(.-)?share=copy')
```
cderv, line 133: I think you want to make the `?share=copy` part optional. This is query parameter for url string, but the url that can be used in shortcode could be `https://vimeo.com/548291210/fdcc90d662`

I believe this is valid usage and should be dealt the same
```` 
{{ vimeo https://vimeo.com/548291210/fdcc90d662 }}
````

I understand we get the links with query parameter when clicking on the share button, but it seems fair to support the version without. Otherwise, we can wait for someone to use that form and report to avoid unnecessary precaution support.

File: src/resources/extensions/quarto/video/video.lua [hunk 2]
```diff
@@ -12,6 +12,17 @@ local interpolate = function(str, vars)
           end))
 end
 
+local function splitString (toSplit, delimiter)
+  if delimiter == nil then
+    delimiter = "%s"
+  end
```
cderv, line 18: Lua tips useful for default argument in function and shorter syntax than full if then
```suggestion
  delimiter = delimiter or "%s"
```

Example: https://riptutorial.com/lua/example/4081/default-parameters

> The idiom `name = name or "Jack"` works because `or` in Lua short circuits. If the item on the left side of an `or` is anything other than `nil` or `false`, then the right side is never evaluated.

src/resources/extensions/quarto/video/video.lua
```diff
@@ -60,16 +71,16 @@ local replaceCommonAttributes = function(snippet, params)
   return result
 end
 
-local checkEndMatch = function(value, matcherFront)
+local checkMatchStart = function(value, matcherFront)
   return string.match(value, matcherFront .. '(.-)$')
 end
```
cderv, line 76: I understand why you changed name, but I would have expect a function that check on start of the string to use a regex starting with special char `^` which means start of string match. 

I believe you function would also match something like `anythinghttps://vimeo.com/`

Probably won't happen, but I am curious why match on anything on the end using `(.-)$`. 

maybe there is the reason I don't see. 

Everythings works - so probably no problem but usually using `^` for matching at start of string.

cderv (approved):
Perfect. That will solve. A few comments below that you are free to ignore as everything is fine otherwise. Just possible improvement for future us and prevent possible undesired behavior.

rich-iannone (approved):
This looks good (taking into account @cderv 's great suggestions).

src/resources/extensions/quarto/video/video.lua
allenmanning: Thanks for suggesting this, I've updated the PR.  👍

src/resources/extensions/quarto/video/video.lua
allenmanning: This is better.  Thanks for suggesting.  Updated.

src/resources/extensions/quarto/video/video.lua
allenmanning: Thanks for suggesting this.  Can you help me with a unit test for this?  I'm not sure how I can get it to fail with a unit test.

cderv: I am not sure how unit test works in Lua but I would suggest either 

* to test `checkMatchStart()` according to what you expect as results - if you expect to be a string that "start with match" , I think you won't get this with the regex used. (from quick test in repl.it with Lua)

* or directly your video function where I would expect `{{ video anythinghttps://vimeo.com/ }}` to be ignore, or shown an invalid URL. 

Are you testing already in the code path for the URL part of the video shortcode to be either a URL or a local file ? I believe we do that elsewhere in our TS code base. 

Maybe it is too much to check, and we can assume the user would use a valid url starting with `http(s)`

cderv (approved):
Thanks for taking the suggestion into account.

allenmanning:
I'm waiting on these warnings to be fixed in the check before merging.  #5495

PR #5662, opened 2023-05-24
Create Lithuanian translation

GegznaV:
YAML file with Lithuanian translation was added.


allenmanning:
Thank you for contributing @GegznaV 🚀

PR #5704, opened 2023-05-26
[R] Use ... in `add_html_caption()` as we overwrite knitr's internals

### LINKED ISSUE #5702: Quarto fails to render htmlwidgets when running {knitr} 1.43 with Error in `add_html_caption()`:
### Bug description

Hello, 

# Current situation

{knitr} has a recent update to v1.43 that makes rendering of R code chunks with htmlwidgets fails with
```bash
$ quarto render Reprex.Rmd 


processing file: Reprex.Rmd
1/5                
2/5 [setup]        
3/5                
4/5 [DT htmlwidget]

Quitting from lines 15-16 [DT htmlwidget] (Reprex.Rmd)
Error in `add_html_caption()`:
! unused argument (xfun::grep_sub("^[^<]*<[^>]+aria-labelledby[ ]*=[ ]*\"([^\"]+)\".*$", "\\1", x))
Backtrace:
  1. global .main()
  2. execute(...)
  3. rmarkdown::render(...)
     ...
 14. sew(res, options)
 15. knitr:::sew.list(x, options, ...)
 16. base::lapply(x, sew, options, ...)
 17. FUN(X[[i]], ...)
 18. knitr:::sew.knit_asis(x, options, ...)
Execution halted
```

This is not specific to {DT}, but also occurs with other htmlwidgets related packages : {collapsibleTree}, ...

# ReprEx

Reprex.qmd is the following (code chunk separator manually edited for correct rendering here): 
```qmd
---
title: "ReprEx htmlwidget"
format: html
---

``{r setup}
library(DT)
xfun::session_info('knitr')
``

``{r DT htmlwidget}
datatable(iris)
``
```

# Workaround

Downgrading {knitr} to 1.42 workaround the problem : 

## same rendering result with knitr 1.42
```bash
$ quarto render Reprex.qmd 


processing file: Reprex.qmd
                                                                                                          
output file: Reprex.knit.md

pandoc 
  to: html
  output-file: Reprex.html
  standalone: true
  section-divs: true
  html-math-method: mathjax
  wrap: none
  default-image-extension: png
  
metadata
  document-css: false
  link-citations: true
  date-format: long
  lang: en
  title: ReprEx htmlwidget
  
Output created: Reprex.html
$ quarto check

      Pandoc version 3.1.1: OK
      Dart Sass version 1.55.0: OK
      Version: 1.3.353

      Version: 3.8.16 (Conda)
      Path: /usr/local/Caskroom/miniconda/base/envs/r-reticulate/bin/python
      Jupyter: (None)

      Version: 4.2.1
      LibPaths:
      knitr: 1.42
      rmarkdown: 2.21

```

# Session Info
## quarto version
```bash
$ quarto -V
1.3.353
```
## Rstudio version 

Rstudio version ( as it also occurs in Rstudio ) : RStudio 2023.03.1+446 "Cherry Blossom" Release (6e31ffc3ef2a1f81d377eeccab71ddc11cfbd29e, 2023-05-09) for macOS
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2023.03.1+446 Chrome/108.0.5359.179 Electron/22.0.3 Safari/537.36

## quarto check
```bash
$ quarto check

      Pandoc version 3.1.1: OK
      Dart Sass version 1.55.0: OK
      Version: 1.3.353

      Version: 3.8.16 (Conda)
      Path: /usr/local/Caskroom/miniconda/base/envs/r-reticulate/bin/python
      Jupyter: (None)

      Version: 4.2.1
      LibPaths:
      knitr: 1.43
      rmarkdown: 2.21

```

### Checklist

cderv:
Fixes #5702 

**knitr** 1.43 has a new feature to support aria-labelledby on HTML Widget, but it was done by adding an argument to `add_html_caption()` that we override in Quarto. 

This means the function we pass to **knitr** does not have the correct number of argument anymore. 

Using `...` in function we replace in knitr's namespace should prevent this error. 

I only did this specific function to make a possible hot fix minimal, but we should probably do that in all our `assignInNamespace()` call. 

@jjallaire do you think this should be "fixed" here ? Or should **knitr** track and know what Quarto is using internally to pay attention to changes ? It seems to me Quarto should adapt, but I may be wrong

More generally, we may need to reconsider the interaction in knitr for Quarto. Maybe we need to add specific hooks in **knitr**. 
This I understand that we don't want to depend on a specific **knitr** version. If the latter is the policy, we need to protect and adapt more closely to change in **knitr**.  

As a feedback for us, this shows that we really need to tests Quarto against current dev version of knitr (and rmarkdown probably). Our current usage of renv prevent that right now.

### DIFF
--- a/src/resources/rmd/patch.R
+++ b/src/resources/rmd/patch.R
@@ -108,7 +108,7 @@ wrap_asis_output <- function(options, x) {
   
   output_div(x, output_label_placeholder(options), classes)
 }
-add_html_caption <- function(options, x) {
+add_html_caption <- function(options, x, ...) {
   if (inherits(x, 'knit_asis_htmlwidget')) {
     wrap_asis_output(options, x)
   } else {


jjallaire:
I think we should generally make sure that upgrading knitr doesn't break quarto. This can be handled on both sides -- in this case it would be great if we could make a change to knitr to prevent the breakage (e.g. somehow make that argument optional). Otherwise many, many RStudio users will find that Quarto no longer works when the update knitr. Yes conceptually this should be on Quarto as we are the ones overriding knitr, but we don't in practice have the ability to update all of the quarto binaries in the field (especially those embedded in RStudio) so I think it has to go both ways.

I think we should also take your fix and back port it to v1.3 just to cover all bases.

In general, we should make these override points more explicit and managed / tested. Testing against development version of knitr seems like a very good step in this direction (we could just do that after the renv::restore).

cderv:
Thanks for clarifying the position. I agree this is hard to update Quarto in the field. 

> in this case it would be great if we could make a change to knitr to prevent the breakage (e.g. somehow make that argument optional)

@yihui can we do this a schedule a patch release for **knitr** ? 

> I think we should also take your fix and back port it to v1.3 just to cover all bases.

@dragonstyle as discussed we should probably take this then and do your backport magic :) 

> Testing against development version of knitr seems like a very good step in this direction (we could just do that after the renv::restore).

Yes I am onto it right now in #5705

yihui:
> @yihui can we do this a schedule a patch release for **knitr** ?

In theory, yes, but it will require an explicit explanation to CRAN if another release needs to be made within 7 days. Usually releasing **knitr** to CRAN is not a smooth process: every time it will be blocked by CRAN's automated revdeps check (which almost always finds false positives due to the huge number of revdeps), and will require a CRAN member's manual approval. Neither problem is too bad. I just need to be careful not to overuse my "release credit".

racing-focus:
> Fixes #5702
> 
> **knitr** 1.43 has a new feature to support aria-labelledby on HTML Widget, but it was done by adding an argument to `add_html_caption()` that we override in Quarto.
> 
> This means the function we pass to **knitr** does not have the correct number of argument anymore.
> 
> Using `...` in function we replace in knitr's namespace should prevent this error.
> 
> I only did this specific function to make a possible hot fix minimal, but we should probably do that in all our `assignInNamespace()` call.
> 
> @jjallaire do you think this should be "fixed" here ? Or should **knitr** track and know what Quarto is using internally to pay attention to changes ? It seems to me Quarto should adapt, but I may be wrong
> 
> More generally, we may need to reconsider the interaction in knitr for Quarto. Maybe we need to add specific hooks in **knitr**. This I understand that we don't want to depend on a specific **knitr** version. If the latter is the policy, we need to protect and adapt more closely to change in **knitr**.
> 
> As a feedback for us, this shows that we really need to tests Quarto against current dev version of knitr (and rmarkdown probably). Our current usage of renv prevent that right now.

I have a file that I am rendering and have the following error.

```{r}
Quitting from lines 201-207 [unnamed-chunk-5] (article_4.qmd)
Error in `add_html_caption()`:
! unused argument (xfun::grep_sub("^[^<]*<[^>]+aria-labelledby[ ]*=[ ]*\"([^\"]+)\".*$", "\\1", x))
Backtrace:
  1. global .main()
  2. execute(...)
  3. rmarkdown::render(...)
     ...
 14. sew(res, options)
 15. knitr:::sew.list(x, options, ...)
 16. base::lapply(x, sew, options, ...)
 17. FUN(X[[i]], ...)
 18. knitr:::sew.knit_asis(x, options, ...)
Execution halted

```

I assume it is related?

```{r}
Using ... in function we replace in knitr's namespace should prevent this error.

```

May I ask how this hot fix works?


Cheers,

PR #5729, opened 2023-05-29
fix: remove section-divs as not for revealjs

mcanouil:
This PR removes `section-divs` from formats other than `revealjs`.

I am not 100 % sure, if the four files needed to be modified, _i.e._, some might be generated.

The expected side effect is for the option to be removed from the `revealjs` reference guide page, indeed, headers are by design already embedded in `section` tags in Reveal.js, thus by design can't be disabled.

For short, `section-divs` is invalid and has no effect in `revealjs` format, which can lead to confusion for users.

<table>
<tr><th>HTML</th><th>Reveal.js</th></tr>
<tr>
<td>

Screenshot 2023-05-29 at 15 31 55

</td>
<td>
Screenshot 2023-05-29 at 15 34 53
</td>
</tr>
</table>

### DIFF
--- a/src/resources/editor/tools/vs-code.mjs
+++ b/src/resources/editor/tools/vs-code.mjs
@@ -15105,7 +15105,7 @@ var require_yaml_intelligence_resources = __commonJS({
           name: "section-divs",
           tags: {
             formats: [
-              "$html-files"
+              "$html-doc"
             ]
           },
           schema: "boolean",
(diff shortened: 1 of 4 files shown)


cderv:
> I am not 100 % sure, if the four files needed to be modified, i.e., some might be generated.

I encountered this in https://github.com/quarto-dev/quarto-cli/pull/5575

You need to modify the YAML file only and then run `quarto build-js` - this will modify the other files

Side note on this: Pandoc does not seem to disallow `--section-divs` flag with `--to revealjs`, and there documentation does not say it does nothing (https://pandoc.org/MANUAL.html#option--section-divs) Maybe that would be worth opening an issue there so that they update their doc ?

mcanouil:
@cderv Updated the branch and ran `quarto build-js`.

PR #5772, opened 2023-06-01
Add cell label in Quarto render status

### LINKED ISSUE #5752: [FR] Replace "Cell 1/N" by label (if provided) when using Jupyter
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/5714

<sup>Originally posted by **bryanwhiting** May 27, 2023</sup>
### Description

This isn't helpful:
```
Cell 1/20...Done
  Cell 2/20...Done
  Cell 3/20...Done
  Cell 4/20...Done
  Cell 5/20...Done
  Cell 6/20...Done
  ```

What would be helpful is for it to print the name of the code chunk, like Rmarkdown does.</div>

### PR DESCRIPTION
**Pull request by yuninxia**
In this pull request, I've refined the visibility and interpretability of Quarto render status messages. 

Key modifications include:

- Cells with labels: The syntax has been updated from `"  Cell {0}/{1}..."` to `"  Cell '<label>' {0}/{1}..."`.

- Cells without labels: The format has been altered from `"  Cell {0}/{1}..."` to `"  Cell '' {0}/{1}..."`.

These refinements boost the readability of the script and contribute to the overall maintainability and professionalism of our codebase. I look forward to your review.

### DIFF
--- a/src/resources/jupyter/notebook.py
+++ b/src/resources/jupyter/notebook.py
@@ -153,27 +153,27 @@ def notebook_execute(options, status):
    # compute total code cells (for progress)
    current_code_cell = 1
    total_code_cells = sum(cell.cell_type == 'code' for cell in client.nb.cells)
-   
-   # find max id length (for progress)
-   max_id_len = max(len(cell.id) for cell in client.nb.cells)
 
+   # find max label length (for progress)   
+   max_label_len = max(len(nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell).get('label', '')) for cell in client.nb.cells)
+   
    # execute the cells
    for index, cell in enumerate(client.nb.cells):
-      # find cell id
-      source_list = cell.source.split('\n')
-      cell_id = cell.id if any(s.startswith("#| label") or s.startswith("#| id") for s in source_list) else ""
-      padding = "." * (max_id_len - len(cell_id))
-      
+      # read cell options
+      cell_options = nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell)
+      cell_label = cell_options.get('label', '')
+      padding = "." * (max_label_len - len(cell_label))
+
       # progress
       progress = (not quiet) and cell.cell_type == 'code' and index > 0
       if progress:
          status("  Cell {0}/{1}: '{2}'{3}...".format(
             current_code_cell - 1,
             total_code_cells - 1,
-            cell_id,
+            cell_label,
             padding
          ))
-
+         
       # clear cell output
       cell = cell_clear_output(cell)
 
--- a/src/resources/jupyter/notebook.py
+++ b/src/resources/jupyter/notebook.py
@@ -153,26 +153,23 @@ def notebook_execute(options, status):
    # compute total code cells (for progress)
    current_code_cell = 1
    total_code_cells = sum(cell.cell_type == 'code' for cell in client.nb.cells)
-
-   # find all the labels in the notebook
-   cell_labels = [
-      (next(iter(re.findall(r'#\| label: (.*)', cell.source)), '').strip() 
-       if cell.cell_type == 'code' else '')
-       for cell in nb.cells
-       ]
-   max_label_len = max(len(label) for label in cell_labels)
+   
+   # find max id length (for progress)
+   max_id_len = max(len(getattr(cell, "id", "")) for cell in nb.cells)
 
    # execute the cells
    for index, cell in enumerate(client.nb.cells):
+      # find cell id
+      cell_id = cell.id if hasattr(cell, "id") else ""
+      padding = "." * (max_id_len - len(cell_id))
+      
       # progress
       progress = (not quiet) and cell.cell_type == 'code' and index > 0
       if progress:
-         label = cell_labels[index]
-         padding = "." * (max_label_len - len(label))
          status("  Cell {0}/{1}: '{2}'{3}...".format(
             current_code_cell - 1,
             total_code_cells - 1,
-            label,
+            cell_id,
             padding
          ))
 
--- a/src/resources/jupyter/notebook.py
+++ b/src/resources/jupyter/notebook.py
@@ -149,19 +149,25 @@ def notebook_execute(options, status):
    # complete progress if necessary
    if (not quiet) and created:
       status("Done\n")
-      
-   # compute total code cells (for progress)
-   current_code_cell = 1
-   total_code_cells = sum(cell.cell_type == 'code' for cell in client.nb.cells)
 
-   # find max label length (for progress)   
-   max_label_len = max(len(nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell).get('label', '')) for cell in client.nb.cells)
+   current_code_cell = 1
+   total_code_cells = 0
+   cell_labels = []
+   max_label_len = 0
+
+   for cell in client.nb.cells:
+      # compute total code cells (for progress)
+      if cell.cell_type == 'code':
+         total_code_cells += 1
+      # map cells to their labels
+      label = nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell).get('label', '')
+      cell_labels.append(label)
+      # find max label length
+      max_label_len = max(max_label_len, len(label))
    
    # execute the cells
    for index, cell in enumerate(client.nb.cells):
-      # read cell options
-      cell_options = nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell)
-      cell_label = cell_options.get('label', '')
+      cell_label = cell_labels[index]
       padding = "." * (max_label_len - len(cell_label))
 
       # progress
--- a/src/resources/jupyter/notebook.py
+++ b/src/resources/jupyter/notebook.py
@@ -154,22 +154,28 @@ def notebook_execute(options, status):
    current_code_cell = 1
    total_code_cells = sum(cell.cell_type == 'code' for cell in client.nb.cells)
 
+   # find all the labels in the notebook
+   cell_labels = [
+      (next(iter(re.findall(r'#\| label: (.*)', cell.source)), '').strip() 
+       if cell.cell_type == 'code' else '')
+       for cell in nb.cells
+       ]
+   max_label_len = max(len(label) for label in cell_labels)
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/resources/jupyter/notebook.py [hunk 1]
```diff
@@ -159,9 +159,16 @@ def notebook_execute(options, status):
...
+        # get label
+        label = ''
+        source_lines = cell.source.split('\n')
+        label_line = next((line for line in source_lines if line.startswith('#| label:')), None)
+        if label_line:
+            label = label_line.replace('#| label:', '').strip()
+
+        status("  Cell '{0}' {1}/{2}...".format(
+           label, current_code_cell- 1, total_code_cells - 1
+        ))
```
mcanouil, line 171: It should be more reliable to get the `id` straight from the cell.
Notebooks, will always have an identifier, but the `id` is replaced by `label` if any.
This would avoid to re-do the parsing of parameters, etc.

> Notebook cell identifier. Note that if there is no cell `id` then `label` will be used as the cell `id` if it is present. See https://jupyter.org/enhancement-proposals/62-cell-id/cell-id.html for additional details on cell ids.

```suggestion
         status("  Cell '{0}' {1}/{2}...".format(
            cell.id,
            current_code_cell - 1,
            total_code_cells - 1,
         )
```

mcanouil: The above suggested code produces the following output:

image

jjallaire: I think that cells with automatically generated IDs shouldn't show the label. I also think that we should measure for the max width of a cell label and pad the other output so all progress messages have the same width (i.e. the `Done` lines up across lines).

mcanouil: I agree with the suggestion to standardise the messages.

A 80 width string filled with spaces and "Done" at the end would do?

Also other messages would need to be "width standardised".

jssonx: Dear @mcanouil @jjallaire,

Thank you for your feedback!

In the updated PR, I changed the output format and improved output alignment.

After reviewing the codebase and documentation, it seems that label information cannot be directly obtained as it has not been parsed before. Therefore, I've optimized my code using regex to extract label sets and store all cell labels from the current file in a list. This will help us during output and ensure alignment by obtaining the maximum label length.

Sample output:
```
  Cell 1/3: 'fig-polar-1'........Done
  Cell 2/3: 'fig-polar-test-2'...Done
  Cell 3/3: ''...................Done
```

mcanouil: It's why, in my opinion, `cell.id` is the best, as the more robust, way to get the label as parsed earlier already by Quarto.
The trade off is that, you get a "label" for all cells.

PS: I am not maintainer of Quarto code base, only a contributor.

jssonx: Thank you for your explanation! 

And I now have a more clear understanding of `cell.id`. Upon further research into it, I discovered that when a label is present in the cell, the `cell.id` represents the label, whereas it represents the id itself when no label is available. I must acknowledge that I had not thoroughly investigated `cell.id` prior to this, and I sincerely apologize for any oversight on my part. I truly appreciate your patience and elucidation on this matter!

Here is the updated sample output in the latest PR:
```
  Cell 1/3: 'fig-polar-1'......Done
  Cell 2/3: 'fig-polar-1024'...Done
  Cell 3/3: 'a1886270'.........Done
```

cscheid: I still think we ought to not emit `a1886270` and other automatically-generated IDs. If at that point in the code it's too late to know which ID is automatically generated and which isn't, then we need to find a way to carry this information into the right part of the code.

jssonx: Thank you for bringing this to my attention! Yes, I agree with the suggestion that perhaps there needs to be a flag for whether the ID is automatically generated or not. The reason is, according to the documentation:

define: `cell_id` is the id in the output.

```
if "#| id":
   cell_id = cell.id
elif "#| label":
   cell_id = cell.id
else:
   cell_id = <auto-generated-id>
```

Since `cell.id` is automatically generated when neither `"#| id"` nor `"#| label"` exists, the above logic is equivalent to:

```
if is_auto_generated(cell.id) == False:
   cell_id = cell.id
else:
   cell_id = <auto-generated-id>
```

Therefore, once we have a flag to know if the id is auto-generated, we can set cell_id to `""`:
```
if is_auto_generated(cell.id) == False:
   cell_id = cell.id
else:
   cell_id = ""
```

We may need to modify this part of the code as well as some code that has called this function:
[quarto-dev/quarto-cli@58a9462](https://github.com/quarto-dev/quarto-cli/blob/main/src/core/jupyter/jupyter.ts#L583-L600)

Could you kindly provide some guidance on how the relevant code should be modified? Thank you!

ps. please ignore the latest commit (`3cea7dc`) in this branch.

jjallaire: We may actually want to marry this with some work that @dragonstyle has been doing on notebook embedding. There, we want people to be able to embed notebook cells by a unique id. We have an idiom there where we first take the `label`, then use the first cell tag (if any). Perhaps we need a few field that represents this embeddable identifier?

jssonx: Thank you! While I was examining the code related to notebook embedding, I found the `nb_cell_yaml_options()` function. This led me to submit a commit ([cc1c803](https://github.com/quarto-dev/quarto-cli/pull/5772/commits/cc1c803467f794c6efb1b341eddec62b662b4b4f)) that might resolve the issue [[FR] Replace "Cell 1/N" by label (if provided) when using Jupyter #5752](https://github.com/quarto-dev/quarto-cli/issues/5752). Furthermore, I've noticed a potential problem with the `nb_cell_yaml_options()` function. Here's the summary:

**Summary of the Latest commit**

This commit introduces a feature where the render status will now display the label when a user has specified `"#| label"` for a "code" cell. When running the `"quarto render"` command, the output will match the following example:

```
  Cell 1/3: 'fig-polar-1'......Done
  Cell 2/3: 'fig-polar-1024'...Done
  Cell 3/3: ''.................Done
```

Notably, even if a user writes `"#| id"` in the "code" cell, the id will not be visible in the render status. 

This is because the return list from `nb_cell_yaml_options()` just includes cell attributes like `label`, `class`, `fig-cap`, `fig-subcap`, and so on, as specified in [cell-attributes.yml](https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/schema/cell-attributes.yml) and [yaml-intelligence-resources.json](https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/editor/tools/yaml/yaml-intelligence-resources.json), and does not include cell attributes like `"#| id"` `"#| tags"` and `"#| export"`. I'm uncertain if this is an issue that needs addressing or if there's a different approach to access these attributes. Any advice on this would be greatly appreciated!

**Additional Note**

I observed that in a recent PR ([Support export: in jupyter cells #4958](https://github.com/quarto-dev/quarto-cli/pull/4958)), we added support for `export`, but this attribute is currently absent from the official documentation [Code Cells: Jupyter](https://quarto.org/docs/reference/cells/cells-jupyter.html). I believe an update to the documentation to include this attribute might be beneficial.

jjallaire: Then `export` syntax is a pre-processing directive used by nbdev so indeed not part of our schema or feature set. It (and the others ) can indeed be safely ignored here!

jjallaire:
Thanks for all of the iteration here! This is looking great save for one thing. We end up doing this twice for each cell (once to compute the max and once to present the label for progress):

```python
nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell)
```

I wonder if we could just map the cells into a `cell_labels` array of strings at the top, compute the max, and then just index into that array during the cell-by-cell progress?

jssonx:
Thank you for your advice!

**Summary of the Latest Commit**

It now avoids calling `nb_cell_yaml_options()` twice. Here's the modified code:

```python
# compute total code cells (for progress)
current_code_cell = 1
total_code_cells = sum(cell.cell_type == 'code' for cell in client.nb.cells)

# map cells to their labels
cell_labels = [nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell).get('label', '') for cell in client.nb.cells]

# find max label length (for progress)
max_label_len = max(len(label) for label in cell_labels)
```

Further optimization is possible here by iterating client.nb.cells just once to boost efficiency, especially with large data (like using Quarto for book writing):

```python
current_code_cell = 1
total_code_cells = 0
cell_labels = []
max_label_len = 0

for cell in client.nb.cells:
   # compute total code cells (for progress)
   if cell.cell_type == 'code':
      total_code_cells += 1
   # map cells to their labels
   label = nb_cell_yaml_options(client.nb.metadata.kernelspec.language, cell).get('label', '')
   cell_labels.append(label)
   # find max label length
   max_label_len = max(max_label_len, len(label))
```

**Future Work: The output information of R is different from Python and Julia. Can we change the output information of R to the same format as Python and Julia?**

The current test results using Python and Julia are consistent, as shown below:
```
  Cell 1/3: 'fig-polar-1'......Done
  Cell 2/3: 'fig-polar-1024'...Done
  Cell 3/3: ''.................Done
```
When using R with `quarto render hello-r.qmd --to html`, the output format is as follows:

```
processing file: hello-r.qmd
1/4                   
2/4 [fig-airquality]  
3/4                   
4/4 [fig-airquality-2]
output file: hello-r.knit.md
```


Concerns with R's format:

1. It differs from Python/Julia.
2. Each cell run leads to two counts. For example, with only one code cell, the output is redundant. There should be `1/1 [fig-airquality]`.
```
processing file: hello-r.qmd
1/2                   
2/2 [fig-airquality]  
output file: hello-r.knit.md
```
3. The `hello-r.knit.md` file isn't found locally, yet the render status still displays `output file: hello-r.knit.md`.

jjallaire:
The R output format is actually coming from a totally different stack (knitr) so we don't control it directly. In more recent versions of knitr it is actually a progress bar so even more different. I'm not worried about this divergence.

jjallaire:
Thanks again so much for this (and for all of the iteration). Merging now!

jssonx:
Thank you for being so supportive!

bryanwhiting:
Thanks for building this!!

PR #5805, opened 2023-06-04
fix: do not break words in website listings

mcanouil:
## Description

This PR proposes a small change in website listing CSS code to avoid words break which can happen in various scenarios which lead to unreadable text/tables.

### Current behaviour

image

### Suggested behaviour

image


dragonstyle:
Good improvement!

PR #5841, opened 2023-06-07
ci: no longer remove stale labels on updates

mcanouil:
This PR fixes an issue in the GHA CI workflow were any updates on "needs-repro" labelled issues triggered the action which then removed the "stale" label, thus leading to a new message byt the BOT.

This PR also revert to default (`false`) for `ignore-issue-updates: false`, _i.e._, on any updates on the issue, the countdown before closing a "stale" issue is reset.

### DIFF
--- a/.github/workflows/stale-needs-repro.yml
+++ b/.github/workflows/stale-needs-repro.yml
@@ -15,7 +15,7 @@ jobs:
           stale-issue-label: "stale"
           any-of-issue-labels: "needs-repro"
           close-issue-reason: "not_planned"
-          ignore-issue-updates: true
+          ignore-issue-updates: false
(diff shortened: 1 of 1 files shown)


cderv (approved):
This looks ok. I think this behave as we want. From discussion with @mcanouil 

- Issue has a  `needs-repo` label added
- Bot passes day 15 and adds `stale` label with a comment
- the person replies with a comment - nothing changes 
- Quarto team has 14 days to look at the comment and manually remove the label any of the label. 
	- removing `stale` but not `needs-repro` will still trigger the bot resetting stale counter
	- removing both means that bot won't do anything anymore
- If after 14 days stale label is still there, then the bot will close.

PR #5857, opened 2023-06-08
fix: display metadata block even without title/subtitle

### LINKED ISSUE #5856: Quarto doesn't print authors and affiliations defined in the YAML if the title is missing
### Bug description

Authors and affiliations in the YAML header aren't printed when there's no title defined.

### Steps to reproduce

```
---
format: html
author:
  - name: Norah Jones
    affiliations:
      - ref: cmu
      - ref: chicago
affiliations:
  - id: cmu
    name: Carnegie Mellon University
    city: Pittsburgh
    state: PA
  - id: chicago
    name: University of Chicago
    city: Chicago
    state: IL
---

Foo
```

### Expected behavior

I'd expect the output to show the author and her respective affiliations.

### Actual behavior

The example above prints `Foo` only.

### Your environment

IDE: RStudio 2023.03.1+446
OS: MacOS Monterey 12.3.1

### Quarto check output

```
      Pandoc version 3.1.1: OK
      Dart Sass version 1.55.0: OK
      Version: 1.3.361

      Version: 3.9.12
      Path: /usr/local/opt/python@3.9/bin/python3.9
      Jupyter: (None)

      Version: 4.2.0
      LibPaths:
      knitr: 1.42
      rmarkdown: 2.21

```

mcanouil:
This PR tweaks the HTML template to include the metadata partial as header if no title or subtitle are provided.

Fixes #5856

### DIFF
--- a/src/resources/formats/html/pandoc/template.html
+++ b/src/resources/formats/html/pandoc/template.html
@@ -37,9 +37,7 @@ $title-block.html()$
 $elseif(subtitle)$
 $title-block.html()$
 $elseif(by-author)$
-<header id="title-block-header" class="quarto-title-block default">
 $title-block.html()$
-</header>
 $endif$
 
 $if(toc)$
--- a/src/resources/formats/html/pandoc/template.html
+++ b/src/resources/formats/html/pandoc/template.html
@@ -36,6 +36,10 @@ $if(title)$
 $title-block.html()$
 $elseif(subtitle)$
 $title-block.html()$
+$elseif(by-author)$
+<header id="title-block-header" class="quarto-title-block default">
+$title-block.html()$
+</header>
 $endif$
 
 $if(toc)$
--- a/src/resources/formats/html/pandoc/template.html
+++ b/src/resources/formats/html/pandoc/template.html
@@ -32,13 +32,7 @@ $for(include-before)$
 $include-before$
 $endfor$
 
-$if(title)$
 $title-block.html()$
-$elseif(subtitle)$
-$title-block.html()$
-$elseif(by-author)$
-$title-block.html()$
-$endif$
 
 $if(toc)$
 $toc.html()$
--- a/src/resources/formats/html/templates/title-block.html
+++ b/src/resources/formats/html/templates/title-block.html
@@ -1,6 +1,8 @@
 <header id="title-block-header" class="quarto-title-block default">
 <div class="quarto-title">
+$if(title)$
 <h1 class="title">$title$</h1>
+$endif$
 $if(subtitle)$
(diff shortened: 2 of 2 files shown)


src/resources/formats/html/pandoc/template.html
```diff
@@ -36,6 +36,8 @@
 $title-block.html()$
 $elseif(subtitle)$
 $title-block.html()$
+$elseif(by-author)$
```
dragonstyle, line 39: I am thinking that this approach would screen out cases where there is no title, no subtitle, and author normalization is disabled, but the same point could be made about including a date (e.g. user put only a date and it didn't appear).

I would suggest we should just always include `title-block` without condition - what do you think?

mcanouil: Makes sense, I made a small change in that direction: cfa2917b120af88bca230298fa9cb5bcdc80913a

mcanouil: I made the change 1704e9aec878dec9cf91241af83c18e195a9b6ae and did few manual testing on simple HTML document.
Seems good.

cderv (commented):
Just thinking that this could lead to possible breakage or at least changes in output for users that would not have a `title` nor a `subtitle`. This will lead to a new `header` HTML node in the document. 

We should just write this in the changelog so that readers are aware. If we care about sharing such changes when we know it happens 😅 

users impacted by this should be rare though though.

mcanouil:
> We should just write this in the changelog so that readers are aware. If we care about sharing such changes when we know it happens 😅

Yes, that was my intent.
I usually fill the changelog once all agreed on changes made by the PR.

cderv:
Perfect ! 

I am not sure what we usually include in changelog. Sometime only the we fix an issue and add a feature like 

* metadata block will now be included in output even when no `title` or `subtitle`

This does not specifically explain that new `<header>` block will be added and will modify HTML outputs.  That is why I wanted to add the thoughts. But maybe this is not important. 

> I usually fill the changelog once all agreed on changes made by the PR.

Yep you're right to do that. 👍 I was not implying it was missing ! 😉

dragonstyle:
Ahh sorry I merged without the changeling addition - can you still add that in a separate commit @mcanouil ?

mcanouil:
Only via a new PR.

dragonstyle:
I can make that change since it's my fault. Sorry about that

mcanouil:
Feel free to make a commit directly in main.
Or I can make a PR in few hours.
Up to you.

cderv:
> I usually fill the changelog once all agreed on changes made by the PR.

See, you must be really quick here ! 😆

PR #5889, opened 2023-06-12
ci: never auto-close needs-repro/stale issues

mcanouil:
As briefly discussed with @jjallaire, let's not close issues with "needs-repro" and "stale".
Only keep the bot message to ask again for a reproducible example after 14 days, then "stale" after an additional 14 days.
We'll know from the "stale" label that there were no update, but the issue will remain open until we're absolutely sure, there is no bug, thus manually closing it.


cderv (approved):
All good. TBH I prefer this 😄 

Thanks for the update !

mcanouil:
I know, I had the same feeling 😉, _i.e._, felt a bit harsh to close without a human being commenting at least 😅

ghost (approved):

allenmanning:
Great work here @mcanouil, thank you.

PR #5909, opened 2023-06-13
Avoid linebreak after 'external link' icon

### LINKED ISSUE #4619: `link-external-icon` should not be line-wrapped
### Bug description

I just tried out the [external links](https://quarto.org/docs/output-formats/html-basics.html#external-links) feature of Quarto's HTML format and noticed that the icon after external links gets line-wrapped like regular whitespace-separated text which is clearly undesirable.

Example screenshot:

Bildschirmfoto am 2023-03-03 um 17 30 47

I see that technically no whitespace is used between the link and the icon. Instead it is inserted via the CSS pseudo-element [`::after`](https://developer.mozilla.org/en-US/docs/Web/CSS/::after). I'm no CSS expert, so I can't judge what the best solution looks like. A possible solution would be to [move `display:inline-block;` from `a.external::after` to `a.external`](https://stackoverflow.com/questions/14851320/how-to-avoid-line-breaks-after-before-in-css). But maybe you got an even better solution.

Thanks for looking into this!

<sup>I'm using the latest stable Quarto CLI v1.2.335</sup>

### Checklist

- [ ] Please include a minimal, fully reproducible example in a single .qmd file? Please provide the whole file rather than the snippet you believe is causing the issue.
- [ ] Please document the operating system you're running. If on Linux, please provide the specific distribution.

### LINKED ISSUE #5639: link-external-icon wrapping has unexpected breaks
### Bug description

````markdown
---
title: external
format:
  html:
    link-external-icon: true
---

This is a fairly lengthy sentence that has an external link to the website of the [Quarto project](https://quarto.org/).
````

When rendered, things look just fine at first:

image

With a slightly different screen size, the period wraps by itself:

image

Even narrower sees the external-link icon wrap apart from the link text:

image

I did not expect that the link text, the external link icon, and the trailing period would be splittable.

```
      Pandoc version 3.1.1: OK
      Dart Sass version 1.55.0: OK
      Version: 1.3.340
      Path: /Applications/RStudio.app/Contents/Resources/app/quarto/bin

      Version: 3.9.6
      Jupyter: (None)

      Version: 3.6.3
      LibPaths:
        - /Library/Frameworks/R.framework/Versions/3.6/Resources/library
      knitr: 1.42.8
      rmarkdown: 2.20

```

### Checklist

### PR DESCRIPTION
**Pull request by rich-iannone**
This change addresses an unwanted line break between the link text and the outward arrow icon within those links marked as *external* links. Smoke test added in `tests/docs/smoke-all/2023/06/13`.

Fixes: https://github.com/quarto-dev/quarto-cli/issues/4619
Fixes: https://github.com/quarto-dev/quarto-cli/issues/5639

### DIFF
--- a/tests/docs/smoke-all/2023/06/13/issue-4619.qmd
+++ b/tests/docs/smoke-all/2023/06/13/issue-4619.qmd
@@ -1,9 +1,0 @@
----
-format:
-  html:
-    link-external-icon: true
----
-
-This is a (marked) external link: [Quarto](https://quarto.org/){.external target="_blank"}.
-
-Size the viewport horizontally to observe the breaking behavior.
--- a/tests/docs/smoke-all/2023/06/13/issue-4619.qmd
+++ b/tests/docs/smoke-all/2023/06/13/issue-4619.qmd
@@ -1,0 +1,9 @@
+---
+format:
+  html:
+    link-external-icon: true
+---
+
+This is a (marked) external link: [Quarto](https://quarto.org/){.external target="_blank"}.
+
+Size the viewport horizontally to observe the breaking behavior.
--- a/src/resources/formats/html/_quarto-rules-link-external.scss
+++ b/src/resources/formats/html/_quarto-rules-link-external.scss
@@ -1,26 +1,19 @@
 a.external:after {
-  display: inline-block;
-  height: 0.75rem;
-  width: 0.75rem;
-  margin-bottom: 0.15em;
-  margin-left: 0.25em;
   content: "";
-  vertical-align: -0.125em;
   @if variable-exists(link-color) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: tests/docs/smoke-all/2023/06/13/issue-4619.qmd [hunk 1]
```diff
@@ -0,0 +1,9 @@
+---
```
allenmanning, line 1: Nice test.

dragonstyle: This test will really only check that the rendering doesn't error AFAIK (e.g. the test runner will look for either an error or warning during the rendering of the document). To truly verify this, you'd want to look at the generated CSS for the page, but we don't really have a great way to do that right now (and really I don't even think that we'd want to look at the rendered CSS, but instead 'see' that the external icon doesn't break, since that is really what we want to test).

I'd recommend that we remove the test for now - I don't think its adding much and will make the test runner a bit slower.

dragonstyle (approved):

rich-iannone:
Thanks @dragonstyle and @allenmanning for looking at this PR. Removed the smoke test and will merge once CI tests pass.

PR #5923, opened 2023-06-14
Add larger offset between any link's underline to avoid collision with underscores

rich-iannone:
This PR is a SCSS adjustment that applies to all `<a>` tags and sets a `text-underline-offset` length value of `3px` to avoid overstriking commonly-used underscore characters. The offset value is lower than a typical underscore but not moved so far as to appear disconnected from the text. 

This is the reporter's screenshot of the HTML render:

text-underline-reporter-screenshot

and here's my screenshot of a render with the change applied

text-underline-offset-avoid-underscores

Fixes: https://github.com/quarto-dev/quarto-cli/issues/5641

### DIFF
--- a/src/resources/formats/html/_quarto-rules.scss
+++ b/src/resources/formats/html/_quarto-rules.scss
@@ -385,6 +385,11 @@ div.sourceCode > iframe {
   }
 }
 
+// link styling
+a {
+  text-underline-offset: 3px;
+}
+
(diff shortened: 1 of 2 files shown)


rich-iannone:
@dragonstyle as with the other issue, I now think the test I added here isn’t needed. I will remove that shortly.

PR #6010, opened 2023-06-23
ci: remove "close notes" from the stale message

mcanouil:
This PR follows #5889 and removes the close notes in the stale message that no longer applies.

PS: prefer the use of "squash + merge" option to merge this PR.
This allows better history by keeping it simple and keeping the PR id in git log.


cderv:
Thanks

PR #6044, opened 2023-06-27
add Posit Cloud PublishProvider

mslynch:
## Description

Adds support for publishing to Posit Cloud by implementing a new `PublishProvider`.

## Checklist

I have (if applicable):

### DIFF
--- a/src/publish/common/bundle.ts
+++ b/src/publish/common/bundle.ts
@@ -12,8 +12,14 @@ import { Tar } from "archive/tar.ts";
 
 import { PublishFiles } from "../provider-types.ts";
 import { TempContext } from "../../core/temp-types.ts";
-import { md5Hash } from "../../core/hash.ts";
+import { md5HashBytes } from "../../core/hash.ts";
 
+/** Creates a compressed bundle file in the format required by Posit Connect and Cloud.
+ * @param type Whether this is a site or document.
+ * @param files Information on what should be included in the bundle.
+ * @param tempContext Temporary directory where file operations will be performed.
+ * @returns The absolute path of the bundle file.
+ */
 export async function createBundle(
   type: "site" | "document",
   files: PublishFiles,
@@ -24,8 +30,9 @@ export async function createBundle(
   for (const file of files.files) {
     const filePath = join(files.baseDir, file);
     if (Deno.statSync(filePath).isFile) {
-      const f = Deno.readTextFileSync(filePath);
-      manifestFiles[file] = { checksum: md5Hash(f) as string };
+      const fileBytes = Deno.readFileSync(filePath);
+      const checksum = md5HashBytes(fileBytes);
+      manifestFiles[file] = { checksum };
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/publish/common/bundle.ts
```diff
@@ -24,8 +30,9 @@ export async function createBundle(
   for (const file of files.files) {
     const filePath = join(files.baseDir, file);
     if (Deno.statSync(filePath).isFile) {
-      const f = Deno.readTextFileSync(filePath);
-      manifestFiles[file] = { checksum: md5Hash(f) as string };
```
mslynch, line 28: I found that this was calculating the wrong checksum – it only works because Posit Connect doesn't validate it. `md5Hash` works with strings, but I couldn't find a way to get it to work properly using the contents of a file. Since Deno's crypto module supports MD5, there's probably no reason to keep `blueImpMd5` – but I'll leave that to a future PR.

src/publish/posit-cloud/posit-cloud.ts
```diff
@@ -0,0 +1,304 @@
...
+
+async function authorizeToken(
+  _options: PublishOptions,
+  _target?: PublishRecord,
+): Promise<AccountToken | undefined> {
+  // get credentials
+  while (true) {
+    const token = await Input.prompt({
+      message: "Token:",
+      hint: "Learn more at https://posit.cloud/learn/guide#account-tokens",
```
jjallaire, line 124: I think here and in the subsequent hint you should be directing the user to the exact page where they provision a token and secret for their account. With this link, they get once sentence about tokens, then need to realize that they should click "Publish Outputs", then read some general conceptual material about Publish Outputs, then realize that they need to skip past "Publish from a Cloud Project", then ???. I think this link should land them directly on the provisioning page and if additional explanation is required it should be there inline.

jjallaire:
Implementation looks great! My only 2 comments are:

(1) Previous note about bringing the user directly to the token provisioning page (if you don't have this yet then I'd still put the right URL here and then just redirect for the time being)

(2) I'm guessing you will want to hide this for the time being. The only place the user will see "Posit Cloud" if they aren't directly looking for it is in the interactive list of providers. You can filter Posit Cloud out this list here: https://github.com/quarto-dev/quarto-cli/blob/main/src/command/publish/cmd.ts#L226. Note that users will still be able to publish via `quarto publish posit-cloud`, they just won't see it in the menu.

src/publish/posit-cloud/posit-cloud.ts
mslynch: Posit Cloud doesn't have such a URL at the moment, unfortunately... The URL for that page is specific to the user's account. I filed an issue for us to add a generic tokens URL and took out the `hint`.

jjallaire (changes requested):
This is good to go after you make the change to not have posit cloud show up in the providers prompt (at least until after you launch)

jjallaire:
You can't remove it entirely from `publishProviders()` (else it won't work in all cases). You need to filter it out when you call for that to build the prompt UI (see link I sent before here: https://github.com/quarto-dev/quarto-cli/blob/main/src/command/publish/cmd.ts#L226)

PR #6167, opened 2023-07-11
Adapt checking knitr capabilities to .Rprofile that would `cat()` some value

### LINKED ISSUE #6154: quarto check on command line cannot find R installation if `.Rprofile` is present
### Bug description

I noticed that if I have an `.Rprofile` present in my R environment that on the command line

```
quarto check
```

fails to find the R installation.  If I remove `.Rprofile` `quarto check` does find the R installation.

Update: I should have noted that Quarto documents with R code chunks work just fine regardless of what `quarto check` reports, so the problem is with the `check` code itself.

### Steps to reproduce

Put a `.Rprofile` file in `/Users/your_user_name` on a Mac

### Expected behavior

`quarto check` should find the R installation.

### Actual behavior

`quarto check` cannot find the R installation if a `.Rprofile` is present; see the Quarto check output below.

### Your environment

Using the command line with macOS 13.4.1 on an M2 chip.

### Quarto check output

This shows the "error" condition, seen when `.Rprofile` is present.

```bash
~> quarto check

      Pandoc version 3.1.2: OK
      Dart Sass version 1.55.0: OK
      Deno version 1.33.2: OK
      Version: 99.9.9
      Path: /Users/bryanh/Documents/Professional/Research/R_Pkgs/quarto-cli/package/dist/bin

      Version: 3.9.6
      Path: /Applications/Xcode.app/Contents/Developer/usr/bin/python3
      Jupyter: (None)

(|) Checking R installation...........✔ All clear for ChemoSpec2D, exCon, FuncMap, HiveR, LearnPCA, LindenmayeR, readJDX, roxut, SpecHelpers!
✖  Package with errors on CRAN: 
  - ChemoSpec (2)
★  Package with notes on CRAN: 
  - ChemoSpecUtils (1)

```

cderv:
This closes #6154 

The case is a user `.Rprofile` that would contain some `cat()` call. This would add some lines in `stdout`, but we currently expect all the `stdout` to be our JSON string. 

This makes the JSON parsing fails and R reported as not found (while `quarto render` works ok)

This PR adds special string content that we can match easily to unwrap the JSON content from stdout to check the capabilities. 

@jjallaire adding you as review to check this is ok to do that. I believe this is without side effect.

### DIFF
--- a/news/changelog-1.4.md
+++ b/news/changelog-1.4.md
@@ -76,6 +76,7 @@
 - Exit if project pre or post render script fails
 - Use InternalError in typescript code, and offer a more helpful error message when an internal error happens.
 - ([#6042](https://github.com/quarto-dev/quarto-cli/issues/6042)): Correctly support empty lines in YAML blocks.
+- ([#6154](https://github.com/quarto-dev/quarto-cli/issues/6154)): `quarto check knitr` does not fail anymore when user's `.Rprofile` contains `cat()` calls.
 
 ## Docusaurus Format
 
--- a/src/core/knitr.ts
+++ b/src/core/knitr.ts
@@ -44,8 +44,8 @@ export async function knitrCapabilities() {
     });
     if (result.success && result.stdout) {
       const jsonLines = result.stdout
-        .replace(/^.*--- JSON_START ---/sm, "")
-        .replace(/--- JSON_END ---.*$/sm, "");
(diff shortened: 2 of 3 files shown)


jjallaire:
This approach LGTM! You may want to call the delimiter `YAML-LINES` as that's the format used here rather than JSON.

cderv:
Oh indeed. I completely mix it up. I'll change and merge. Thanks !

PR #6248, opened 2023-07-17
Detect language within fenced code block when echo: fenced 

cderv:
Detect language within fenced code block when echo: fenced  is used for code annotationWelcome to the quarto GitHub repo!

closes #6244 

This suppose that when `echo: fenced` there won't be any language on cell and so only class found will be `cell-code`. In that case we extract from the text content of CodeBlock the language between `` ```{{lang}} `` 

We could also try to apply the regex no matter the language on the CodeBlock but it felt too wide...

This PR also adds support for HTML and Markdown code cells using HTML comments as comment chars.

### DIFF
--- a/news/changelog-1.4.md
+++ b/news/changelog-1.4.md
@@ -134,3 +134,4 @@
 - ([#5902](https://github.com/quarto-dev/quarto-cli/issues/5902)): Support paired shortcode syntax.
 - ([#6013](https://github.com/quarto-dev/quarto-cli/issues/6013)): Don't error if citation is passed as a boolean value in metadata via flags
 - ([#6207](https://github.com/quarto-dev/quarto-cli/issues/6207)): When QUARTO_R is set to a non-existing path, a warning is now thrown like with QUARTO_PYTHON. Quarto still fallback to search a working R version.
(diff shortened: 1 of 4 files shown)


dragonstyle:
🔥

dyavorsky:
@cderv As of 2024.04.02 on Quarto version 1.4.552, I don't believe the support for markdown cells in Code Annotation has been fixed.

In the following example, code annotation does not work when the language is "markdown" but it does work when the language is most other things, eg "r"

`````
```markdown
# Header        # <1>
This is a test  # <2>
```
1. a header
2. some content
`````

**Comment by cderv** [comment]
The right comment chars needs to be used. This works 
````markdown
```markdown
# Header       <!-- <1> -->
This is a test  <!-- <2> -->
```
1. a header
2. some content

````

as in the tests shown in this PR. 

HTML use also HTML comment. 
LaTeX uses `%` so this will work
````markdown
```latex
\usepackage{graphics}  % <1>
```
1. a header
````

dyavorsky:
This make sense.  Very helpful -- I appreciate the fast reply!  Can I [buy you a coffee](https://www.buymeacoffee.com/) as a apology for calling you out? :)

PR #6277, opened 2023-07-19
Run CI tests in parallel to reduce overall CI time

cderv:
This PR brings lots of changes to how tests are run on CI, but also impact running tests locally and in parallel. 

I tried to document everything in the README (version from this PR here: https://github.com/quarto-dev/quarto-cli/blob/feature/parallel-smokes/tests/README.md) 

## The concept

Idea is to build on `run-parallel-tests.sh` idea to create buckets of tests and then based on that create a Matrix for GHA so that we can start several jobs runner (1 per bucket and per OS) in order to split our tests in GHA in parallel runners. 

At the end, this is working and we decrease overall time even if not perfect. 

## Changes to make that happens 

* `test-smokes-parallel.yml` is now the main CI workflow we trigger. This will run `run-parallel-tests.sh` to get a JSON string for buckets of tests, and use this to create the Matrix that is used to call our `test-smokes.yml` using `workflow_call` event as many time as runner we need. 

* To allow that `run-parallel-tests.ts` has evolved. 
  * It now takes argument, and has a specific flag `--json-for-ci` to create the adequate JSON string. 
  * Challenge was about splitting smoke-all tests documents. So now we can time individual tests and we use that when we create buckets for CI. Otherwise, we put `smoke-all.test.ts` as untimed in its own bucket. The later applies to local run of parallel tests. 

* `test-smokes.yml` behave also differently when run from a `workflow_call` than scheduled tests. The later is still sequential running (one call to `./run-tests.sh`). The former is a for-loop around `run-tests.sh` or `./run-tests.ps1` to run each test in the buckets. 

* Doing that required some changed in `./run-tests.sh`, especially how we run smoke-all tests. 
  * So now you can do `./run-tests.sh docs/smoke-all/2022/**/*.qmd` and the `smoke-all.test.ts` is not needed. (`/run-tests.ps1` can also do that)
  * Timing test will now compute timing for each smoke-all document and record in a file
  * Timing file can now be configured with the env var. 

I may have missed some highlights but as I said 
* It should be documented in README
* There are some comments in the scripts file 
* And code speaks for itself. 

Probably some minor tweaking to do. @cscheid @dragonstyle I would love your review on this. Main concerns

* Does it seems ok to you ? 
* Does `run-tests.sh` still works for you locally on mac ? Same for parallel-tests
* Did I missed anything ? 
* Is there any question ? Which would mean it requires more documentation. 

## Left to do 

- [ ] @cscheid as you can see I have now a `timing-for-ci.txt` which detailed tests. I did not overwrite your current timing.txt for now. If the new process fits we can probably have only one file with the detailled timing for both CI and local runs. 
- [ ] Are we ok on the trigger event for the YAML ? Scheduled daily smoke test is still a sequential run as we  don't have timing constraint on this IMO. 
- [ ] Maybe clean the git history of this branch before merging ? Don't know how we usually do it but there are lots of tests commits in there !! Or maybe squash this one ? 

## Later improvmeent

- Add a specific workflow to auto update the timing file directly in CI runners and commit the new timing for our following tests. 

Thanks ! 

(and hopefully this PR won't fail on CI 😅 )


cderv:
I opened PR from my fork and not from the branch in the repo 🤦 Closing and reopening sorry !

Issue #6161, opened 2023-07-10
Add known feature limitations to documentation 


cderv:
Opening this after asking to @cwickham and following https://github.com/quarto-dev/quarto-cli/discussions/6140 

We need a good way to add known limitation to docs. Currently we have several patterns I think 

*  "Knownlimitation" 
	* https://quarto.org/docs/prerelease/1.3/tables.html#limitations, 
	* https://quarto.org/docs/prerelease/1.4/typst.html#known-limitations
* "Format Support" 
    * https://quarto.org/docs/authoring/callouts.html#format-support

Having a common way to document it would ease the discovery I guess.

Ideas: 

* Special common callouts
* A common sections 
* More subtle using collapse content 
* or even on-hover content if we want to "hide" by default but still be easily findable. 

TBD

This issue can also be a recipient to track all limitation to be documented 

## Known limitations: 

- [ ] https://github.com/quarto-dev/quarto-cli/discussions/6140

PR #6297, opened 2023-07-20
docs: add limitations related to `downlit`

mcanouil:
This PR adds limitations notes with the use of `downlit` via `code-link: true`.

Fixes #6199


src/resources/schema/cell-codeoutput.yml
```diff
@@ -91,6 +91,7 @@
     short: "Include line numbers in code block output (`true` or `false`)"
     long: |
       Include line numbers in code block output (`true` or `false`).
+      Limitation: does not work properly with `code-link` set to `true`.
```
cderv, line 94: `code-line-numbers` always works. It is `code-link` which does not work if `code-line-numbers` is set. 

So I don't think adding that here in `code-line-numbers` reference page does really help. Also because it will be shown in other reference page than `html` and the limitation is only in HTML (as `code-link` is only for HTML and R documents)

src/resources/schema/document-code.yml
```diff
@@ -45,6 +48,7 @@
     long: |
       The style to use when displaying code annotations. Set this value
       to false to hide code annotations.
+      Limitation: does not work properly with `code-link` set to `true`.
```
cderv, line 51: Same here are for `code-line-numbers`. 

I believe code annotations works ok - it is just that there will be no code linking applied by **downlit**

Try this 

````markdown
---
format: 
  html:
    code-link: true
    code-annotations: hover
---

```{r}
#| message: false
library(dplyr) # <1>
first(mtcars) # <2>
last(1:10)
```

1. Test
2. Test
````

and this 

````markdown
---
format: 
  html:
    code-link: true
---

```{r}
#| message: false
library(dplyr)
first(mtcars)
last(1:10)
```
````

I don't think noting that `code-annotations` does not work properly is the same. This is `code-link` that won't work.

File: src/resources/schema/document-code.yml [hunk 3]
```diff
@@ -32,7 +32,10 @@
       Code linking is currently implemented only for the knitr engine 
       (via the [downlit](https://downlit.r-lib.org/) package). 
       A limitation of downlit currently prevents code linking 
-      if `code-line-numbers` is also `true`.
+      if `code-line-numbers` is also `true` and proper display
+      of code annotations if `code-annotations` is also set to `true`.
+      For `code-line-numbers` and `code-annotations` to work properly,
+      it is recommended to set `code-link` to `false`.
```
cderv, line 38: I believe we should indeed add the limitation in there for `code-link` description. The limitation with `code-annotations` is missing. Basically the limitation is: 

`code-link: true` has no effect when `code-line-numbers` or `code-annotations` features are used.

cderv (commented):
Thanks. 

I added some comments because I don't think that the addition really clarifies the situations. 

Basically the current limitation due to downlit procesing is that 
`code-link: true` has no effect when `code-line-numbers` or `code-annotations` features are used. 

But both `code-line-numbers` and `code-annotations` features works well. Do you have an example where they don't work ? 

This is a limitation for code-linking features which uses R packages and this is for now added to 
https://quarto.org/docs/output-formats/html-code.html#code-linking

We could also add mention of `code-annotations` impact in there IMO.

src/resources/schema/document-code.yml
mcanouil: See <https://github.com/quarto-dev/quarto-cli/discussions/6187>.

I would not say discarding a parameter and outputting a warning is the definition of `code-annotations` and `code-link` working properly.

src/resources/schema/cell-codeoutput.yml
mcanouil:
```suggestion
```

mcanouil: Fair point, I overlooked and missed `formats: [$html-all, ms, $pdf-all]`.

Reverted in 3c185cda684c637464951da9f9bf106ff33473f5

src/resources/schema/document-code.yml
mcanouil: I reworded a bit the limitation sentence, to make it simpler and more accurate.

9f77c75b3911375c62e20843ba16007a00b97321

src/resources/schema/document-code.yml
mcanouil: How about changing to:
> When `code-annotations` is set to `true`, `code-link` is automatically disabled due to downlit current limitations (*for HTML-based formats using knitr as the engine*).

cderv: I saw #6187  - there is no example in it, and nothing that shows that Code Annotations feature is not working as it should. 

> I would not say discarding a parameter and outputting a warning is the definition of code-annotations and code-link working properly.

And I did not say (or at least meant) that both are working correctly. I am saying that `code-annotations` is working ok but if you add `code-link` then we know code linking won't work. So we emit a warning so that the user know when rendering (and not discover later that no code linking is in the doc). 

To me the limitation is to `code-link` and the information should be in this reference description. It will be only shown on HTML Reference, and really is the limitation. 

> It's a failsafe. User should not have to be aware of that after the facts.

I am not sure what we could do more when two features don't work together than 

* do not error and proceed with deactivating the conflicting feature
* Warn user so that it is known and not silently tweaked
* Document `code-link` limitation. 

Overall, we should indeed document limitations - that is the purpose of main epic https://github.com/quarto-dev/quarto-cli/issues/6161

Just the place of it seems to me important, to not give the impression that a feature does not work. (and by working I mean the individual feature doing what it is supposed to do - not no warning whatsoever)

src/resources/schema/cell-codeoutput.yml
mcanouil: 3c185cda684c637464951da9f9bf106ff33473f5

(VSCode PR extension did a weird thing with my replies...)

src/resources/schema/document-code.yml
mcanouil: I agree, let's only document these limitation under `code-link`.

mcanouil: reverted in ad6c8f7c42b850d666403423c6d6a275a654c3bb

mcanouil:
As suggested, I made a companion PR:
- quarto-dev/quarto-web#793

src/resources/schema/document-code.yml
```diff
@@ -31,8 +31,10 @@
 
       Code linking is currently implemented only for the knitr engine 
       (via the [downlit](https://downlit.r-lib.org/) package). 
-      A limitation of downlit currently prevents code linking 
-      if `code-line-numbers` is also `true`.
+      A limitation of downlit prevents code linking 
+      if `code-line-numbers` and/or `code-annotations` are set to `true`.
+      For `code-link` to work properly, `code-line-numbers` and `code-annotations`
+      require to be set to `false`.
```
cderv, line 37: Don't know if you tried but slight side effect.... setting `code-annotations: false` does also prevent code linking it seems if the annotations are present in the chunk. 

Basically, if one wants code linking in a cells, it should not contains any annotation using `# <1>` syntax. 

I don't know why `code-annotations: false` make code linking not work because I would have thought the HTML would not be changed in that case. (which is the underlying issue with downlit)

I would say that the advice would be 

* For `code-link` to work properly, `code-line-numbers` should not be activated on the cell and   No code annotations syntax added.  

But again, probably a bug that `code-annotations` still do something... unless this is the `# <1>` syntax that makes **downlit** fails... 🤔 

Are you observing like me ?

mcanouil: FYI, I made a repository to check various combinations (but only when defining the option in the yaml frontmatter): https://github.com/mcanouil/quarto--code-link--code-annotations

mcanouil: After checking, code linking is not preventing by the use of code annotations, it is the other way around as I initially described, _i.e._, code annotation does not work properly when `code-linking: true`.
image

mcanouil: FYI, for `code-line-numbers`, see <https://github.com/mcanouil/quarto--code-link--code-line-numbers>.
This indeed prevents code linking.

cderv: Really weird - I can indeed reproduce and see that code links works in your example, but I maintain that in my previous example it still does not. Did you try ?  https://github.com/quarto-dev/quarto-cli/pull/6297#discussion_r1270405944

Simplified example 
````markdown
---
format: 
  html:
    code-link: true
    code-annotations: true
---

```r
stats::rnorm(1) # <1>
stats::rnorm(2) # <2>
```

1. Test 1
2. Test 2 

````

image


So we should definitely try every possible case and check how that works - for now I don't understand why this would work in the above and not in your case. Looking into it. 

Thanks for the repo by the way !

cderv: Sorry it has been a while. Do we need to check more cases for theses options before merging ?

mcanouil: I have to say, indeed it's been a while and I did not update https://github.com/mcanouil/quarto--code-link--code-line-numbers with your last example.
I'll have to add those tests and check the results.

cscheid:
I think this change should be done differently; we should make a short note about how it's implemented and then link to some page in quarto.org where we have a full explanation. I don't want the schema description to become an essay.

PR #6321, opened 2023-07-24
Don't add code annotation post processor for PDF when deactivated

### LINKED ISSUE #6320: opting out code annotations in PDF format still include custom command
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/6316

<sup>Originally posted by **DillonHammill** July 24, 2023</sup>
### Description

I would like to display code annotations with hover in rendered HTML format but NOT in PDF format. I have included the following options in my `_quarto.yml`:

```{r}
format:
  html:
    theme: cosmo
    code-annotations: hover
  pdf:
    documentclass: scrreprt
    header-includes:
      - \usepackage{setspace}\doublespacing
      - \usepackage{ragged2e}
    code-annotations: false
```

I am able to successfully render HTML with code annotations but I encounter errors when trying to render the PDF without code annotations:
```{r}
Rendering PDF
running xelatex - 1
  This is XeTeX, Version 3.141592653-2.6-0.999994 (TeX Live 2022) (preloaded format=xelatex)
   restricted \write18 enabled.
  entering extended mode
  
updating tlmgr

updating existing packages
ERROR: 
compilation failed- error
Undefined control sequence.
<argument> \circled 
                    {1}
l.5010 ... }\hspace*{\fill}\NormalTok{\circled{1}}
```

Also is there any plan to treat code annotations as standard text that can be arranged with code/figures (e.g. on the right of a code block)?</div>

cderv:
fix #6320 

See detailled analysis in https://github.com/quarto-dev/quarto-cli/issues/6320#issuecomment-1648065596

I believe we need to avoid adding the post processor for LaTeX when `code-annotations: false` because we won't do any Lua processing in that case 
https://github.com/quarto-dev/quarto-cli/blob/49f1bd16e341f59087b346481756be50281419da/src/resources/filters/quarto-pre/code-annotation.lua#L511-L524

### DIFF
--- a/src/format/pdf/format-pdf.ts
+++ b/src/format/pdf/format-pdf.ts
@@ -51,6 +51,7 @@ import { isLatexPdfEngine, pdfEngine } from "../../config/pdf.ts";
 import { formatResourcePath } from "../../core/resources.ts";
 import { kTemplatePartials } from "../../command/render/template.ts";
 import { copyTo } from "../../core/copy.ts";
+import { kCodeAnnotations } from "../html/format-html-shared.ts";
 
 export function pdfFormat(): Format {
   return mergeConfigs(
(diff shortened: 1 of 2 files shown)


cderv:
From @mcanouil comment in https://github.com/quarto-dev/quarto-cli/discussions/6316#discussioncomment-6530721 it seems the we don't get the expected result when `code-annotations: none` as 
````markdown
---
format:
  pdf:
    code-annotations: none
---

```r
1 + 1 # <1>
2 + 2 
```

1. A comment

````

gets me 
image


* Post processor does happen and add the `\circled` command
* But Lua filter runs and correctly insert the preamble

It seems like this should also be fixed... :think

cderv:
About my last comment, I opened a new issue to keep things scoped. 

* https://github.com/quarto-dev/quarto-cli/issues/6332

This means this PR is only about fixing `code-annotations: false`

PR #6558, opened 2023-08-19
Revert to Pandoc 3.1.2

jjallaire:
We've noticed that new behavior introduced in Pandoc 3.1.4 affects `--id-prefix` such that IDs in visual editor documents all get the prefix (as opposed to just footnotes). This issue is reported here: https://github.com/quarto-dev/quarto-cli/issues/6548

This PR takes us back to a previous version of Pandoc while we wait for some upstream change/resolution.

### DIFF
--- a/configuration
+++ b/configuration
@@ -10,7 +10,7 @@
 # Binary dependencies
 export DENO=v1.33.4
 export DENO_DOM=v0.1.35-alpha-artifacts
-export PANDOC=3.1.6.1
+export PANDOC=3.1.2
 export DARTSASS=1.55.0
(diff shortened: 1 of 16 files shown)


jjallaire:
@dragonstyle This isa PR to temporarily roll back to Pandoc 3.1.2 (while we wait for a resolution in the breakage of `--id-prefix` in Pandoc 3.1.4, see https://github.com/jgm/pandoc/issues/9008). I noticed there are some tests failing related to Manuscripts and notebook embedding, any idea what might be causing these?

Issue #4746, opened 2023-03-09
monobackgroundcolor being ignored


dragonstyle:
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/4736

<div type='discussions-op-text'>

<sup>Originally posted by **harrelfe** March  9, 2023</sup>
I can't get monobackgroundcolor to take effect as described in https://quarto.org/docs/output-formats/html-themes.html#sass-variables.  Here is a reproducible example.

````
---
title: "Test"
format:
  html:
    monobackgroundcolor: red
    backgroundcolor: cornsilk
    self-contained: true
    toc: true
    toc-location: left
    toc-title: Contents	
---

Here is some code: `xvariable <- pi`.  Regular text follows.

````
</div>

dragonstyle:
@cderv I'm assigning us both b/c you really found the core issue - if you have time / inclination to fix, feel free. If not, I'll try to grab it sometime tomorrow AM!

cderv:
It seems you've done it! 😉 Thanks !

PR #6782, opened 2023-09-09
Add header html tag to CSS for `monobackgroundcolor`

### LINKED ISSUE #6781: `monobackgroundcolor` ignored by document title and section title
### Bug description

As the title suggests, `monobackgroundcolor` seems to be ignored by by document title and section title.

### Steps to reproduce

````qmd
---
title: "Document title with `code`"
format:
  html:
    monobackgroundcolor: "transparent"
---

# Section title with `code`

Some text with `code`
````

### Expected behavior

`monobackgroundcolor` should affect document title and section title as well.

### Actual behavior

`monobackgroundcolor` does not affect document title or section title. Example below:

图片

### Your environment

- IDE: RStudio 2023.06.0 Build 421
- OS: Ubuntu 22.04

### Quarto check output

```bash
quarto check

      Pandoc version 3.1.1: OK
      Dart Sass version 1.55.0: OK
      Version: 1.3.361
      Path: /opt/quarto/bin

      Version: 3.10.12
      Path: /usr/bin/python3
      Jupyter: (None)

      Version: 4.3.0
      Path: /usr/lib/R
      LibPaths:
        - /home/nick/R/x86_64-pc-linux-gnu-library/4.3
        - /usr/lib/R/site-library
        - /usr/lib/R/library
      knitr: 1.43
      rmarkdown: 2.22

```

mcanouil:
This PR adds h1-6 tags to the CSS rules to apply `monobackgroundcolor` on.

Fixes #6781


src/resources/formats/html/bootstrap/_bootstrap-rules.scss
```diff
@@ -907,6 +907,12 @@ pre {
 // Maps the pandoc 'monobackgroundcolor' to bootstrap
 // Note this only targets code outside of sourceCode blocks
 @if variable-exists(mono-background-color) {
```
cderv, line 909: Looking at this, I am wondering if this part of the code (the oldest from https://github.com/mcanouil/quarto-cli/commit/59591f61a705414c398f0fd96d003132ae193722) is not unnecessary anymore. 
https://github.com/quarto-dev/quarto-cli/blob/e637c76b14e5327a4053ff0ffcaa8cab27b672dd/src/resources/formats/html/bootstrap/_bootstrap-rules.scss#L907-L924
It seems we are duplicating it with what is below 
https://github.com/quarto-dev/quarto-cli/blob/e637c76b14e5327a4053ff0ffcaa8cab27b672dd/src/resources/formats/html/bootstrap/_bootstrap-rules.scss#L926-L949

The later is more conditional to variables than the former, and it will overwrite it too. 

Possibly that was the issue for #4746 fixed by adding a duplicated conditional in 9fff5b7cc4ece31767c91ebf5752d979d6f07999

Just sharing this observation as it could be an opportunity to simplify. We could also leverage more SASS syntax to avoid writing and maintaining the expanded version of all tags and rules with `code:not(.sourceCode)` - but that is ok like this also.

cderv (commented):
Thanks @mcanouil ! 

Just a note about possible simplification of our style file that we could do

src/resources/formats/html/bootstrap/_bootstrap-rules.scss
mcanouil: From my few tests, I did not notice something wrong with this "duplication" and assumed it was meant to fix something elsewhere.

cderv: Yes I think it would be the second. I believe it is duplicate because when adding the second, the first already existed. And then it was overriding the first, which caused #4746 I believe.

mcanouil: Merged the two rules into one (the second): f36fb64f11140c9a32605e476ae7f5c2e2a6113f

mcanouil:
Closing in favour of:
- #6969

PR #6896, opened 2023-09-19
Update R dependencies and use latest renv tooling

cderv:
- Update to renv 1.0.3
- Adapt settings to use PPM by default and R version we have on GHA
- Update to latest version of CRAN packages
- **BIG CHANGE**: Switching to [explicit dependencies discovery](https://rstudio.github.io/renv/reference/dependencies.html?q=dependen#explicit-dependencies) using a `DESCRIPTION`. There is a lot of files to scan otherwise and it slows down start with renv, and update of dependencies. So now a new R deps should be added manually in that file
- I added a `.renvignore` to still speed up dependencies when we run `renv::dependencies()` manually 

Let's see if our CI runs ok with these changes


cderv:
I'll add a dev doc too maybe.

PR #6925, opened 2023-09-20
JS file for reveal-plugin extension needs to be a EJS template

### LINKED ISSUE #6922: Quarto create extension for revealjs plugin does not fill EJS template
This is what I get doing `quarto create` and choosing extension for revealjs

````
<%= `window.Reveal${classname} = function () {
  return {
    id: "Reveal${classname}",
    init: function (deck) {
      // TODO: Implement your plugin functionality
      // Learn more at https://revealjs.com/creating-plugins/
      
      // This example shuffles the deck when the 'T' key is pressed
      deck.addKeyBinding({ keyCode: 84, key: "T" }, () => {
        deck.shuffle();
      });

    },
  };
};` %>
````

So I believe this should be filled but it is not.

cderv:
Fix #6922

I believe this is the only thing missing @dragonstyle right ?


dragonstyle (approved):
yes!

PR #7097, opened 2023-09-30
Bugfix/col margin

dragonstyle:


cscheid (commented):
Hm. This reverts a change I believe I had made for a good reason relatively recently. Let me investigate.

Update: `git log` shows the culprit to be `8ba05ff2ae45c88a1d619190396441a5025e0e2f`, so it must have been one of the last changes I made before the squash-3-months-of-work hell commit 🤦.

So I must have made a bad fix in that commit, and since we have better test coverage in the last 2 weeks, your fix is Good, Actually. Thank you!

Issue #7088, opened 2023-09-29
Default listing contains extraneous empty <a> tag in each item


petrbouchal:
### Bug description

In current pre-release, a default listing produces the code pasted below, which results in a slab of blank space at the top of each listing item.

Right below the <body> tag, there is a <p> tag containing an empty <a> tag. This results in an empty strip at the top of each listing item  (reflecting the padding CSS setting for <p> tags).

This does not happen in the stable version.

```html
<div id="listing-zacit" class="quarto-listing quarto-listing-container-default">
<div class="list quarto-listing-default">
<div class="quarto-post image-right" data-index="0" data-listing-file-modified-sort="1695670046091" data-listing-reading-time-sort="1" data-listing-word-count-sort="1">
<div class="body">
<p><a href="[../pruvodce/kde-zacit.html](http://localhost:4444/pruvodce/kde-zacit.html)"></a></p><a href="[../pruvodce/kde-zacit.html](http://localhost:4444/pruvodce/kde-zacit.html)">
<h3 class="no-anchor listing-title">
Průvodce: kde začít
</h3>
<div class="listing-subtitle">

</div>
</a>
</div>
<div class="metadata">
<p><a href="[../pruvodce/kde-zacit.html](http://localhost:4444/pruvodce/kde-zacit.html)"></a></p><a href="[../pruvodce/kde-zacit.html](http://localhost:4444/pruvodce/kde-zacit.html)">
</a>
</div>
</div>
</div>
<div class="listing-no-matching d-none">
Žádné nalezené výsledky
</div>
</div>
</div>
```

### Steps to reproduce

The listing is specified like so in the page YAML:

```yaml
listing:
  - id: rizeni
    categories: false
    sort: poradi
    type: default
    include: 
      blok: "rizeni"
    fields: 
      - title
```

and inserted using 

```qmd
:::{.pruvodce-box}

::: {#zacit}
:::

:::
```

Only one file matches the `include` filter. On the same page, I have other listings with the same setup that multiple files and the same issues appears in them as well.

The one matches file is:

```qmd
---
title: "Průvodce: kde začít"
blok: ['zacit']
poradi: 1
zacit: true
editor: source
aliases: 
  - /pruvodce/kde-zacit
---
```

### Expected behavior

No extraneous tags with no content should be inserted.

### Actual behavior

[...]

PR #7116, opened 2023-10-02
fix: listing display issue caused by #7088 fix

mcanouil:
9bd1efbd02f8930d7076d391fea98fab0240082f (#7088) introduced a style regression in listing when there is a code div in title and/or description.

Reprex:

```bash
quarto create project blog mysite
cd mysite
sed -i '' 's/Code/`Code`/g' "posts/post-with-code/index.qmd"
quarto render
```

| Before this PR | With this PR |
|--------|--------|
| image | image | 
| image | image |

### DIFF
--- a/src/resources/projects/website/listing/quarto-listing.scss
+++ b/src/resources/projects/website/listing/quarto-listing.scss
@@ -615,13 +615,7 @@ div.quarto-post {
 
   a {
     color: $body-color;
-    display: flex;
-    flex-direction: column;
     text-decoration: none;
-
-    div.description {
-      flex-shrink: 0;
-    }
   }
 
   .metadata {


dragonstyle (approved):
Great catch thank you!

mcanouil:
Hopefully, the change should have no impact elsewhere as the flex display in `<a>` element- has not much impact, but I wonder why it was added in the first place (a while back by the way).

dragonstyle:
Yes agree - I think it makes sense that when the `a` tag was used as the container to all the other elements it might have a role in layout, though in hindsight I think it makes less sense to use it in a container in this way (since it isn't clear what precisely will appear within it). I'm not sure what role it was playing given how the HTML evolved, so I think this is a good improvement...

PR #7202, opened 2023-10-11
Add support for line highlighting for Docusaurus via code-line-numbers 

### LINKED ISSUE #7201: Support line highlighting for Docusaurus via `code-line-numbers`
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/7199

<sup>Originally posted by **eitsupi** October 11, 2023</sup>
### Description

It would be great if the following line highlighting that Quarto supports in revealjs format also worked in Docusaurus.

````md
```{.python code-line-numbers="6-8"}
import numpy as np

r = np.arange(0, 2, 0.01)
theta = 2 * np.pi * r
fig, ax = plt.subplots(subplot_kw={'projection': 'polar'})
ax.plot(theta, r)
ax.grid(True)
plt.show()
```
````

https://quarto.org/docs/presentations/revealjs/index.html#line-highlighting

The syntax on Docusaurus is like this:

````md
```jsx {1,4-6,11}
import React from 'react';

function MyComponent(props) {
  if (props.isBar) {
    return <div>Bar</div>;
  }

  return <div>Foo</div>;
}

export default MyComponent;
```
````

https://docusaurus.io/docs/markdown-features/code-blocks#highlighting-with-metadata-string

The problem here is that `revealjs` is hard-coded in the following part where this conversion is performed, and other formats cannot be added later.
It would be great if third party formats could enable this feature.

https://github.com/quarto-dev/quarto-cli/blob/880565f89b0df66b8ae0f69d30e4959e54464f58/src/resources/filters/quarto-pre/line-numbers.lua#L13-L16</div>

### PR DESCRIPTION
**Pull request by cderv**
close #7201

This follows syntax at https://docusaurus.io/docs/markdown-features/code-blocks#highlighting-with-metadata-string

This PR also improve the handling of `code-line-numbers` when string are passed for revealjs

### CODE REVIEW

File: src/resources/filters/quarto-pre/line-numbers.lua [hunk 1]
```diff
@@ -1,29 +1,35 @@
...
           -- use the pandoc line numbering class
           el.attr.classes:insert("number-lines")
-          -- remove for all formats except reveal
-          if not _quarto.format.isRevealJsOutput() then
-            el.attr.attributes["code-line-numbers"] = nil
+          -- remove for all formats except reveal and docusaurus
+          if type(lineNumbers) == "string" and (_quarto.format.isRevealJsOutput() or _quarto.format.isDocusaurusOutput()) then
+            if _quarto.format.isDocusaurusOutput() and lineNumbers:match("|") then
+              -- docusaurus does not support step separator like revealjs
+              lineNumbers = lineNumbers:gsub("|", ",")
```
eitsupi, line 19: Since the Docusaurus syntax comes from here and is quite different, it may not be necessary to do a conversion from the revealjs syntax.
https://www.npmjs.com/package/parse-numeric-range

src/resources/filters/quarto-pre/line-numbers.lua
```diff
@@ -1,29 +1,35 @@
...
         local lineNumbers = lineNumbersAttribute(el)
+        el.attr.attributes[constants.kCodeLineNumbers] = nil
         if lineNumbers ~= false then
           -- use the pandoc line numbering class
           el.attr.classes:insert("number-lines")
-          -- remove for all formats except reveal
-          if not _quarto.format.isRevealJsOutput() then
-            el.attr.attributes["code-line-numbers"] = nil
+          -- remove for all formats except reveal and docusaurus
+          if type(lineNumbers) == "string" and (_quarto.format.isRevealJsOutput() or _quarto.format.isDocusaurusOutput()) then
```
eitsupi, line 16: It seems to me that this would mean that the new formats added by extension would not be able to use this feature unless the Quarto CLI source code is updated, which would not be ideal, would it?

eitsupi (commented):
Thanks for the very quick update!

src/resources/filters/quarto-pre/line-numbers.lua
cderv: Sorry by I don't understand the question. The docusaurus "extension" is built in now: 
- https://quarto.org/docs/output-formats/docusaurus.html
- https://github.com/quarto-dev/quarto-cli/tree/main/src/resources/extensions/quarto/docusaurus

So any change to it will require a new update of Quarto CLI. 

Also `code-line-numbers` attributes is handled in Quarto source code so this will require a new version of the CLI. 

What do you have in mind ?

eitsupi: Sorry my question is not clear.
For example, it seems that Material for MkDocs has a similar feature, but if someone has created a Quarto extension for Material for MkDocs, they cannot access here.
https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#highlighting-specific-lines

src/resources/filters/quarto-pre/line-numbers.lua
dragonstyle: I'm in favor or removing the step separator and supporting Reveal syntax when possible. In the long run it would be nice to parse the range expression and be able to generate a compatible syntax for the various formats that we're targeting (e.g. have a `quarto syntax` for ranges which will emit reveal, Docusaurus, material, etc...)

dragonstyle (approved):
LGTM

src/resources/filters/quarto-pre/line-numbers.lua
eitsupi: > In the long run it would be nice to parse the range expression and be able to generate a compatible syntax for the various formats that we're targeting (e.g. have a `quarto syntax` for ranges which will emit reveal, Docusaurus, material, etc...)

It's definitely the best! Thanks.

src/resources/filters/quarto-pre/line-numbers.lua
cderv: Ok I understand. 

> if someone has created a Quarto extension for Material for MkDocs (off course outside of quarto-cli)

Is this hypothetical or is there one which needs this ? Just to gauge priority here. 

It seems to me this is a new feature where a format would need to declare it is supporting line highlighting. We choose to support `code-line-numbers` options for line highlight too, but this means that we need to teach quarto which format can support it, as we don't have a mechanism yet for formats to tells quarto. 

@dragonstyle do we have already some mechanism for an extension format (or a format in general) to tell some quarto process that something is supported ? 

In this case we need to know when to pass `code-line-numbers: <string>` as attributes on the CodeCell for the format to process. This is not a Pandoc feature so it requires dedicated processing

* Revealjs handles it in a Plugin
* Docusaurus handles it in the Lua Writer
* a mkDocs extension would need to do the same. 

Anyhow, if we were to support that, this would require a new CLI version anyway. I'll probably merge this and we'll see how to support more formats in the future

cderv: I opened https://github.com/quarto-dev/quarto-cli/issues/7213

eitsupi: @cderv Thank you for summarizing this. I agree that merging this is sufficient for now.

> Is this hypothetical or is there one which needs this ? Just to gauge priority here.

I just wanted to point out that this is an issue with the current implementation.
I personally don't want to use Quarto and Material for MkDocs in combination right away (I currently prefer Docusaurus!).

However, given the popularity of Material for Mkdocs, I'm sure there will be users who need something like that.

cderv: Ok good. I'll take note on that and see if I can add that for 1.4 maybe. Seems easy enough.

PR #7220, opened 2023-10-12
Meca `manifest-version` is "1"

rowanc1:
@dragonstyle I am going through and working with Quarto manifests now. Noticed that the manifest version you are writing is 2.0, I think that it should be "1" according to the [spec](https://groups.niso.org/higherlogic/ws/public/download/23902/NISO_RP-30-2020_Manuscript_Exchange_Common_Approach_MECA.pdf):

image


dragonstyle:
You are correct - I think I was confusing the NISO spec version for the manifest version! TY for the PR and I will merge once the tests complete! 💯

PR #7239, opened 2023-10-16
Handle preamble in layout for hugo-md format

cderv:
follow-up on #7208

This was discovered after fixing our tests infrastructure that was missing test. 

I used the same logic as for other formats regarding preamble handling


cscheid (approved):
Thank you, that looks exactly right.

PR #7362, opened 2023-10-25
Update sidebar toggle icon

cpsievert:
In bslib, we recently updated the sidebar toggle icon from [`arrow-bar-left`](https://icons.getbootstrap.com/icons/arrow-bar-left/) to [`chevron-left`](https://icons.getbootstrap.com/icons/chevron-left/). Here is the PR that makes that change (which also has some before/after screenshots) https://github.com/rstudio/bslib/pull/841.

For consistency (and simplicity), I'd recommend that Quarto makes the same change.


dragonstyle:
Thanks @cpsievert

PR #7417, opened 2023-10-30
Improvment to auto install process when detecting error

### LINKED ISSUE #7252: detect tlmgr issue and refresh installation info
We start seeing in our CI
````
compilation failed- tlmgr needs to be updated, but the update failed
    LaTeX Error: Mismatched LaTeX support files detected.
    (LaTeX)        Loading 'expl3.sty' aborted!
    (LaTeX)        
    (LaTeX)        The L3 programming layer in the LaTeX format
    (LaTeX)        is dated [202](https://github.com/quarto-dev/quarto-cli/actions/runs/6547127605/job/17779023801#step:25:203)3-08-29, but in your TeX tree the files require
    (LaTeX)        at least 2023-10-10.
    
    For immediate help type H <return>.
````

This is a know issue that requires some refreshing. We have seen the issue with tinytex
* https://github.com/rstudio/tinytex/issues/424

We probably need to implement the same detection and commands to run.

### PR DESCRIPTION
**Pull request by cderv**
This aims to fix #7252

This is another way to fix initial report of this problem at #7087 based on what we do in https://github.com/rstudio/tinytex R package

* `Loading 'expl3.sty' aborted!` is considered as any other error, and `expl3.sty` package will be looked for
* When triggerring install for a package, if already install we do run an update first 
* The update included a rebuild of format tree with fmtutils (https://tug.org/texlive/doc/fmtutil.html) - which could be outdated following some `tlmgr` update

Trigerring CI to see if this works ok. The issue from #7252 is hard to reproduce and test, but we haven't seen it in TinyTeX for while, after the same fix https://github.com/rstudio/tinytex/commit/5946a2a1bf3c6ecefbd1213178a4e473ff4c26dc  (https://github.com/rstudio/tinytex/issues/291)

@yihui pinging you in case you want to have a look

### DIFF
--- a/src/command/render/latexmk/texlive.ts
+++ b/src/command/render/latexmk/texlive.ts
@@ -446,7 +446,7 @@ function tlmgrCommand(
   }
 }
 
-// Exectute fmtutil
+// Execute fmtutil
 // https://tug.org/texlive/doc/fmtutil.html
 function fmtutilCommand(context: TexLiveContext) {
   const fmtutil = texLiveCmd("fmtutil-sys", context);
--- a/src/command/render/latexmk/texlive.ts
+++ b/src/command/render/latexmk/texlive.ts
@@ -239,6 +239,12 @@ async function installPackage(
     if (updateResult.code !== 0) {
       return Promise.reject();
     }
+
+    // Rebuild format tree
+    const fmtutilResult = await fmtutilCommand(context);
+    if (fmtutilResult.code !== 0) {
+      return Promise.reject();
+    }
   }
 
   // Run the install command
@@ -268,6 +274,12 @@ async function installPackage(
       quiet,
     );
     if (updateResult.code !== 0) {
+      return Promise.reject();
+    }
+
+    // Rebuild format tree
+    const fmtutilResult = await fmtutilCommand(context);
+    if (fmtutilResult.code !== 0) {
       return Promise.reject();
     }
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/command/render/latexmk/texlive.ts [hunk 1]
```diff
@@ -417,3 +445,16 @@ function tlmgrCommand(
     return execTlmgr([tlmgr.fullPath, tlmgrCmd, ...args]);
   }
 }
+
+// Exectute fmtutil
```
cderv, line 449:
```suggestion
// Execute fmtutil
```

bhattmaulik:
I recently had the same issue. My tinytex installation is from the tinytex R package `(tinytex::install_tinytex(bundle = "TinyTeX-2"))`. To solve it, I first updated the tlmgr `(tinytex::tlmgr_update())`. And finally, I also tried `tinytex:::refresh_all`. But none of them helped. Finally, I uninstalled and reinstalled tinytex, and then it got solved.

cderv:
Thanks for sharing @bhattmaulik. 

> I also tried tinytex:::refresh_all

I am surprised this did not work... Did the issue happened only with Quarto or also with R Markdown ? (as you are mention **tinytex** R package, I am assuming you are a Rmd user too)

bhattmaulik:
(1) I didn't try R Markdown. You are right that I used to be an R Markdown user earlier, but now I have completely moved to quarto. But you have given me a good idea to debug the problem, if it happens next time. (2) TinyTeX-2 is a full bundle. So, I also didn't expect to have this error. I had the issue with expl3.sty, as you have mentioned in the first comment. In fact, my first attempt was to do `tinytex::parse_install`, but it said it couldn't find anything for this command. Only after that I followed the other steps, which I mentioned in my previous comment.

cderv:
Ok thanks for the feedback. This may be indicated the issue is more complex that it seems. 

Though `tinytex::parse_install(file = "expl3.sty")` should have done the right thing. 

Anyhow, let see if this work ok.

PR #7464, opened 2023-11-03
fix(changelog-1.4.md): make contributor mention as actual links

mcanouil:
This PR simply turns the `@USER` mention in the changelog as actual links to the contributors' GitHub profiles.

### DIFF
--- a/news/changelog-1.4.md
+++ b/news/changelog-1.4.md
@@ -1,13 +1,13 @@
 ## Languages
 
-- Add Serbian-Latin translation (author: @n_grubor)
-- Add Slovak translation (author: @tom67)
-- Improve Italian translation of 'proof' (author: @espinielli)
-- Add Greek translation (author: @cultab)
-- Add Norwegian translation (author: @lektorodd)
-- Add Lithuanian translation (author: @GegznaV)
-- Add Traditional Chinese (Taiwan) translation (author: @bobby1030)
-- Update Catalan translation (author: @jmaspons)
+- Add Serbian-Latin translation (author: [@n_grubor](https://github.com/n_grubor))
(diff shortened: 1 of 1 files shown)


cscheid:
Notice that 

1) we have typos on some of those names
2) we know because github auto-links the correct ones:

image

Is this necessary for some use case?

mcanouil:
Well since the @ name are used in the change log and the change log is advertised on Quarto website.
I thought if we want to acknowledge the contributor using their GitHub account, explicit link is needed for them to appear on the website.

(I did not have the auto links on the GitHub Apps 😅)

cderv:
Are we already really thorough in adding Github username to changelog for anyone who opens an issue ? Are we doing it only for PR contributor ? 

I ask because you share a ggplot2 pkgdown example, and the way of working in our R packages usually is to add the username and issue number, also for an issue open and not fixed by the user. There are added in NEWS.md usually using github syntax (no markdown link)

Then I believe **pkgdown** when rendering the NEWS file is catching those github handle and transforming them to HTML link.  
> Issues and contributors will be automatically linked to the corresponding pages on GitHub if the GitHub repo can be discovered from the DESCRIPTION (typically from a URL entry containing github.com)
    _source: https://pkgdown.r-lib.org/reference/build_news.html_

Personally, I would prefer our website rendering to do the same if we want to have links on the website, instead of us adding the link on each new changelog item. (We do it already for issue where it could also be done automatically I guess)

mcanouil:
If you have something creating the links automatically, then yes, it seems appropriate to do the same thing for username.

Regarding the "rule" to add username, I have no idea what it is, but seems to be PR authors.

mcanouil:
Outdated.

PR #7538, opened 2023-11-10
Add annotations for GAP (www.gap-system.org)

ChrisJefferson:
## Description

This add annotation support for the 'GAP' programming language (www.gap-system.org). It is a language designed for discrete maths.

I hope this isn't big enough to need contributor agreement, mainly because I don't current have printer access, but I can probably sort that out if required. I'd also be happy for someone to throw this away and do the same thing if required (GAP uses '#' as comment marker).


dragonstyle (approved):
Thanks for the submission - this is small enough to be ok skipping the contributor agreement- ty!

PR #7557, opened 2023-11-13
[GHA] Directly commit to main the update version.txt file

cderv:
This PR solves the problem of update `version.txt` each time we do a pre-release 

Challenges are: 

* `version.txt` leaves in main - This requires a commit to main **after we trigger a pre-release**
* Release process failure needs to be handled correctly 
	 * Commit and tag done needs to be correctly reverted if needed
	 * `pip install` should work, even if done during the release process window (~15min)

This PR does the following: 

* `version.txt` is commited to main at start of release with a tag creation 
* release publishing happens - it takes roughly 15 minutes
	* if success: a new github release is created
	* if failure: **a revert commit of the version.txt + tag deletion is done** 
	
During those 15 minutes, the repo has a version.txt that points to no Github release really. Tag does exists, but no Github release with assets yet. As a failsafe mode if someone install from main during this time: 
	* setup.py tries to download the version from `version.txt`
	* if error - meaning no existence of release and assets - `setup.py` will look for previous commit in history of `version.txt` calling Git command and look the content - this release must exists and is used. 

## Other things this PR does : 

* No more draft PR mode - checkbox is removed
	* 	Draft PR are not really used in our workflow, and this would not be easy to handle with `version.txt`
	* Unchecking `pre-release` when doing manual run will do the same thing: Build everything - just not publish. Which it seems we really needed

* Pre-release description has been updated to a new message with more links and information about the content than just the last commit message. 

I tested all this in my fork so it is working as expected - unless I missed some specificity of this repo. Though I'll be reactive to fix if problems next time we try to do a pre-release.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -35,14 +35,18 @@ jobs:
           read_version=$(cat version.txt)
           version=$(echo $read_version | grep -o '[[:digit:]]*\.[[:digit:]]*' | head -1)
           build_number=$(($(echo $read_version | grep -o '[[:digit:]]*' | tail -1)+1))
+          version_full=$version.$build_number
           echo "version=$version" >> $GITHUB_OUTPUT
-          echo "version_full=$version.$build_number">> $GITHUB_OUTPUT
+          echo "version_full=$version_full">> $GITHUB_OUTPUT
+          echo "**Full Changelog**: https://github.com/cderv/quarto-cli/compare/v${read_version}...v${version_full}" > "release_note.md"
 
       - name: Upload Artifact
         uses: actions/upload-artifact@v3
         with:
           name: News
-          path: ./news/changelog-${{steps.read-version.outputs.version}}.md
+          path: |
+            ./news/changelog-${{steps.read-version.outputs.version}}.md
+            ./release_note.md
 
       - name: Update version.txt
         run: |
@@ -432,7 +436,7 @@ jobs:
       - name: Rename news
         run: |
           ls -la ./News
-          mv ./News/changelog-${{needs.configure.outputs.version_base}}.md ./News/changelog.md
+          mv ./News/news/changelog-${{needs.configure.outputs.version_base}}.md ./News/news/changelog.md
 
       - name: Generate Checksums
         id: generate_checksum
@@ -486,7 +490,7 @@ jobs:
           tag_name: ${{needs.configure.outputs.tag_name}}
           target_commitish: ${{ needs.configure.outputs.version_commit }}
           name: ${{needs.configure.outputs.tag_name}}
-          generate_release_notes: true
+          body_path: ./News/release_note.md
           draft: false
           prerelease: ${{ inputs.pre-release }}
           files: |
@@ -500,7 +504,7 @@ jobs:
             ./Windows Zip/quarto-${{needs.configure.outputs.version}}-win.zip
             ./Mac Installer/quarto-${{needs.configure.outputs.version}}-macos.pkg
             ./Mac Zip/quarto-${{needs.configure.outputs.version}}-macos.tar.gz
-            ./News/changelog.md
+            ./News/news/changelog.md
             quarto-${{needs.configure.outputs.version}}-checksums.txt
 
   docker-push:
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -14,11 +14,6 @@ on:
         required: false
         type: boolean
         default: true
-      draft:
-        description: "Draft"
-        required: false
-        type: boolean
-        default: false
 
 jobs:
   configure:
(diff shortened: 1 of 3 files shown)


cscheid (approved):
This all looks great. One minor paranoid question.

Is it possible to accidentally run two build workflows in parallel? If so, what happens then?

cderv:
> Is it possible to accidentally run two build workflows in parallel? If so, what happens then?

Releases builds are triggered only on schedule or manually. Scheduled build are dry-run really - so no problem. 
Manual build are our responsibility. It is technically possible to trigger two releases build that should publish, if we don't pay attention if one is already running before triggering a new one. 

In this scenario, if no failure happens, it should all go well - there will just be two releases one right after the other. If a failure happens, then there will be the revert commit and untag. So if sequentially this goes well, probably no issue, but maybe a rare case could happen with conflict state to solve. 

Though what are really the change we do two builds in parallel ? Could this happens ? 

If we want to protect against that, we could leverage [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) to have the second triggered be queued until the first is finish. Probably worth it. 
What do you think ?

cderv:
Thanks for the idea - I added concurrency when we do a publishing so that any jobs will be queued when `publish-release` is checked on Manual release 
````yaml
concurrency:
  # make publishing release concurrent (but others trigger not)
  group: building-releases-${{ inputs.publish-release && 'prerelease' || github.run_id }}
````

File: .github/workflows/create-release.yml [hunk 1]
```diff
@@ -504,9 +527,47 @@ jobs:
             ./Windows Zip/quarto-${{needs.configure.outputs.version}}-win.zip
             ./Mac Installer/quarto-${{needs.configure.outputs.version}}-macos.pkg
             ./Mac Zip/quarto-${{needs.configure.outputs.version}}-macos.tar.gz
-            ./News/changelog.md
+            ./News/news/changelog.md
```
dragonstyle, line 530: There are workflows on quarto-web that use this changelog (attached to the release as `changelog.md`) to generate the release notes shown on the website. It may all be ok, but just want to raise in case this changes the path to the changelog and breaks those downstream actions.

(here https://github.com/quarto-dev/quarto-web/blob/main/.github/workflows/actions/release-info/index.js)

dragonstyle (approved):
One note, otherwise looks good.

If this path needs to change, we can update the action on quarto-web to look in the new path too...

.github/workflows/create-release.yml
cderv: Thanks for pointing that out. 

I don't think it will have some impact because we do upload from a different path, but the release will only have the filename. It seems we download directly from the release asset, so `changelog.md` will correctly be the name. 

Same way as `Windows Zip` or `Mac Zip` and not included in the name. 

Looking at my test release in fork it seems to be ok ! 

I'll what out for a Quarto Web error on this after merging this. 

Thanks again !

cderv:
Thanks for the review! 

I made sure that `version.txt` is up to date in merge now (7e50517) so I can merge this PR and now all the next pre-release build will have an incremental number based on `version.txt`. And tag will follow. No more v1.4.489 jump to v1.4.498.

cderv:
One thing I am just thinking is that I need to handle the `1.5` release creation correctly... 

I'll merge this PR and handles the 1.5 switch in another PR. I probably need a check against `QUARTO_VERSION` env var and increment the `version.txt` accordingly (and not just `1.4.xy(z+1)`)

.github/workflows/create-release.yml
```diff
@@ -14,62 +14,71 @@ on:
...
-          echo "build_number=$QUARTO_BUILD_NUMBER" >> $GITHUB_OUTPUT
+          read_version=$(cat version.txt)
+          version=$(echo $read_version | grep -o '[[:digit:]]*\.[[:digit:]]*' | head -1)
+          build_number=$(($(echo $read_version | grep -o '[[:digit:]]*' | tail -1)+1))
+          version_full=$version.$build_number
+          echo "version=$version" >> $GITHUB_OUTPUT
+          echo "version_full=$version_full">> $GITHUB_OUTPUT
+          echo "**Changelog since last release**: https://github.com/${GITHUB_REPOSITORY}/compare/v${read_version}...v${version_full}" > "release_note.md"
+          echo -e "\n\nFull ${version} changelog up to this version:\n" >> "release_note.md"
+          echo -e "  * View: https://github.com/cderv/quarto-cli/blob/v${version_full}/news/changelog-${version}.md\n" >> "release_note.md"
```
coatless, line 49: Shouldn't this line be: 

```
https://github.com/${GITHUB_REPOSITORY}/blob/v${version_full}/news/changelog-${version}.md\n" >> "release_note.md"
```

cderv: yes it should have been but was missed. I was off for evening time but check back the first release, and already fixed at bb03084 

Thanks for watching out and the feedback !

PR #7570, opened 2023-11-14
[GHA] Auto update: Handle the case of new minor / major version creation

cderv:
This complements previous #7557 by handling the case of a new minor release. 

We are in 1.4 - when doing 1.4 release, we'll build 1.5. In that `./configuration` will be updated for  `QUARTO_VERSION` env var. So we check that; 

* If the value is the same as in `version.txt`, we are still in the same release
* If not, then this is a new minor release
	* 	We reset the build number to `0` so that it will be `1.5.0`
	*  We create new changelog file `changelog-1.5.md` with a default content, which is committed 

If a failure happens in this process, both files will be revert (the whole commit)

I believe this will do it - right @dragonstyle ? 

Also this PR adds support for correctly revert any commit and tag if a cancellation happens.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -77,7 +77,7 @@ jobs:
         if: ${{ inputs.publish-release }}
         id: version_commit
         with:
-          add: "version.txt"
+          add: '["version.txt", "news/changelog-${{steps.read-version.outputs.version}}.md"]'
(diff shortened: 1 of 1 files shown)


dragonstyle:
This looks great and will result in much less annoying build numbering! Great improvement

dragonstyle (approved):

PR #7573, opened 2023-11-14
[gha] Add caching for Typst package to avoid downloading from remote 

cderv:
We are caching on GHA the typst packages downloaded. 

As documented at 
* https://typst.app/blog/2023/package-manager#how-packages-are-accessed
* https://github.com/typst/packages#downloads

the package are first search locally before downloading. 

So by caching and restoring at each run, we insure that no remote access will be done for typst packages. 

I did not put invalidation as it seems cache is quite low size, and also no easy way to invalidate the packages to cache. 

Maybe this `cache-typst` action could be put in `quarto-dev/quarto-actions` instead of here, as it seems it could be useful to others. It could even be a Marketplace action maybe 🤔  there was none doing that (but this is easy to setup though)

### DIFF
--- a/.github/actions/cache-typst/action.yml
+++ b/.github/actions/cache-typst/action.yml
@@ -1,0 +1,32 @@
+name: "Cache Typst package"
+description: "Configures the caching for Typst packages"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Typst Cache path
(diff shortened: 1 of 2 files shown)


cscheid (commented):
👍 !

PR #7641, opened 2023-11-20
within citations, if page is undefined do not make it a string

### LINKED ISSUE #7638: When using `citations: true` and `ieee.csl`, `pages = {undefined}` is being printed
### Bug description

When using `citations: true`, `pages = {undefined}` is being printed, e.g.:

```
@online{krishnamurthy2023,
  author = {Krishnamurthy, Dheepak},
  title = {Publishing Documents with {Quarto}},
  pages = {undefined},
  date = {2023-11-19},
  url = {https://kdheepak.com/blog/publish-documents-with-quarto},
  langid = {en}
}
```

See https://kdheepak.com/blog/publish-documents-with-quarto/ as a live example.

I tried with a new project and using `ieee.csl` seems to cause this issue. Here's the test project for your reference:

[test-citations.zip](https://github.com/quarto-dev/quarto-cli/files/13413716/test-citations.zip)

### Steps to reproduce

Use the test project attached in the previous form input to reproduce the bug.

### Expected behavior

if no pages, then remove that entry.

### Actual behavior

_No response_

### Your environment

- OS: MacOS Ventura 13.5.2

### Quarto check output

```bash
quarto check
Quarto 1.4.489
      Pandoc version 3.1.8: OK
      Dart Sass version 1.55.0: OK
      Deno version 1.33.4: OK
      Version: 1.4.489
      Path: /Users/USERNAME/Applications/quarto/bin

      TinyTeX: (not installed)
      Chromium: (not installed)

      Using: Installation From Path
      Path: /Library/TeX/texbin
      Version: 2023

      Version: 3.10.12 (Conda)
      Path: /Users/USERNAME/miniforge3/bin/python
      Jupyter: 5.4.0
      Kernels: julia-1.9, rust, python3

```

cderv:
This would have the side effect of having the undefined field shows in output

Fixes #7638 following change in 2e7ba02abf92f3ef01d3fb9fdbc8e1734c10f8be

cc @dragonstyle

### DIFF
--- a/src/quarto-core/attribution/document.ts
+++ b/src/quarto-core/attribution/document.ts
@@ -394,7 +394,7 @@ export function documentCSL(
 
   // Process keywords
   const kwString = citationMetadata.keyword;
-  if (kwString && typeof (kwString) === "string") {
+  if (kwString && typeof kwString === "string") {
     extras.keywords = kwString.split(",");
   } else if (inputMetadata.keywords) {
     const kw = inputMetadata.keywords;
@@ -465,7 +465,9 @@ function synthesizeCitationUrl(
 function pages(citationMetadata: Metadata): PageRange {
   let firstPage = citationMetadata[kFirstPage];
   let lastPage = citationMetadata[kLastPage];
-  let pages = `${citationMetadata[kPage] as string}`; // Force pages to string in case user writes `page: 7`
+  let pages = citationMetadata[kPage]
+    ? `${citationMetadata[kPage] as string}` // Force pages to string in case user writes `page: 7`
+    : undefined;
   if (pages && pages.includes("-")) {
     const pagesSplit = pages.split("-");
     if (!firstPage) {


kdheepak:
Thanks for the quick PR!

The function you modified returns this:

```typescript
  return {
    firstPage: firstPage as string,
    lastPage: lastPage as string,
    page: pages,
  };
```

but `firstPage` and `lastPage` may also be `undefined`, right? Wouldn't you need to do a similar check for those two lines?

Alternatively, do it in the first couple of lines and return this instead:

```typescript
  return {
    firstPage,
    lastPage,
    page: pages,
  };
```

cderv:
Thanks for chiming in. 

I don't think we need it because 
````ts
> undefined as string
undefined
> `${ undefined}` as string
"undefined"
````

For pages we added in https://github.com/quarto-dev/quarto-cli/commit/2e7ba02abf92f3ef01d3fb9fdbc8e1734c10f8be the special treatment of passing to a string using ``` `${ var }` ``` syntax. And this was the syntax that make `undefined` become a string as `"undefined"` instead of staying `undefined`

dragonstyle (approved):
sweet

PR #7675, opened 2023-11-22
Use %ProgramData% to install TinyTeX when %APPDATA% is not suitable for TeX Live

### LINKED ISSUE #7620: Adapt TinyTeX installation to latest change regarding installation path with space / non-ascii character
Related to 
* https://github.com/rstudio/tinytex/issues/420

and see https://yihui.org/en/2023/11/tinytex-path/

This is a limitation of TeX Live for now, so when this happens TinyTeX install script will now select `$env:ProgramData` as installation place. 

From change by @yihui, this is what happens in **tinytex** R package, and/or TinyTeX install script

- Users will see a warning message when reinstalling or using TinyTeX that was previously installed to `APPDATA` (https://github.com/rstudio/tinytex/pull/429). The message will tell users how to move TinyTeX to `ProgramData`, i.e.,
  ```r
  tinytex::copy_tinytex(to = Sys.getenv('ProgramData'), move = TRUE)
  ```
- For users who install TinyTeX for the first time, the installation path will be `ProgramData`.
- For users who specify a custom installation path via `tinytex::install_tinytex(dir = ...)`, no checking on the path is done currently (but perhaps should be).

If quarto uses the installation script `tools/install-bin-windows.bat` in this repo, there isn't anything to do, otherwise applying the same logic is probably needed.

yihui:
Fixes #7620.

When `%APPDATA%` contains spaces, TeX Live may not work (https://github.com/rstudio/tinytex/issues/420). When it contains non-ASCII characters, TeX Live will not work for sure.

More info: https://yihui.org/en/2023/11/tinytex-path/

### DIFF
--- a/src/tools/impl/tinytex-info.ts
+++ b/src/tools/impl/tinytex-info.ts
@@ -23,7 +23,12 @@ export function hasTinyTex(): boolean {
 export function tinyTexInstallDir(): string | undefined {
   switch (Deno.build.os) {
     case "windows":
-      return expandPath(join(getenv("APPDATA", undefined), "TinyTeX"));
+      let appDir = getenv("APPDATA", undefined);
+      // use ProgramData if APPDATA is not set or contains spaces or non-ASCII characters
+      if (!appDir || !appDir.match(/^[!-~]+$/)) {
+        appDir = getenv("ProgramData", undefined);
+      }
+      return expandPath(join(appDir, "TinyTeX"));
     case "linux":
       return expandPath("~/.TinyTeX");
     case "darwin":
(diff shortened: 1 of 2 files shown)


src/tools/impl/tinytex-info.ts
```diff
@@ -23,7 +23,12 @@ export function hasTinyTex(): boolean {
 export function tinyTexInstallDir(): string | undefined {
   switch (Deno.build.os) {
     case "windows":
-      return expandPath(join(getenv("APPDATA", undefined), "TinyTeX"));
+      let appDir = getenv("APPDATA", undefined);
+      // use ProgramData if APPDATA is not set or contains spaces or non-ASCII characters
+      if (!appDir || !appDir.match(/^[!-~]+$/)) {
```
yihui, line 28: Strictly speaking, the range `!-~` doesn't include the full ASCII table. It excludes spaces (the space is right before `!` in the table: https://en.wikipedia.org/wiki/ASCII#Character_set) and other characters that are very unlikely to be present in a file path.

cderv (approved):
Thanks - this could maybe solve some issue we have reported. (I'll find them to reference)

@dragonstyle it seems simple enough, but is there anything else that could be impacted that I am missing (where we could expect APPDATA installation and not elsewhere ?)

Thanks

dragonstyle (approved):
This looks great, I don't think we should run into other problems as everything should be running through that install dir function....

PR #7696, opened 2023-11-24
Bring support for rendering from R file

### LINKED ISSUE #6660: Render a document from a script (.R, .py)
This idea cames from several discussion - especially on the R side as we have `knitr::spin()` and `rmarkdown::render()` working with R script

* https://github.com/quarto-dev/quarto-cli/discussions/4969
* https://github.com/yihui/knitr/issues/2284
* https://github.com/quarto-dev/quarto-cli/issues/588

`quarto render` could consider rendering to output format a .R / .py script that represents a notebook. 

* Existing syntax for R script to represent Notebook content (text, code, chunk option, ...): https://bookdown.org/yihui/rmarkdown-cookbook/spin.html
* Existing syntax for Python script to represent Notebook content: https://jupytext.readthedocs.io/en/latest/formats-scripts.html

## Notes

These are from internal discussion on this already

### General 

* This kind of feature would be opt-in to avoid rendering a script that shouldn't be. Probably by using a specific comment in the script that quarto could check
  * Example for .py : `# %% [markdown]` if we see this, then we auto render 
    ````
	# %% [markdown]
	# ---
	# title: "My document"
	# ---
	````

  * Other idea: Requiring a title field in YAML header
	  * For R 
	     ```
		#' ---
		#' title: "My document"
		#' ---
		```
	  * For Python
	     ```
		# %% [markdown]
		# ---
		# title: "My document"
		# ---
		```
	  
### For Python 

* The [percent format](https://jupytext.readthedocs.io/en/latest/formats-scripts.html#the-percent-format) seems a good idea considering it is widely used

This issue is to track the idea and have a place for further notes.

cderv:
This closes #6660 

It does the same logic as for **Jupyter** in https://github.com/quarto-dev/quarto-cli/pull/6700

Though proposal here is to use `knitr::spin()` directly, which has support for `.qmd` since v1.44. 

There is question about performance, but we could easily improve by implementing the `knitr::spin()` logic internally in TypeScript. I'll suggest we wait for feedback on usage with calling R directly. 

It works ok  though  little thing I noticed : It seems we run the `asMappedString()` from a input file at two places

Inside `target()` in our call
https://github.com/quarto-dev/quarto-cli/blob/1a8c1aff841651f0d01f0ab150de1517f80f7a0f/src/execute/engine.ts#L214
where we run 
https://github.com/quarto-dev/quarto-cli/blob/74391fc1099e8c167c692370636df1d758406194/src/execute/rmd.ts#L84

and then a second time when rendering project as we do run 
https://github.com/quarto-dev/quarto-cli/blob/1a8c1aff841651f0d01f0ab150de1517f80f7a0f/src/project/project-config.ts#L97

to find resources and this calls
https://github.com/quarto-dev/quarto-cli/blob/74391fc1099e8c167c692370636df1d758406194/src/project/project-config.ts#L110

which needs to get markdown from input file too. 

I am not sure why we need to run all this twice instead of using a stored version from the first run. But I am probably missing something. 

Should we try caching the results of first spin ? Or that could cause issue ? 

On other topic, I'll add a documentation file for the website. Let's note that `#+` is not specifically needed and YAML syntax can directly be used this way. 
````
#| echo: false
#| fig-with: 7
#  This is a normal R comment.
plot(cars)
````

This makes it quite coherent with percent script syntax I believe. I'll probably document only this way on Quarto Web, but the `#+` still works. 

Thanks

### CODE REVIEW

File: src/execute/rmd.ts [hunk 1]
```diff
@@ -353,3 +371,35 @@ function resolveInlineExecute(code: string) {
     (expr) => `${"`"}r .QuartoInlineRender(${expr})${"`"}`,
   )(code);
 }
+
+export function isKnitrSpinScript(file: string) {
+  const ext = extname(file).toLowerCase();
+  if (ext == ".r") {
+    const text = Deno.readTextFileSync(file);
+    // Consider a .R script that can be spinned if first line starts with a special `#'` comment
+    return /^\s*#'/.test(text);
```
jjallaire, line 380: Is there something more we could look for? (as I think this would trigger for R code that uses any type of roxygen comment e.g. a package source file or plumber file)

cderv: Oh indeed. I did not thought package files would be found while rendering with quarto but probably better to be safe here. 

There is no specific spin syntax except `#+` for chunk options but I believe we prefer to use yaml. 

Probably best is requiring a yaml header in the file. First like would need to be `#´ ---`. This would require a user to add one. 

Otherwise we could require a specific fields for quarto. `# /* ` is a special comment that spin will ignore and won't affect output. 
So something like 
```
# /* quarto */
```
but this is quite new and different syntax. 

Maybe requiring yaml header is the simplest here and not too much to ask.

jjallaire:
Yes, I think the YAML header is the thing to do

cderv:
Ok. I agree too. That is better. 

About the two passes we do for the markdown content import which leads to two runs of spin(), do you have insights ? 

I am thinking of caching first spin results in temp folder but not sure this is robust. It feels unneeded but I may be missing something. 
I did not check if we do two percent script import too in this case.

jjallaire:
I wouldn't try to untangle why that is called twice or do any caching. I actually tried to do this already for Jupyter and I broke something I didn't understand related to includes. I think the better solution is to just write a typescript version that is cheaper (don't need to do this for this PR but perhaps a follow-up PR?)

kylebutts:
So exciting y'all. Made a little demo video here: https://x.com/kylefbutts/status/1729145164881674539?s=20

jjallaire:
@cderv We should also make sure that RStudio knows to make "Preview" available for R scripts.

PR #7842, opened 2023-12-08
Adapt the footer text color when slide has dark-background

### LINKED ISSUE #4156: Footer text colour doesn't change with darker background colours
### Bug description

When using revealjs mode with Quarto if you set a slide to a dark background and include a footer the footer colour does not adjust to the dark background.

The footer appears to be moved outside of the slide section where revealjs sets `has-dark-background` class so it isn't possible to just apply scss logic within `src/resources/formats/revealjs/quarto.scss`.

## Steps to reproduce

Using the [attached file](https://github.com/quarto-dev/quarto-cli/files/10544325/footer-colour.txt) do the following:

```bash
$ mv footer-colour.txt footer-colour.qmd

$ quarto preview footer-colour.qmd
```

## OS Information

I've tested this on the most recent nightly release and still have this issue.

Running on macOS Ventura 13.1.

### Checklist

cderv:
closes #4156 

This PR is a proposal so that colors of footer and slide numbers are adapted when on light / dark slides when dark / light background is used. 

Check this PR by checking out the branch and then 

## Dark theme, light slide 

````markdown
---
title: Footers
format: 
  revealjs: 
    incremental: true
    slide-number: true
    theme: dark
---

# Custom footer {background-color="lightblue"}

Something on a red background

::: footer
Amazing custom footer thing [QUarto](https://quarto.org)
:::

````

image

## Dark slide on light theme
 
````markdown
---
title: Footers
format: 
  revealjs: 
    incremental: true
    slide-number: true
    theme: dark
---

# Custom footer {background-color="lightblue"}

Something on a red background

::: footer
Amazing custom footer thing [QUarto](https://quarto.org)
:::

````

image

Only limitation - the tweak happens when the slide changes, so when going to slide directly using a direct url, the color may not be adjusted yet before any slide is changed. 

@dragonstyle I know you looked into that. What do you think about this solution ?

### DIFF
--- a/news/changelog-1.4.md
+++ b/news/changelog-1.4.md
@@ -70,7 +70,7 @@
 - ([#7366](https://github.com/quarto-dev/quarto-cli/issues/7366)): `smaller: true` now applies correctly on nested slides.
 - ([#7394](https://github.com/quarto-dev/quarto-cli/issues/7394)): Fix issue with mermaid diagrams in revealjs slides when `output-location: fragment`.
 - ([#4988](https://github.com/quarto-dev/quarto-cli/issues/4988)): targets for links on numbered code lines are removed, as revealjs doesn't support them because navigation is done by slide only.
-- ([#4156](https://github.com/quarto-dev/quarto-cli/issues/4156)): footer text on slide with dark background have now an adapted text muted color based on `$dark-bg-text-color`.
+- ([#4156](https://github.com/quarto-dev/quarto-cli/issues/4156)): footer and slide number text on slide with dark background have now an adapted text muted color based on `$dark-bg-text-color`.
 
 ## PDF Format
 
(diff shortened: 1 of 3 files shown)


dragonstyle (commented):
This looks good to me- the only question I have is whether `has-dark-background` and `has-light-background` should be exclusive of one another. It seems like in the current implementation, both classes could be applied and in that case, the latter rule would just win since they are conflicting... Not the end of the world, but just seeking clarification...

cderv:
I think they can't be applied at the same time by revealjs, and this code assumes that. Though it does not enforce it, just assume that if one if FALSE the other is TRUE. 

https://github.com/hakimel/reveal.js/blob/993b8f302a4dd9ed2b823ede71a42380eb244bb6/js/controllers/backgrounds.js#L226-L240

Would be better to enforce the independance in our code N

dragonstyle:
| Would be better to enforce the independance in our code ?

I personally think so (it just felt odd to me that both could be true at the same time since they appear to be exclusive). The only other question is whether neither could be true - in that case, then I think you have it just right.

cderv:
> The only other question is whether neither could be true

````
		// If this slide has a background color, we add a class that
		// signals if it is light or dark. If the slide has no background
		// color, no class will be added
````

I believe this is applied only when there is a background color applied on the slide; so in some presentation, there may not be any class like this on section.

PR #7938, opened 2023-12-17
Enable Manuscript reference page

cwickham:
@dragonstyle this change gets the reference page infrastructure in quarto-web to work for manuscript projects, but I don't understand if it has larger consequences.

See https://github.com/quarto-dev/quarto-web/pull/952 for PR that adds reference page to quarto-web.


cwickham:
Closed in favor of edits to `quarto-web/tools/reference.ts`  https://github.com/quarto-dev/quarto-web/pull/952

PR #8007, opened 2023-12-21
Resolve user paths before mergin template partials

dragonstyle:
Since we are merging absolute paths to format resources into the partials provided by the user, we need to resolve the user paths into absolute paths before we merge them. This will allow us to to use project semantics for resolving the path - (e.g. `/` prefix path).


dragonstyle:
Opening to watch tests

PR #8018, opened 2023-12-22
Attributes are passed through in Title when using empty heading

cderv:
Related to #8012 report (solves part of it)

We should consider empty headers with attributes when matching against Pandoc Title regex

Dummy example 

````markdown
---
format: html           
---

## {.empty-header}

Content

## Python slide

```{python}
1 + 1
```
````

image

It context of revealjs, this is a problem has empty headers can have a meaning (creating a new section without title) 

````markdown
---
format: revealjs           
---

## {.empty-header}

Content

## Python slide

```{python}
1 + 1
```
````

image

### DIFF
--- a/src/core/pandoc/pandoc-partition.ts
+++ b/src/core/pandoc/pandoc-partition.ts
@@ -18,7 +18,7 @@ export function firstHeadingFromMarkdown(markdown: string): string | undefined {
   return partitioned.headingText;
 }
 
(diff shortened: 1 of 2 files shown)


dragonstyle (approved):
This looks good to me. I confirmed that Pandoc does parse `## {.foo}` as a heading with attributes/classes as you'd expect so we appear to be handling this incorrectly in its current form.

PR #8191, opened 2024-01-09
Serve - Restore previous output of serve command for RStudio Shiny Docs

### LINKED ISSUE #8186: Change in quarto serve log output is causing IDE to not open the preview
Related to 
* https://github.com/rstudio/rstudio/issues/14039#issuecomment-1882864335

It seems between 1.4 and 1.3 , `quarto serve` has changed what is logged from 

* In 1.4 we now log `Browse at http://localhost:7328/`
* In 1.3 we logged `Listening on http://127.0.0.1:7822`

RStudio IDE is supposed to react when `Listening on ` is in the output for shiny documents, and it seems we are also suppose to react on it
https://github.com/quarto-dev/quarto-cli/blob/9f6968358fb00034c08cc613294c906a184a853a/src/execute/rmd.ts#L215-L225

I believe we are emitting the `Browse at` now  - did we emit the 'Listening on' at the time ? I can't find this in history of changes... 

Can't find what has changed yet... is this in `quarto-dev/quarto` maybe ?

dragonstyle:
Fixes #8186

### DIFF
--- a/src/command/serve/serve.ts
+++ b/src/command/serve/serve.ts
@@ -17,9 +17,10 @@ import {
   printBrowsePreviewMessage,
   resolveHostAndPort,
 } from "../../core/previewurl.ts";
-import { isServerSession } from "../../core/platform.ts";
+import { isRStudio, isServerSession } from "../../core/platform.ts";
 import { openUrl } from "../../core/shell.ts";
 import { notebookContext } from "../../render/notebook/notebook-context.ts";
+import { info } from "log/mod.ts";
 
 export async function renderForServe(
   file: string,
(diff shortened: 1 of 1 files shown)


cderv:
@dragonstyle FWIW RStudio IDE also can fix this: https://github.com/rstudio/rstudio/commit/fb1838bc0593abadf4ccf1d55a8ad31e600c2391

Doing in both places insure backward compatibility for newer quarto in older IDE, other next IDE release will fix it. I'll test next daily when out tomorrow

cderv:
So I tested latest Daily 2024.04.0-daily+200 and it is now working ok as they watch for "Browse at" now. 

I think we are all good and don't need this change. Unless make new quarto work on older Rstudio is best. We do make an exception in message though in this case

dragonstyle:
I think we should take this change to preserve compatibility with previous versions of RStudio (the fix is very narrow to RStudio + serve, so feels very safe to me)

cderv:
Yes safe indeed ! We can definitely take it.

PR #8420, opened 2024-01-24
Add missing --no-prompt check in quarto use binder

cderv:
This really 1.4 missing feature for new `quarto use binder` https://quarto.org/docs/projects/binder.html

but maybe too late to merge. I let you decide. It seems small enough though...

### DIFF
--- a/src/command/use/commands/binder/binder.ts
+++ b/src/command/use/commands/binder/binder.ts
@@ -152,7 +152,7 @@ export const useBinderCommand = new Command()
         info(
           "\nNo files which provide dependencies were discovered. If you continue, no dependencies will be restored when running this project with Binder.\n\nLearn more at:\nhttps://www.quarto.org/docs/prerelease/1.4/binder.html#dependencies\n",
         );
-        const proceed = await Confirm.prompt({
+        const proceed = !options.prompt || await Confirm.prompt({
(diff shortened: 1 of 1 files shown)


dragonstyle (approved):

cderv:
Thanks ! 

Do we backport this to 1.4 or not ? 

Just to know if I include it into the R package and on what condition

dragonstyle:
I think I would personally be ok back porting just because the change is so small / safe.

I do think that in general, configuring Binder is sufficiently complex that doing so non-interactively does seem like more of an edge case (until we add this support in r package, that is!)

cderv:
> I do think that in general, configuring Binder is sufficiently complex that doing so non-interactively does seem like more of an edge case (until we add this support in r package, that is!)

I agree with that. I was hesitant to add it in the R package, but it felt easy enough, until I hit this blocker. 
Situation is : 

- Either it will be in 1.5 and I need to conditionally stop execution if quarto version used with R is < 1.5
- Either it is in 1.4 patch, and I need to check the version too, with minor version 
- Or we don't add it at all in the R package - that would be ok too. 

Only reason I am asking. I am fine with 1.5 too - easy enough to add it like and see if some R users are asking / using it.

PR #8561, opened 2024-02-02
Website Nav - use text for aria-label

### LINKED ISSUE #8555: after body page navigation is wrong on some website
We can see this on several website : 
image

I believe they are all deployed on Netlify (to be confirmed) 

It is possible this issue comes from problem we have in our after body template. 

First we are missing a quote here: 
https://github.com/quarto-dev/quarto-cli/blob/c2d85c54f2e5a6d3f5f0d6d06f328654b9781f26/src/resources/projects/website/templates/nav-after-body-postamble.ejs#L7

Easy fix  (dd89c2f - possibly this was the issue in https://github.com/quarto-dev/quarto-cli/issues/7319 initially)

other problem :  both in line above and in this one 
https://github.com/quarto-dev/quarto-cli/blob/c2d85c54f2e5a6d3f5f0d6d06f328654b9781f26/src/resources/projects/website/templates/nav-after-body-postamble.ejs#L16

we are inserting in `aria-label` more than text. 

Example from: https://feaz-book.com/introduction

This is the HTML source on github in the repo https://github.com/EmilHvitfeldt/feature-engineering-az/blob/5f5ef5b4c48f502b28e565773f6dcec0f0f39c5c/_book/introduction.html#L1740-L1744
````html
  <div class="nav-page nav-page-next">
      <a href="./numeric.html" class="pagination-link" aria-label="<span class='chapter-number'>1</span>&nbsp; <span class='chapter-title'>Numeric Overview</span>">
        <span class="nav-page-text"><span class="chapter-number">1</span>&nbsp; <span class="chapter-title">Numeric Overview</span></span> <i class="bi bi-arrow-right-short"></i>
      </a>
  </div>
````

Note that we have a rather complex `aria-label` we should not have IMO, which mix `'` and `"`. 

This is what we can see in the source page on the deployed website (https://feaz-book.com/introduction with view-source mode in chrome)
```html
  <div class="nav-page nav-page-next">
      <a aria-label='<span class=\'chapter-number\'>1</span>&nbsp; <span class=\'chapter-title\'>Numeric Overview</span>' class='pagination-link' href='/numeric'>
        <span class="nav-page-text"><span class="chapter-number">1</span>&nbsp; <span class="chapter-title">Numeric Overview</span></span> <i class="bi bi-arrow-right-short"></i>
      </a>
  </div>
```

So quotes are changed, and mess up a bit, which lead to a bad DOM and the problem of the screenshot. 

So something to look into in the sidebar content that gets inner HTML representation with `<span>` and not just the text without HTML structure.

dragonstyle:
Books were injecting HTML tags into the nav item text, which would cause HTML tags to be embedded within the aria-label. Instead, use a plain text representation.

Fixes #8555

### DIFF
--- a/src/project/types/book/book.ts
+++ b/src/project/types/book/book.ts
@@ -180,11 +180,12 @@ export const bookProjectType: ProjectType = {
   ) => {
     const chapterInfo = chapterInfoForInput(context, input);
     if (chapterInfo && number) {
-      return Promise.resolve(
-        numberChapterHtmlNav(text, chapterInfo),
-      );
+      return Promise.resolve({
+        text,
+        html: numberChapterHtmlNav(text, chapterInfo),
+      });
     } else {
-      return Promise.resolve(text);
+      return Promise.resolve({ html: text, text: text });
     }
   },
 
--- a/src/project/types/types.ts
+++ b/src/project/types/types.ts
@@ -66,7 +66,7 @@ export interface ProjectType {
     input: string,
     text: string,
     number: boolean,
-  ) => Promise<string>;
+  ) => Promise<{ html: string, text: string}>;
   incrementalRenderAll?: (
     context: ProjectContext,
     options: RenderOptions,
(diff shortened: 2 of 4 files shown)


dragonstyle:
When back porting, be sure to also grab: dd89c2f1a50e280f9cc50e6ff957847ae8a7fabe

PR #8645, opened 2024-02-08
Native Julia engine based on QuartoNotebookRunner.jl

jkrumbiegel:
## Description

This PR implements a native Julia execution engine as an alternative to the Jupyter workflow. The benefit is that only quarto and Julia have to be installed, without any Python intermediary. That makes it easier for Julia users to install and maintain a working quarto system.

We are executing qmd files via [QuartoNotebookRunner.jl](https://github.com/PumasAI/QuartoNotebookRunner.jl/), which spins up a control server to communicate with via tcp. The control server itself keeps a number of worker processes which can be used for keepalive scenarios, to keep latency low.

Another benefit vs. the Jupyter workflow is that our control server runs out of a private Julia environment that quarto sets up the first time a Julia render is initiated. This means users do not have to install anything in their own Julia environments to run quarto, which is consistent with best practices not to crowd global environments.

In our internal testing, this system already works quite well, so we are interested in getting maintainer feedback on the implementation, as well as potential paths forward.

The engine currently only executes Julia code blocks, it is conceivable, though, to later add support for specialized R and Python compatibility via RCall.jl and PythonCall.jl or similar packages on top. This is not in scope of this PR, however.

## Checklist

I have (if applicable):


cscheid:
@jkrumbiegel Just checking in to see if you folks are looking for comments for us here or are still working on the PR.

jkrumbiegel:
@cscheid yes we're looking for comments, to see what your thoughts about the current implementation are, how it fits into the existing framework. I've made the PR a draft because there are no tests yet. I wanted to start adding such things after the general approach has been greenlit.

cscheid:
@jkrumbiegel Ah, sorry for the delay, in that case. I hope to have comments for you by the end of the week.

jkrumbiegel:
Thank you!

src/execute/engine.ts
```diff
@@ -29,9 +29,11 @@ import { mergeConfigs } from "../core/config.ts";
...
 import { pandocBuiltInFormats } from "../core/pandoc/pandoc-formats.ts";
 import { gitignoreEntries } from "../project/project-gitignore.ts";
+import { juliaEngine } from "./julia.ts";
 
 const kEngines: ExecutionEngine[] = [
   knitrEngine,
+  juliaEngine,
   jupyterEngine,
   markdownEngine,
 ];
```
cscheid, line 39: One thing we'd like to do in the future is add the ability for third-party engines to register themselves as an extension, so that you don't have to ask us for a PR every time you want a new engine or change.

When you merge against `main`, you'll have to fix a small conflict here and it'll be clear what it's about. (There currently isn't the ability to actually register the engine externally, but I hope we'll do that in 2024 still.)

src/execute/julia.ts
```diff
@@ -0,0 +1,455 @@
...
+  claimsLanguage: (language: string) => {
+    return language.toLowerCase() === "julia";
+  },
+
+  partitionedMarkdown: async (file: string) => {
+    return partitionMarkdown(Deno.readTextFileSync(file));
+  },
+
+  // TODO: ask dragonstyle what to do here
+  executeTargetSkipped: () => false,
```
cscheid, line 76: In some code paths, Quarto needs to create an engine and `ExecutionTarget` even if we don't intend to eventually execute the markdown.

In those cases, we call `executeTargetSkipped()` to signal that the specific target was skipped. In the Jupyter engine, this causes a cleanup of the notebook. If you don't create transient files prior to the execution, then I think this is fine.

cscheid:
Hi there, sorry for the delay. We would definitely like to merge this. In the small, everything in the PR looks good. There's going to be one conflict you'll need to solve in the engine declaration where I recently made a small change to how the global list of engines is managed.

You write this:

> Another benefit vs. the Jupyter workflow is that our control server runs out of a private Julia environment that quarto sets up the first time a Julia render is initiated. This means users do not have to install anything in their own Julia environments to run quarto, which is consistent with best practices not to crowd global environments.

I really like this, and I think it's pretty consistent with how we'd like Quarto to operate. However, I have a few questions. How do you manage required packages in this environment? How and where are they declared/installed? How are versions declared/maintained?

Currently, Quarto defers to the user to maintain those environments. In a vacuum, I like the idea of Quarto taking over the maintenance of the environment, but I'm not sure all users will like that they will end up with duplicated packages in their system. They probably will run into surprises if upgrading their own environment doesn't mean that the environment we set up will not be upgraded as well.

Finally, I see a couple of failures in the Julia notebook when using multiline YAML arrays. We'd like to fix that before merging.

Please let me know if you need any more help here, but I'm quite excited. This looks like a really nice solution for pure-Julia users.

cscheid:
> Could you add an example here that fails and we'll get it sorted?

Sorry, it's our test suite, here: https://github.com/quarto-dev/quarto-cli/actions/runs/7829788399/job/21362500154?pr=8645

MichaelHatherly:
YAML parsing for that case is fixed in the most recent version, so @jkrumbiegel upgrading that should help. There's a different issue related to `display` usage that we'll need to fix in QNR, I will sort that next week.

cscheid:
> So the server depends just on our package, and we fix the patch version, so quarto can decide to update that version whenever they want, and can be sure it's always compatible. One could depend on the minor but currently it was simpler and safer to fix the patch.

Sorry, I think I phrased my question poorly. My question was about what happens with dependencies of the code in Quarto documents, not Quarto itself. 

Currently, Quarto takes a hands-off approach to those, so if a Quarto project (or file) needs specific libraries installed, then those libraries are the responsibility of the user. You've described the changes here as "[the] control server runs out of a private Julia environment", which I originally took to mean that package management now has become the responsibility of the engine somehow. But now I think this was a bad interpretation on my part, so I just remain confused.

We need that the base Quarto program to not a required dependency on Julia, because of the consequences this would have on our installers. In the long term, we are going to drop the baseline Python dependency as well and implement the Jupyter protocol ourselves. For example, users do not need R to be installed on the system in order to run Quarto. (I'm not claiming that your PR fails here, I just haven't verified that your PR satisfies this requirement)

cscheid:
> I have a question, too, could you give some guidance how you want tests to be set up?

We use Deno's test infrastructure, and drive it with `tests/run-tests.sh`, which run on CI (and you can see the results above). Roughly speaking, we have two kinds of tests:

1) any file that passes the `test/**/*.test.ts` glob is run through `deno test`. See [tests/smoke/book/render-book.test.ts](https://github.com/quarto-dev/quarto-cli/blob/main/tests/smoke/book/render-book.test.ts), for example.

2) in addition, our test suite attempts to render Quarto files and projects inside the directory `tests/docs/smoke-all`. This gives you a way to quickly check for unexpected crashes, but also allows you to make format-specific assertions. The easiest way to find these is to search for the `_quarto:` string in `.qmd` files inside `tests/docs/smoke-all`. We don't have documentation on the assertion available, but if y'all made it through the typescript part of our code base, this is going to be comparatively easy :)

jkrumbiegel:
I don't think Julia is in any way required for quarto the way it's currently set up, however I think the julia engine had some key where it "claims" a qmd if it has julia code blocks. I'm not sure how that all works exactly, but we'd want to make sure that the current jupyter workflow is not impacted if users don't want the native backend (yet).

cscheid:
> I'm not sure how that all works exactly

I think the julia engine had some setting where it "claims" a qmd if it has julia code blocks. If there's no explicitly specified engine, then Quarto [asks each of the registered engines in order](https://github.com/quarto-dev/quarto-cli/blob/main/src/execute/engine.ts#L95-L140) whether they're responsible for a given language.

This code is assuming 

> but we'd want to make sure that the current jupyter workflow is not impacted if users don't want the native backend 

This is a good point, and I'd like to have some simple tests that verify the previous configurations aren't affected. Right now, the simplest way to do it is have the native julia engine never claim the julia language cells. This would negatively affect your experience a bit since it would require explicitly setting `engine: julia` in all your documents. But I don't think we want the previous engine autodiscovery experience for `julia` cells to change...

> (yet)

yet :)

jkrumbiegel:
@cderv or maybe you could help me out how I could see logging information during the CI test runs? They seem to be swallowed unlike on my local machine, so I cannot do much to find out what's going on

cderv:
The `ConnectionRefused` does not ring a bell at all. I will try to see the CI log to understand. 

We can try to re-run workflow in debug-mode (https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging), but I don't know what we will see exactly, it is more debug for the action steps usually. 

You will not have the right to do this in this repo probably (re-run with logging, stop some runs...) 

Usually if I need to to some "admin" stuff on repo for a PR, I will use my fork to run the CI as there I can have all the right I want. This could require some adjustment in workflow on the fork to avoid connection to some unavailable resources from the fork, but I don't think we have some in the parallel testing workflow. 

So this could also be an option to run CI from this branch in your fork directly (Update your fork, activate actions, create a PR in your updated fork - this should trigger CI there). 

This is only advice - I'll do my best to be reactive as soon as you ping me here. For now, I'll look at what I can see in the log and try re-run in debug mode

cderv:
Also we can probably tweak the PR on the workflow files so that this PR is only tested on your new test files for some times and only install what is required. Just thinking of that to improve the speed of test while this is debugged. 

I can help you set up this if you want.

jkrumbiegel:
Good idea with the reduced CI on the fork, I could try that. You can see the `lsof` command from the last commit which I tried to run before the server connects. On my local machine, when I run `run-tests.sh` I see the printout of that, but somehow not on CI. Are there different logging settings for CI? This is usually how I debug CI, by printing as much contextual information as I need to understand the bug.

cderv:
> I triggered rerun of the two linux failed job with debug logging

This did not bring much information 

> Good idea with the reduced CI on the fork, I could try that.

I know well our CI test logic so, do ask if you need me to do it. Easy way would be to trigger the test only on the .qmd test file you added, and remove Python and R installation. 

> You can see the lsof command from the last commit which I tried to run before the server connects. On my local machine, when I run run-tests.sh I see the printout of that, but somehow not on CI. Are there different logging settings for CI? This is usually how I debug CI, by printing as much contextual information as I need to understand the bug.

I don't know exactly for Deno direct output. Usually to log debug information from our TS code, we have a debug mode (https://quarto.org/docs/troubleshooting/#verbose-mode) and this will show any information from `debug()` call like
https://github.com/quarto-dev/quarto-cli/blob/b0a42303c1e46d85682a9841b6d058291c33fad9/src/command/render/cmd.ts#L164

We need to set `QUARTO_LOG_LEVEL` in CI then, and we could see more, especially your addition with `debug()`. 

For debugging in CI interactively, I use usually https://mxschmitt.github.io/action-tmate/ to get access to live runners and do some  interactive command or checks. If that can help...

jkrumbiegel:
@cderv So I have followed your advice and debugged a running CI session with tmate. I'm now relatively certain that this is some kind of deno problem, probably https://github.com/denoland/deno/issues/6751. As it turns out, the julia server actually starts up correctly when the first notebook is rendered via the julia engine, however deno throws the OS Error 111 when trying to connect to it. I could find out, testing interactively, that after the first fail, the server process remained functional in the background, and the next time `quarto render ...` was called everything worked.

So I believe deno has some kind of problem connecting to a child process of itself. I've added a workaround in https://github.com/quarto-dev/quarto-cli/pull/8645/commits/e280ee8d08e92d778f7215c17ef2bc3799621e57 mostly to check my assumptions, and while this step itself failed (as expected) the tests passed afterwards, accessing the server process.

I'm now not quite sure how to proceed, and what also puzzles me is why the detached child process works for Python. It seems to me that

https://github.com/quarto-dev/quarto-cli/blob/248a3a4b1cd8985eb062cd53e53b23326c13eb2d/src/resources/jupyter/jupyter.py#L216-L223

is doing something similar to what I'm doing in 

https://github.com/quarto-dev/quarto-cli/blob/8c22e07c5be7e0ee102b290c4aafb2dd2a9232c2/src/resources/julia/start_quartonotebookrunner_detached.jl#L14

Currently, I've commented out the windows runners in my script as they were hanging indefinitely so the workaround doesn't work. Any ideas for a path forward? As mentioned above, I don't see this at all on my local Mac.

cscheid:
@jkrumbiegel I've found the folks on the deno discord to be very responsive; I'm happy to mediate the discussion if you'd prefer not to join the discord, but I would appreciate if you can create a pure-deno repro for the bug in that case.

jkrumbiegel:
Another update - I was probably incorrectly ascribing the problems to deno. I found one race condition in the server startup that fixed problems on Linux, although it never manifested on my local machine which hid it from my view for a while.

The problem of hanging test sessions on windows might still be related to how deno handles subprocesses, or it's a julia problem, I don't really know. After I managed to get things to run reusing the python detached process logic you're using for jupyter, I managed to work around the problem on windows using a powershell command. Now the tests passed on my own fork. (I just think in the meantime quarto changed a bit so I get failures here at the moment). But overall, good news.

cderv:
That is indeed good news. 

> The problem of hanging test sessions on windows might still be related to how deno handles subprocesses, or it's a julia problem, I don't really know.

Did you test this locally on windows ? if not, I can do this. I am running windows locally so I can see if your initial solution works. 

Though I believe using PowerShell is ok - we use it another place to unzip when installing extensions on Windows. 

Some context on windows situations that I am thinking of 

We are currently using `Deno.run()` in `execProcess()`, and I am thinking there could be some differences on Windows. So maybe there is indeed an issue. Converting to `Deno.Command()` is currently on-going
* #8893 
I don't know if we identified problem yet with Jupyter on windows. 

It was for quoting problems with `Deno.run()` but we needed to execute some command through a `.bat` file 
https://github.com/quarto-dev/quarto-cli/blob/a41e0e9f206218011a52d5202718d5e93273019e/src/core/windows.ts#L143-L159

Anyhow, if this is working with Powershell, I believe this is ok - powershell 5 should be available on Windows environment, running Quarto.

jkrumbiegel:
Yes I went with the solution after seeing multiple uses of powershell in the code base.

> Did you test this locally on windows ? if not, I can do this. I am running windows locally so I can see if your initial solution works.

If you want to try it, you just need to comment out the if branch with the windows powershell logic and use the julia branch instead. You should be able to render documents from the command line, at least I was able to on CI. It's just when you actually run the bucket tests that the test sets never finish with this method.

> Converting to Deno.Command() is currently on-going

I am already using `Deno.Command` because at the very beginning I already had issues with hangs using `execProcess`, and I was a bit intimidated by the complexity in that function. `Deno.Command` just worked and wasn't deprecated, so I just went with that.

cscheid:
> Converting to Deno.Command() is currently on-going

Just to add a bit of context, Quarto is blocked on converting to Deno.Command because of [this upstream issue](https://github.com/denoland/deno/issues/22621) If you're not using our calls to `execProcess` (and sophisticated stream redirection), I imagine you won't see problems.

jkrumbiegel:
Hm did something recently change on master that could have made my tests fail again? The failure messages in CI don't really tell me anything. On my own fork where I hadn't updated master yet I only got failures after merging it in:

image

cscheid:
Yes, they have; engines have a new method `markdownForFile`; it exists to manage things like percent scripts and ipynb inputs. In your case, it should be enough to copy the implementation from `kMarkdownEngine`, namely

```ts
  markdownForFile(file: string): Promise<MappedString> {
    return Promise.resolve(mappedStringFromFile(file));
  },
```

I forgot to mention this to you, I apologize. I've been slowly making changes to the ExecutionEngine types which will (eventually) make it easier for third-party engines like your to exist, but you're getting caught in the crossfire.

jkrumbiegel:
> a new method markdownForFile

That one failed the first builds after I pushed again to the origin repo, yes, but I fixed that quickly. Now only the windows runners are failing again after I had their tests pass a couple times on my fork's CI. So maybe it's something else that happened in the meantime, I just don't have an idea right now what. Or the passes I got were spurious. The error with which CI fails now is really nondescript though, it doesn't give me an idea where to look.

cscheid:
I was away from the repo for a work trip, and I see the tests pass now 🎉 ! Let me think and discuss with the group about how we want to document and advertise this, but I'm looking forward to merging this soon. Thank you for the patience and the work!

jkrumbiegel:
Great to hear that!

I mostly see three remaining items that I still wanted to tackle for the first version

- a key signing mechanism that the jupyter kernels use to ensure the kernel doesn't execute code that doesn't come from the user that started it
- a timeout for the worker processes, so that they don't hang around forever
- a quarto command that allows to start / check / stop the server process. Right now you can only start it implicitly by rendering a notebook

cscheid:
> a quarto command that allows to start / check / stop the server process. Right now you can only start it implicitly by rendering a notebook

I suspect this will be harder for us to figure out, simply because there's no analogous commands right now (we also only allow implicitly starting the jupyter daemon, for example).

Software-wise it'd be easy to add a new command under `src/command`. But this would be a hard thing to teach users and fit in the way our docs are structured right now.

jkrumbiegel:
> But this would be a hard thing to teach users and fit in the way our docs are structured right now.

How do you handle this on the python side with the kernels, do you simply trust that they will all be gotten by timeouts at some point? I guess I could add a main timeout to the server process and the workers, so as soon as all the workers have timed out, the server would also be allowed to do so.

Currently, the only way to kill the server process (except restarting the computer) is to `kill $pid` where you can get the pid by passing `--execute-debug`. Having it time out seems sensible.

jkrumbiegel:
@cscheid with QuartoNotebookRunner 0.7 I have added the HMAC verification now, as well as timeout features on the individual "kernels" (worker processes held by the main service process) and a timeout on the service process itself. Those were the last larger items I wanted to tackle before a review, I'll therefore switch this from draft mode to review mode now.

cscheid:
Picking this up again. The only thing missing from this is writing the docs for it, and we prefer to have the docs written to merge together with the feature.

I think the most appropriate page to describe this new feature is https://quarto.org/docs/computations/julia.html.

Specifically, I think we could split that page into two main H2 headings, for example "Using the `jupyter` engine" and "Using the `julia` engine".

I'd be happy to give you concrete advice in how the docs look, but it might be easier for us to work on a fork of quarto.org that you have commit rights to. If that's ok with you, I'll fork quarto-web on my local repo and add you (and colleagues) as collaborators. We can coordinate here or over email, however you prefer.

jkrumbiegel:
Sounds good, coordinating over github seems most straightforward I think. We can open an issue on the fork for example.

jkrumbiegel:
The docs PR is here: https://github.com/cscheid/quarto-web/pull/1

I've reverted the windows special spawning behavior back in, the tests were passing but I could see from the timings that deno was still waiting for the unref'd process to exit (with a timeout).

jkrumbiegel:
@cscheid what are the next steps, is it ok if I occasionally keep pushing minor changes to this branch until it is merged? I'm not sure if any review activity is in progress, so I don't know if I keep making the PR a moving target by doing this.

jkrumbiegel:
Bumping

cscheid:
Thanks for bumping; I'm so sorry I missed this in the notification flood. It's totally ok to keep adding, but I think we're ready to merge. I'll review the docs, merge there and then here. Thanks again!

cscheid:
This should be available on pre-releases as soon as we merge the docs (ie, hopefully later today). We plan to make 1.5 a stable release around May-June.

cscheid:
@jkrumbiegel it's merged 🎉 . I do have a few post-merge things I want to do to improve this. I'll open a separate issue to track.

PR #8753, opened 2024-02-16
fix: About page behaviour changed between 1.4.533 and 1.4.534

### LINKED ISSUE #8544: About page behaviour changed between 1.4.533 and 1.4.534
It appears between 1.4.533 and 1.4.534, the elements of "about" is no longer respecting the position defined by the use of the id of the about element.
Moving the fenced div around has no effect.

Reproducible example at: https://github.com/mcanouil/quarto-issues-experiments/tree/main/quarto-cli-8544

| v1.4.533 | v1.4.534 |
|--------|--------|
| image | image | 

If you put the whole "content" in a fenced div, everything stays together but still on top of the about element.

image

mcanouil:
Fixes #8544

This is a follow up on #8585

CC @cwickham 

| Style | Status    | Before | After |
|-------|-----------|--------|--------|
| Solana | Incorrect | image | image |
| Broadside | Incorrect | image | image |
| Trestles | Correct | image | image |
| Jolla | Correct | image | image |
| Solana | Incorrect | image | image | 
| Marquee | Correct | image | image |

### DIFF
--- a/src/resources/projects/website/about/about.scss
+++ b/src/resources/projects/website/about/about.scss
@@ -107,7 +107,6 @@ div.quarto-about-solana {
   flex-direction: column;
   padding-top: 3em !important;
   padding-bottom: 1em;
-  grid-row: content-top/content-bottom;
 
   .about-entity {
     display: flex !important;
@@ -264,7 +263,6 @@ div.quarto-about-broadside {
   display: flex;
   flex-direction: column;
   padding-bottom: 1em;
-  grid-row: content-top/content-bottom;
 
   .about-main {
     display: flex !important;


dragonstyle:
Is the Solana template correct in the above example (I can't completely understand what I'm seeing, but it strikes me as not being correct)?

mcanouil:
It is correct (as far as I can tell) after the change, at least regarding positioning of "external content".

dragonstyle:
I pushed a commit which goes further than this PR (it is this exact fix but for all templates). I also modified your test case slightly to illustrate the various permutations (attached as zip).

[quarto-cli-8544.zip](https://github.com/quarto-dev/quarto-cli/files/14336718/quarto-cli-8544.zip)

I believe at the end of this fix, all is working as designed.

PR #8811, opened 2024-02-21
Improve default in create project when no-prompt 

### LINKED ISSUE #8809: new quarto create project options missing a default title  when non prompt
@dragonstyle follow up on 45df91d9bf71e32a67fedfaf9b8b35480c9cdeeb

````
> quarto create project default test-project --no-prompt --no-open
ERROR: TypeError: Cannot read properties of undefined (reading 'split')

Stack trace:
    at capitalizeTitle (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/core/text.ts:28:14)
    at Object.createArtifact (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/create/artifacts/project.ts:194:24)
    at Command.fn (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/create/cmd.ts:101:55)
    at eventLoopTick (ext:core/01_core.js:183:11)
    at async Command.execute (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/vendor/deno.land/x/cliffy@v0.25.4/command/command.ts:1790:7)
````

I think there is no default title defined when using `--no-prompt`.  

Is this required to pass one ? 

If so, we should probably just error better, and we're missing a test on create project with `--no-prompt` as we would have found the issue with this title addition (it would have required to modify the test)

But otherwise, using same default as when you are prompt is also an option
````
> quarto create project
 ? Type » default
 ? Directory » test-project
 ? Title (test-project) »
````

https://github.com/quarto-dev/quarto-cli/blob/5446285d6aebd56cca2785610edad9c18ea06ea3/src/command/create/artifacts/project.ts#L177-L186

cderv:
This PR improve user experience fro 
````bash
quarto create project <type> <directory> <title> --no-prompt
````

This will now error like this 
````powershell
> quarto create project --no-prompt
ERROR: A directory is required for project creation with `quarto create project`
````

and passing no title will use the directory name 
````powershell
> quarto create project default test-project --no-prompt --no-open
WARNING: No 'title' for project provided in `quarto create project`. Using 'test-project' as default.
Creating project at C:\Users\chris\AppData\Local\Temp\RtmpCyCVMf\file7824480e6a30\test-project:
  - Created _quarto.yml
  - Created test-project.qmd
````

with a warning to indicate the non provided argument 

If that is ok, then I'll try to add test creation with `--no-prompt`

I have some in Quarto R package and this is how I found this in the first place. 

fixes #8809

### DIFF
--- a/src/command/create/artifacts/project.ts
+++ b/src/command/create/artifacts/project.ts
@@ -25,6 +25,7 @@ import { join } from "path/mod.ts";
 
 // ensures project types are registered
 import "../../../project/types/register.ts";
+import { warning } from "log/mod.ts";
 
 const kProjectTypes = projectTypes();
 const kProjectTypeAliases = projectTypeAliases();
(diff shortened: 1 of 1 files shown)


dragonstyle (approved):
This looks like a good fix, once you make the type system happy :)

PR #8825, opened 2024-02-21
typst theorems

gordonwoodhull:
Here are typst theorems, along with tests for a few formats and a row in the feature-format matrix.

It was easier to use the [typst-theorems](https://github.com/sahasatvik/typst-theorems) extension than to simplify the feature wrapper. 

The light green background color is from their example and hardcoded here... haven't looked at how to add new parameters and defaults.

image

Seems like black magic getting the newlines to work out right in pandoc (and typst may be more sensitive to these). Hoping I will understand blocks and inlines better over time.

### DIFF
--- a/src/resources/filters/customnodes/theorem.lua
+++ b/src/resources/filters/customnodes/theorem.lua
@@ -99,7 +99,8 @@ local function ensure_typst_theorems(reftype)
   if not letted_typst_theorem[reftype] then
     letted_typst_theorem[reftype] = true
     local theorem_type = theorem_types[reftype]
-    quarto.doc.include_text("in-header", "#let " .. theorem_type.env .. " = thmbox(\"" .. theorem_type.env .. "\", \"" .. theorem_type.title .. "\", fill: rgb(\"#eeffee\"))")
+    quarto.doc.include_text("in-header", "#let " .. theorem_type.env .. " = thmbox(\"" ..
+     theorem_type.env .. "\", \"" .. theorem_type.title .. "\")")
   end
 end
 
@@ -160,11 +161,9 @@ end, function(thm)
     if name then
       tappend(callthm.content, name)
     end
-    quarto.log.output(el.content)
     callthm.content:insert(pandoc.RawInline("typst", "\")["))
     tappend(callthm.content, quarto.utils.as_inlines(el.content))
     callthm.content:insert(pandoc.RawInline("typst", "] <" .. el.attr.identifier .. ">"))
-    quarto.log.output(callthm.content)
     return callthm
 
   else
--- a/src/resources/filters/customnodes/theorem.lua
+++ b/src/resources/filters/customnodes/theorem.lua
@@ -88,6 +88,22 @@ _quarto.ast.add_handler({
   end
 })
 
+local included_typst_theorems = false
+local letted_typst_theorem = {}
+local function ensure_typst_theorems(reftype)
+  if not included_typst_theorems then
+    included_typst_theorems = true
+    quarto.doc.include_text("in-header", "#import \"@preview/ctheorems:1.1.0\": *")
+    quarto.doc.include_text("in-header", "#show: thmrules")
+  end
+  if not letted_typst_theorem[reftype] then
+    letted_typst_theorem[reftype] = true
+    local theorem_type = theorem_types[reftype]
+    quarto.doc.include_text("in-header", "#let " .. theorem_type.env .. " = thmbox(\"" .. theorem_type.env .. "\", \"" .. theorem_type.title .. "\", fill: rgb(\"#eeffee\"))")
+  end
+end
+
+
 _quarto.ast.add_renderer("Theorem", function()
   return true 
 end, function(thm)
@@ -136,7 +152,21 @@ end, function(thm)
     -- JATS XML theorem
     local lbl = captionPrefix({}, type, theorem_type, order)
     el = jatsTheorem(el, lbl, name)          
-    
+
+  elseif _quarto.format.isTypstOutput() then
+    ensure_typst_theorems(type)
+    -- el.content:insert(1, pandoc.RawInline("typst", "#" .. theorem_type.env .. "(\"" .. thm.name .. "\")["))
+    local callthm = pandoc.Para(pandoc.RawInline("typst", "#" .. theorem_type.env .. "(\""))
+    if name then
+      tappend(callthm.content, name)
+    end
+    quarto.log.output(el.content)
+    callthm.content:insert(pandoc.RawInline("typst", "\")["))
+    tappend(callthm.content, quarto.utils.as_inlines(el.content))
+    callthm.content:insert(pandoc.RawInline("typst", "] <" .. el.attr.identifier .. ">"))
+    quarto.log.output(callthm.content)
+    return callthm
+
   else
     -- create caption prefix
     local captionPrefix = captionPrefix(name, type, theorem_type, order)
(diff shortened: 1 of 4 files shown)


tests/docs/smoke-all/crossrefs/theorem/theorem-1.qmd
```diff
@@ -0,0 +1,44 @@
+---
+keep-typ: true
```
cscheid, line 2: Is `keep-typ` necessary here (and in the other file as well)?

I seem to remember that we changed our tests to do that automatically when the test spec includes a `typst` entry (similarly for pdf). Another way that we achieve this is by changing `output-ext` in format. This is a bit of magic right now, but

```
format:
  typst:
    output-ext: typ
```

should also make it so that Quarto stops at `typ` rather than producing a PDF file.

src/resources/filters/customnodes/theorem.lua
```diff
@@ -88,6 +88,22 @@ _quarto.ast.add_handler({
   end
 })
 
+local included_typst_theorems = false
+local letted_typst_theorem = {}
+local function ensure_typst_theorems(reftype)
+  if not included_typst_theorems then
+    included_typst_theorems = true
+    quarto.doc.include_text("in-header", "#import \"@preview/ctheorems:1.1.0\": *")
+    quarto.doc.include_text("in-header", "#show: thmrules")
```
cscheid, line 97: I think this will be the place where we change the background color of the theorem. We'll have to learn some typst to do it, but I think our theorem should try to match LaTeX's typesetting a bit more, and the light green background is somewhat jarring.

cscheid:
> Seems like black magic getting the newlines to work out right in pandoc (and typst may be more sensitive to these).

There's no easy way out of this one (as you can observe by the fact that our output is littered with spurious line breaks). Right now I think this is good, and we shouldn't try too hard to emit squeaky-clean typst.

> Hoping I will understand blocks and inlines better over time.

The general model is that a Pandoc top-level document has a list of Block elements. Hierarchical elements contain elements of the same "type": Divs are Blocks and can only contain Blocks; Spans are Inlines and can only contain Inlines. The `Plain` and `Para` elements are special cases of Blocks that can only contain inlines. Usually, `Plain` is the node that fixes the "div-to-inline impedance mismatch".

[Pandoc writers](https://github.com/jgm/pandoc/tree/main/src/Text/Pandoc/Writers) walk the AST and emit the output however they choose. The [Typst writer](https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Writers/Typst.hs) appears to use `vcat` for many of its Blocks visitors. I can't find the definition of `vcat` there, so it must be coming from one of the unqualified imports. The rest of the code base has

```hs
vcat :: [Text] -> Text
vcat = T.intercalate "\n"
```

Some other parts of their source have code like `if isTightList defs then vcat else vsep` which leads me to suspect that `vsep` is going to be something like Text.intercalate "\n\n" (and `vsep` seems to only be used in the context of deciding whether a node is a [tight list](https://pandoc.org/MANUAL.html#extension-compact_definition_lists) or not).

cscheid (commented):
This looks great, but we should try to get rid of the light background color before merging.

src/resources/filters/customnodes/theorem.lua
cscheid: We could at least provide customization at the document level using quarto metadata. It'll be a chance to walk you through a new part of the code.

src/resources/filters/customnodes/theorem.lua
```diff
@@ -88,6 +88,22 @@ _quarto.ast.add_handler({
...
+  end
+  if not letted_typst_theorem[reftype] then
+    letted_typst_theorem[reftype] = true
+    local theorem_type = theorem_types[reftype]
+    quarto.doc.include_text("in-header", "#let " .. theorem_type.env .. " = thmbox(\"" .. theorem_type.env .. "\", \"" .. theorem_type.title .. "\", fill: rgb(\"#eeffee\"))")
```
cscheid: > The green is in the example code I used a few lines down

Would you look at that. Reading! 🤦

tests/docs/smoke-all/crossrefs/theorem/theorem-1.qmd
gordonwoodhull: I didn't know about `output-ext`, changing the two tests to use it.

The tests fail with `No such file or directory` otherwise...

cscheid: > I didn't know about `output-ext`, changing the two tests to use it.

Originally it's there to allow `.html` -> `.htm` hacks and so on, but we've now added puns for latex and typst where if you use `tex` or `typ` respectively, we don't actually call the final output recipe.

> The tests fail with No such file or directory otherwise...

That makes sense.

cderv: I think you remember that because we had the discussion on slack when you added `ensureTypstFileRegexMatches`

Using `ensureTypstFileRegexMatches` requires `keep-typ: true` . You added a comment to remember also

https://github.com/quarto-dev/quarto-cli/blob/3dcd0d11842d94686a3607c95dbbe3a842d3e83c/tests/verify.ts#L238-L250

It will match `.pdf` file, and look for the `.typ` file with same name. 

Before this addition, when I did the first typst test I just used  `output-ext: typ` and `ensureFileRegexMatches()` 

`ensureTypstFileRegexMatches` will run a PDF conversion so it is better to test that the .typ produce render to `.pdf`, but when this is not needed the other way can be used. Though running Typst is quite fast. 

Hope it helps clarify the situation (which is not the best but works ok)

gordonwoodhull:
Pushed fixes for the above.

> our theorem should try to match LaTeX's typesetting a bit more

It's unclear to me (I'm sure I will learn over time) how much we should be getting Typst to imitate LaTeX and how much we should follow styles and conventions of the Typst world.

Here are screenshots of LaTeX and Typst output for the example:
image

image

Differences spotted (there are probably others):
* LaTeX uses italics for the theorem text
* Typst uses larger margins
* LaTeX creates a hyperlink for the crossref
* Typst has a fancy ligature on Th and (more aggressive?) hyphenation

cscheid:
> - LaTeX uses italics for the theorem text
> - Typst uses larger margins

That's ok for now (I don't think we're consistent on those between HTML, DocX and PDF)

> - LaTeX creates a hyperlink for the crossref

If I read your .typ output, so will typst, but you have to hover on the PDF.

> Typst has a fancy ligature on Th and (more aggressive?) hyphenation
That's to be expected. I think we're good to go here!

cscheid (approved):
LGTM!

gordonwoodhull:
Great, will merge once the checks complete!

> hover the link on the PDF

Confirmed, there is no visual styling but the link is there.

src/resources/filters/customnodes/theorem.lua
gordonwoodhull: to be fair it was an extremely long line of code (wrapped now)

cderv:
Just to share about the color on links in PDF. 

We do default in Quarto add the color on non print PDF 
https://github.com/quarto-dev/quarto-cli/blob/3dcd0d11842d94686a3607c95dbbe3a842d3e83c/src/command/render/pandoc.ts#L1062-L1081

As you can see this is a Pandoc variable that activate the feature in the template
https://pandoc.org/MANUAL.html#links

They didn't add the variable for Typst. So non links by default or opt-in.

Issue #5689, opened 2023-05-25
sidebar with single sidebar entry gets shown across entire website


machow:
### Bug description

When I create a quarto website with a single sidebar, it gets shown across all pages. However, if I add an additional sidebar entry, e.g. with `- id: dummy-sidebar`, it shows the original sidebar on only the relevant pages.

(I'm running quarto version 1.3.340)

Here an example setup where the sidebar displays on the home page.

**_quarto.yml**

```yaml
project:
  type: website

website:
  sidebar:
  - contents:
    - reference/index.qmd
    id: reference

  # note that uncommenting below fixes the issue
  #- id: dummy-sidebar
```

**index.qmd**

```
---
title: homepage
---

I am a test page
```

**reference/index.qmd**

```
---
title: reference
---

I am the reference page
```

### Checklist

- [X] Please include a minimal, fully reproducible example in a single .qmd file? Please provide the whole file rather than the snippet you believe is causing the issue.
- [X] Please [format your issue](https://quarto.org/bug-reports.html#formatting-make-githubs-markdown-work-for-us) so it is easier for us to read the bug report.
- [X] Please document the RStudio IDE version you're running (if applicable), by providing the value displayed in the "About RStudio" main menu dialog?
- [X] Please document the operating system you're running. If on Linux, please provide the specific distribution.
- [X] Please provide the output of `quarto check` so we know which version of quarto and its dependencies you're running.

dragonstyle:
This is currently working as designed - we treat a single sidebar as 'global' by default, thinking that most commonly users will want the sidebar to behave in this global fashion. There are few different ways to workaround this using `sidebar: true/false` in document front matter or `_metadata.yml / _quarto.yml` (A few outlined here: https://github.com/quarto-dev/quarto-cli/issues/1853 ).

Would any of these solve the issue you're running into?

machow:
Hey, thanks for your response! This seems like pretty surprising behavior, and the solutions seem to involve quite a bit of extra files / tracking things. But it seems like we can just continue using a dummy entry, and that works pretty well!

I'll mention the dummy entry in the other issue, in case it's helpful.

dragonstyle:
Y there  are sort of 2 different issues here:

- That the sidebar is global (we could add an option like `global` or something that would make it simple to get the behavior you'd like without resorting to the dummy id)

[...]

PR #8871, opened 2024-02-24
Website Sidebar - global sidebar

dragonstyle:
We need to wait to automatically assign an id to a sidebar as we use undefined `id` as representing a global sidebar.

Addresses #8869

See also https://github.com/quarto-dev/quarto-cli/issues/5689

## Checklist

I have (if applicable):


dragonstyle:
Note: I think that this should likely be back ported, as the current behavior is more severe than we intended.

PR #8913, opened 2024-02-28
deno arm build are now available in denoland/deno main repo

cderv:
Before it was built in another project https://github.com/LukeChannings/deno-arm64/

This PR adapts our configuration following the update in #8907 which broke Create Installer workflow


cderv:
I have triggered manual build of the installers to check and this is now fixed for ARM installers 

See https://github.com/quarto-dev/quarto-cli/actions/runs/8078438508

PR #8979, opened 2024-03-04
[GHA] FIx CI failure for R on v1.4 patch branc

cderv:
This is short term quick fix to unblock CI. 

Possibly older version of package does not play well on windows and older R version. I'll check later. Hopefully, this will solve. 

Also, I don't know why the cache did not apply here... 🤔 I'll look into it.


cderv:
We updated all deps in #8980 so this is outdated

PR #8980, opened 2024-03-04
Update test deps on v1.4

cderv:
This cherry-pick eb0f765527ac65cb8d2ce33252016b4fbc09a2fb with also updating each file to their latest addition for same test suits in main and v1.4 branch

This should insure also the same cache is used on `main` and `v1.4` branch for patch. This will avoid long runs when doing a new monthly patch release. 

I'll be sure to update things between branch if any modification is done now that I am aware of the potential issues. 

Opening this PR to check this works ok !

closes #8979


cderv:
It was checked correctly : https://github.com/quarto-dev/quarto-cli/actions/runs/8148170998

Everything is passing. And this correctly use the cache for `main` for quicker runs.

PR #9619, opened 2024-05-08
Typst CSS

gordonwoodhull:
This is about ready to merge, once there is a new Pandoc release.

With full CSS rgb and named color support and the border shorthands, the Lua script to do CSS->Typst translation is about 4x as long as it was (now ~900 lines).

Posting the PR for review tomorrow as well as to see how the tests do.

Will follow up with open questions & areas to review, or bring those tomorrow.

This subsumes #9500.


gordonwoodhull:
Ah yes, will need Pandoc 3.1.14 to run the tests in CI

gordonwoodhull:
gt and Great Tables get some nice effects out of putting borders on spans. 

See Time and Measurement in this gt screenshot:

image

It is also possible to put borders on a Typst [#highlight()](https://typst.app/docs/reference/text/highlight/) but it would take a lot more styling for it to look good, so I've disabled it.

Naive Typst highlight with border (and transparent color 😮) looks like this:

image

I mean, gt is using `width: 100%` and `display: inline-block`, so it's not exactly obvious how to map this automatically.

image

gordonwoodhull:
Pandoc 3.2 doesn't seem to cause any failures.

Rough edges we should probably handle before merging: 
- [ ] for some reason, it's only finding 5 of the ff matrix tests on Linux, but all 174 of them on Windows. https://github.com/quarto-dev/quarto-cli/actions/runs/9069590810/job/24919483138?pr=9619
- [ ] (Intermittent?) mediabag failure on Windows: https://github.com/quarto-dev/quarto-cli/actions/runs/9068888314/job/24917216923?pr=9619
- [ ] I have seen a ff matrix test time out: https://github.com/quarto-dev/quarto-cli/actions/runs/9068888314/job/24917216549?pr=9619

cscheid:
> * [ ]  (Intermittent?) mediabag failure on Windows: https://github.com/quarto-dev/quarto-cli/actions/runs/9068888314/job/24917216923?pr=9619

I wonder if this is because we're hammering placeholder.co and we should use some other solution instead.

gordonwoodhull:
I thought the globstar feature (eg `qmd-files/**/*./qmd`) seemed like a relatively "new" Unix feature (c2010 😉).

Turning it on with `shopt -s globstar && ...`, all ff matrix tests are running.

I can repro the mediabag error locally with 

```
quarto render document.qmd -t pdf -M keep-tex:true
```

That's kind of weird, maybe coincidental.

* Does not happen without `keep-tex`
* File does not have any pdf options
* Don't think tests in CI run with `keep-tex`

gordonwoodhull:
Okay, this is only blocked by mediabag issues on both platforms. It's weird that it's always the same file, and I can repro the error locally, but I am giving up on it for the moment.

I have some hardening to do on Typst CSS inputs and need to test malformed input, which I can do before or after merge.

cscheid:
> 2\. I can repro the mediabag error locally with
> 
> ```
> quarto render document.qmd -t pdf -M keep-tex:true
> ```

Can you just clarify which `document.qmd` this is referring to?

gordonwoodhull:
> Can you just clarify which `document.qmd` this is referring to?

```
dev-docs/feature-format-matrix/qmd-files/crossref/float/figure/document.qmd
```

cscheid:
Ok - I think that failure is placeholder.co etc no longer being reliable.

gordonwoodhull:
There are very occasional mermaid timeouts, but everything else is running well.

Ok to merge?

cscheid:
Let's do it!

PR #9785, opened 2024-05-28
confluence - fix characters being added to title when title is unique

rossbowen:
When publishing to Confluence, the docs mention:

>  Quarto may add some additional characters to meet the Confluences requirement that every page in a space has a unique name.

I noticed that characters were being added even if there wasn't a page with the same title within my space. However, I noticed there were pages with the same title in _other_ spaces. To me it looks like a small error in the way the CQL has been written, which this PR amends.

### DIFF
--- a/src/publish/confluence/api/index.ts
+++ b/src/publish/confluence/api/index.ts
@@ -94,7 +94,7 @@ export class ConfluenceClient {
     const CQL_CONTEXT =
       "%7B%22contentStatuses%22%3A%5B%22archived%22%2C%20%22current%22%2C%20%22draft%22%5D%7D"; //{"contentStatuses":["archived", "current", "draft"]}
 
-    cql = `${cql}&spaces=${space.key}&cqlcontext=${CQL_CONTEXT}`;
+    cql = `${cql}%20and%20space=${space.key}&cqlcontext=${CQL_CONTEXT}`;
 
     const result = await this.get<ContentArray>(`content/search?cql=${cql}`);
     return result?.results ?? [];


src/publish/confluence/api/index.ts
```diff
@@ -94,7 +94,7 @@ export class ConfluenceClient {
     const CQL_CONTEXT =
       "%7B%22contentStatuses%22%3A%5B%22archived%22%2C%20%22current%22%2C%20%22draft%22%5D%7D"; //{"contentStatuses":["archived", "current", "draft"]}
 
-    cql = `${cql}&spaces=${space.key}&cqlcontext=${CQL_CONTEXT}`;
+    cql = `${cql}%20and%20space=${space.key}&cqlcontext=${CQL_CONTEXT}`;
```
cderv, line 97: It seems indeed this is not good : https://developer.atlassian.com/cloud/confluence/cql-keywords/

Does this means the other `&` in `&cqlcontext=${CQL_CONTEXT}` is also wrong ?

cderv (commented):
Thanks ! 

Did you test the change by any chance ? Does it solve it for you ? 

This is very helping that you did not just opened an issue, but took the time to show change in PR. Thank you

src/publish/confluence/api/index.ts
rossbowen: I believe the `cqlcontext` is fine, because it is its own query parameter. The doc I looked at was here:

https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-search/#api-wiki-rest-api-search-get-request

I can see both `cql` and `cqlcontext` are query parameters, so they can be separated with `&`.

rossbowen:
I did test but I will admit I got some errors - but different ones to the one which was above!

I took the failures to be a bit of a problem with my set up and a false positive... but all the confluence tests were passing. I'll give it another go locally!

rossbowen:
Apologies, I thought you meant running the Quarto tests!

I did a bit of manual testing with my Confluence setup, the process would be:

- Create two spaces, A and B.
- Create a document within space A with a title of "Hello world".
- Attempt to `quarto publish` a document into space B with the title of "Hello world". It should publish without any additional characters being added to the title.
- Attempt to `quarto publish` a document into space A with the title of "Hello world". It should publish _with_ additional characters.

PR #9795, opened 2024-05-29
ipynb - check if svg display data is encoded or not

### LINKED ISSUE #9793: embed svg plots are wrongly importly
From @cwickham 

* `main.qmd`
````
---
title: "Untitled"
format: typst
keep-md: true
---

{{< embed plots.qmd#plot >}}
````
* `plots.qmd`
````
---
title: "Plots"
fig-format: svg
---

```{r}
#| label: plot
plot(1:10)
```
````

````powershell
❯ quarto render main.qmd
pandoc
  to: typst
  output-file: main.typ
  standalone: true
  shift-heading-level-by: -1
  default-image-extension: svg
  wrap: none
  citeproc: false

metadata
  title: Untitled

[typst]: Compiling main.typ to main.pdf...error: failed to parse SVG (unknown token at 1:1)
    ┌─ \\?\C:\Users\chris\Documents\DEV_OTHER\00-TESTS\test-quarto\main.typ:274:5
    │
274 │ #box(image("main_files/figure-typst/plots-plot-output-1.svg"))
    │      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

ERROR: Error
    at Object.complete (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/output-typst.ts:79:13)
    at async file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render.ts:281:23
    at async withTimingAsync (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/core/timing.ts:50:20)
    at async Object.complete (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render.ts:276:7)
    at async Object.onPostProcess (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-files.ts:720:28)
    at async renderFileInternal (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-files.ts:689:3)
    at async renderFiles (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-files.ts:325:9)
    at async render (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-shared.ts:102:18)
    at async Command.actionHandler (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/cmd.ts:248:26)
````

Something is wrong with the svg file. It is not an SVG file 
````powershell
❯ head -n 2 .\main_files\figure-typst\plots-plot-output-1.svg
PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4KPHN2ZyB4bWxucz0iaHR0cDov
L3d3dy53My5vcmcvMjAwMC9zdmciIHhtbG5zOnhsaW5rPSJodHRwOi8vd3d3LnczLm9yZy8xOTk5
````
It seems like base64 encoded - and we can check it is 
````r
> substr(rawToChar(xfun::base64_decode(xfun::file_string("main_files/figure-typst/plots-plot-output-1.svg"))), 1, 50)
[1] "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<svg xmlns="
````

So something is not right in how the svg plot is created when `fig-format` is used in the embeded file. 

Reminder note: embed file works by rendering the embeded file internally to a ipynb file (like `--to ipynb`). This means **knitr** cells will execute with `knitr::pandoc_to()` being `ipynb`. **knitr** and R packages used in this context are not used to that, and may not be aware of what to do specifically, and output some unwanted defaults. 

Either this is this type of issue we had in the past, or a new one related to SVG and ipynb. 

The intermediate generated by `quarto render main.qmd` rendering for the embeded file rendering 
````
Rendering qmd embeds
[1/1] plots.qmd
````
is 
````markdown
---
title: "Plots"
fig-format: svg
keep-md: true
---


::: {.cell}

```{.r .cell-code}
#| label: plot
plot(1:10)
```

::: {.cell-output-display}
![](plots_files/figure-ipynb/plot-1.svg)
:::
:::

````

So we get a svg file inserted using Markdown syntax. Just the file content is not ok.

### PR DESCRIPTION
**Pull request by cderv**
fixes #9793 

We assumed svg mime type data was never encoded. 

However,  this is not the case when the Notebook file is produced by a pandoc convert `--to ipynb`. In this case, Pandoc will base64 encode the svg file. The svg data needs to be decoded before being written to the file at embedding time.Welcome to the quarto GitHub repo!

For Quarto, this will mostly happens in render context when embed is used as we generate this type of ipynb document internally. 
It could also happens if a Python code in cell would also output a `display_data` with svg mime type and encoded data. It does not seem to be the common usage. 

This is quite a narrow fix I believe, and as mentioned in #9793 , this is doing the same kind of check that we do in another function 
https://github.com/quarto-dev/quarto-cli/blob/122b0f3093c560a6c24365fc9241bb83d4590ef4/src/core/jupyter/jupyter.ts#L991-L997

cc @cwickham if you want to try on your side, though I added your example as a test

### DIFF
--- a/src/core/jupyter/jupyter.ts
+++ b/src/core/jupyter/jupyter.ts
@@ -1930,11 +1930,15 @@ function mdImageOutput(
     ? (data as string[]).join("")
     : data as string;
 
-  // base64 decode if it's not encoded svg
-  // when used in embed context, Pandoc will generate ipynb with base64 encoded svg data
-  // https://github.com/quarto-dev/quarto-cli/issues/9793
   const outputFile = join(options.assets.base_dir, imageFile);
-  if (mimeType !== kImageSvg || !imageText.trimStart().startsWith("<svg")) {
+  if (
+    // base64 decode if it's not svg
+    mimeType !== kImageSvg ||
+    // or if it is encoded svg; this could happen when used in embed context,
+    // as Pandoc will generate ipynb with base64 encoded svg data
+    // https://github.com/quarto-dev/quarto-cli/issues/9793
+    !/<svg/.test(imageText)
+  ) {
     const imageData = base64decode(imageText);
 
     // if we are in retina mode, then derive width and height from the image
--- a/src/core/jupyter/jupyter.ts
+++ b/src/core/jupyter/jupyter.ts
@@ -1930,9 +1930,11 @@ function mdImageOutput(
     ? (data as string[]).join("")
     : data as string;
 
-  // base64 decode if it's not svg
+  // base64 decode if it's not encoded svg
+  // when used in embed context, Pandoc will generate ipynb with base64 encoded svg data
+  // https://github.com/quarto-dev/quarto-cli/issues/9793
   const outputFile = join(options.assets.base_dir, imageFile);
-  if (mimeType !== kImageSvg) {
+  if (mimeType !== kImageSvg || !imageText.trimStart().startsWith("<svg")) {
     const imageData = base64decode(imageText);
 
     // if we are in retina mode, then derive width and height from the image
--- a/news/changelog-1.5.md
+++ b/news/changelog-1.5.md
@@ -123,6 +123,7 @@ All changes included in 1.5:
 - ([#9635](https://github.com/quarto-dev/quarto-cli/issues/9635)): Respect `{shortcodes=false}` when resolving `include` shortcodes.
 - ([#9664](https://github.com/quarto-dev/quarto-cli/pull/9664)): Add `placeholder` shortcode to produce placeholder images.
 - ([#9665](https://github.com/quarto-dev/quarto-cli/issues/9665)): Fix issue with key-value arguments of shortcode handlers in code contexts.
+- ([#9793](https://github.com/quarto-dev/quarto-cli/issues/9793)): `embed` shortcode now correctly retrieve svg image from embdedded cell.
 
 ## Lightbox Images
 
--- a/tests/docs/smoke-all/2024/05/29/9713/main.qmd
+++ b/tests/docs/smoke-all/2024/05/29/9713/main.qmd
@@ -1,0 +1,8 @@
+---
+format: typst
+_quarto:
+  tests:
+    typst: null
+---
+
+{{< embed plots.qmd#plot >}}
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: src/core/jupyter/jupyter.ts [hunk 1]
```diff
@@ -1930,9 +1930,15 @@ function mdImageOutput(
...
-  // base64 decode if it's not svg
   const outputFile = join(options.assets.base_dir, imageFile);
-  if (mimeType !== kImageSvg) {
+  if (
+    // base64 decode if it's not svg
+    mimeType !== kImageSvg ||
+    // or if it is encoded svg; this could happen when used in embed context,
+    // as Pandoc will generate ipynb with base64 encoded svg data
+    // https://github.com/quarto-dev/quarto-cli/issues/9793
+    !/<svg/.test(imageText)
```
cderv, line 1940: @cscheid @gordonwoodhull I don't know if this is robust or not. It seems simple enough. 

Otherwise, I was thinking of testing for base64 string ? 
Either using regex 
````typescript
  function isBase64(value: string) {
    return /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{4})$/
      .test(value.replace(/\s+/gm, ""));
  }
````
Or a encode decode check 

````typescript
  function isBase64(value: string) {
    value = value.replace(/\s+/gm, "");
    try {
      return encodeBase64(decodeBase64(value)) == value;
    } catch {
      return false;
    }
  }
````

Would this utility be useful or too fragile for the regex one, and too computing intensive for the second one (I don't know if decode + encode can impact performance) ?

gordonwoodhull: Given that 
* the code already assumes that every image mime type except SVG is base64 encoded
* `<` is not a valid base64 chararacter

I think your fix is correct, and there is no need for the other stuff. I would probably pin it with `/^<svg/` but it doesn't matter.

cderv: > I would probably pin it with /^<svg/ but it doesn't matter.

This is what I tried first 😅 https://github.com/quarto-dev/quarto-cli/pull/9795/commits/e607284655af9752be3ac09f999bba9d9f821710 but test failed, because it seems some figures output in ipynb will include a whole HTML doc, so not starting with `<svg` - I know - not expected 😞

PR #9595, opened 2024-05-06
fix: remove trailing / from site-url and ensure canonical gets one

mcanouil:
This pull request fixes the issue where a trailing slash was not being removed from the site URL. The `synthesizeCitationUrl` function now removes the trailing slash from the base URL before constructing the citation URL.

<details><summary>Outdated</summary>
<p>

I did not change the "else" part because I was unsure how to trigger this part of the codepath, but using the same if/else condition on length would not hurt I believe.

```diff
    } else {
      const relativePath = relative(
        rootDir,
        join(dirname(input), basename(outputFile)),
      );
      const part = pathWithForwardSlashes(relativePath);
+     if (part.length === 0) {
-     return `${baseUrl}/${part}`;
+       return `${baseUrl}/${part}`;
+     } else {
+       return `${baseUrl}/${part}/`;
+     }
    }
```

</p>
</details> 

Fixes #9524

PR #9969, opened 2024-06-12
Adapt our test logic to correctly support project with subfolders

cderv:
This PR tries to simplify how our project rendering works by improving the project detection for a tested file and the output file path detection to take into account subfolder. 

This solves the test problem at #9595 and once merged, we should see that test will pass. 

Opening this PR to see if any test is beaking due to this change in CI. 

BTW this work has surfaced another bug regaring quoting of `rel` attributes. It seems one has been missed from last time it was being fixed (f356b962eb8551e2748242d7c9bfc8de24acd861) and test was not run because not correctly name. 

I'll check other possible problem with this in our ejs template and add a changelog


cderv:
All tests passes ok in usual time, so I think this is good change.

PR #9982, opened 2024-06-12
test - running tests with timing also needs to be configured with v8-flags.

cderv:
This follows #9749  fix

We missed a spot which broke the timing of tests since then. 

This should fix it. Let's see ! 

I'll run the timing on this PR too


cderv:
Triggered at https://github.com/quarto-dev/quarto-cli/actions/runs/9488371177

cderv:
All good ! So merging. 

We should have a new file with test timing soon;

PR #10029, opened 2024-06-15
style(cosmo): harmonise value box colour (opacity)

mcanouil:
This pull request harmonises the value box opacity in the Cosmo style. The `valuebox-bg-info` and `valuebox-bg-danger` colours have been adjusted to have consistent opacity, *i.e.*, no opacity as the other value box background.

To-do: add changelog entry if approved.


cderv:
@dragonstyle if you happen pass around ? Any reason to have set transparency ? 

My feeling tells me it may be on purpose for default styling choice.

cscheid:
1. This changes more than just `cosmo`, right?
2. CSS changes are pretty hard to test for regression, and so I don't want to take them as late as we are into 1.5.

mcanouil:
This PR was not necessarily for 1.5.

The changes should only affect value box for the cosmo theme.

If not then that would mean there is something very wrong with the structure and the naming inside this SCSS file.

cscheid:
Sorry: I saw this and missed the fact that it was scoped for `cosmo`

image

So this looks good. Let's just wait for 1.6 first.

mcanouil:
A bit of context: apparently the purpose of the 70% was to reduce brightness for info/danger.

An alternative, could be to use `darken` instead.
For instance, using 2% darkening on all colours:

image

jjallaire:
Most instructive will be to look at these variations in an actual dashboard. What Charles and I found was that more saturated colors tended to take up too much "weight" visually visa-vi the rest of the dashboard, making them feel visually top-heavy. I don't doubt we could come up with improvements, but you need to look at the variations in context to really know.

mcanouil:
A real life example of a dashboard with/without opacity:

https://github.com/quarto-dev/quarto-cli/assets/8896044/f78654d5-edbb-4f1a-a1a6-77425898a8a4


(I don't have yet "darken" examples)

mcanouil:
Dropping this.

PR #10056, opened 2024-06-18
ci(bug_report.yml): fix indent

mcanouil:
The `validations` key was not indented properly as it was under `attributes`.


mcanouil:
@cderv or @cscheid the issue template is broken because the option was not indented properly.
We need to merge the fix.

cderv (approved):

mcanouil:
Thanks, the issue template is back.

PR #10148, opened 2024-06-26
tests that exercise juice via gt, pandas

gordonwoodhull:
In discussing #10134, which turned out to be `juice` failing in a Windows dev environment, @cderv pointed out that the current "tests" for html table output from packages in `tests/docs/typst/tables` [^1] are not even smoke tests, and won't get run automatically. 

There are already thorough tests for css properties in the ff matrix, but nothing that exercises `juice`. So here are typst output tests for `gt` and `pandas` that cannot succeed unless `juice` is working properly.

[^1]: Really these documents can be considered a wanna-be gallery for Typst CSS - maybe we can consider having such a place on quarto-web eventually.


cderv:
Thanks a lot ! 

I was curious to understand why the test was passing, but in fact this is because we don't call `quarto.cmd` or `quarto` sh script from CI. 

We call `deno run` on the test file which will then call directly our Typescript `quarto()` function. 

So no call to quarto.cmd - Our CI does not currently test calls to Quarto binary, or Quarto scripts. 

I'll probably add this to the list of improvement for CI. 

Thanks a lot for the PR !

PR #10612, opened 2024-08-26
typst brand yaml: typography

gordonwoodhull:
Here is PR 2/3 for typst brand yaml. 

It builds on #10543 typst brand yaml: color

Includes some schema fixes based on my interpretation of the spec.

### Font sources

Implements `brand-font-file` (and `brand-font-family`) only, not `brand-font-google` yet. My guess is that google forge and URL sources will be downloaded on the front-end and resolved files will be passed to the backend. I am not sure if this is correct or not.

To this end, `brand-typography-options` now optionally can have `files`. Otherwise we end up with more types for all the "resolved" typography options and that gets messy.

### Open but typed object in schema?

I don't now how to encode `brand-font-with` correctly in the schema. It is an open object where the keys are strings and values are `brand-font`. So `typography.with` is not properly validated.


gordonwoodhull:
Superceded by #10647

PR #10893, opened 2024-09-26
Trying update action/upload-artifacts to v4

### LINKED ISSUE #7922: [GHA] Adapt our workflow to benefit from action artifacts updates
Several improvement in V4 : https://github.com/actions/toolkit/tree/main/packages/artifact#v2---whats-new

However, it is breaking change and we need to adapt the workflow. 

* https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/

cderv:
Closes #7922


cderv:
Only breaking change for us is that we produce now 3 artifacts of `tinitex`. So I'll update the TinyTeX repo action that needs to be adapted for the downloads

PR #10939, opened 2024-10-01
Update table term for french in _language.yml.

famuvie:
Use Tableau instead of Table.

Is there a standard reference for these terms? Working in France, I find that my French colleagues use the term "Tableau" instead of "Table" in reports.

## Description

Change the term "Table" by "Tableau" in the French language specification, according to common use.

## Checklist

I have (if applicable):


cderv:
> Change the term "Table" by "Tableau" in the French language specification, according to common use.

I think this is your proposal. but this PR is invalid because you are modifying the default file, and not the french version of this. 

We do use `Table`

https://github.com/quarto-dev/quarto-cli/blob/01b2f041fb633b2b9bd597dc938e0b9d1b271c0b/src/resources/language/_language-fr.yml#L83

https://github.com/quarto-dev/quarto-cli/blob/01b2f041fb633b2b9bd597dc938e0b9d1b271c0b/src/resources/language/_language-fr.yml#L98

@mcanouil @maelle What do you think of Tableau instead of Table ? 🤔

cderv (changes requested):
Wrong file modified in this proposal

mcanouil:
I think we already discussed this somewhere.

> according to common use.

A source? As far as I know, both are correct and are heavily used.
Since the language can be customised, I would not change anything here.

cderv:
Thanks I forgot this discussion. 

Agreed on the breaking change. 

@famuvie are you aware you can change the translation ?

famuvie:
> @famuvie are you aware you can change the translation ?

Yes, thank you.

If both alternatives are arguably valid, I'm fine with it. I just wanted to contribute a fix I keep having to do because my French collaborators keep _correcting_ this.

Sorry for missing the right file 🤷

maelle:
I don't have any specific opinion but out of curiosity I went and looked at what Microsoft Office uses... Tableau apparently? https://support.microsoft.com/fr-fr/office/ins%C3%A9rer-un-tableau-a138f745-73ef-4879-b99a-2f3d38be612a

famuvie:
> I don't have any specific opinion but out of curiosity I went and looked at what Microsoft Office uses... Tableau apparently? https://support.microsoft.com/fr-fr/office/ins%C3%A9rer-un-tableau-a138f745-73ef-4879-b99a-2f3d38be612a

That explains why my collaborators insist in using _Tableau_ 😂

mcanouil:
Well, they actually use both

> Pour une table plus grande ou pour personnaliser un tableau, sélectionnez Insérer > tableau > Choisir une ligne et une colonne.

cderv:
@famuvie thanks a lot for the feedback. 

FWIW I do think that this translation has also been made "Table" to match LaTeX which use also "Table" with french translation for babel; 

Considering it can be customized project-wise, and through custom extension, I don't think we'll change the default. 

Again thanks a lot for the feedback !

PR #10972, opened 2024-10-04
Don't use deno cache with Quarto run when using dev version

cderv:
Follow up on 
- https://github.com/quarto-dev/quarto-cli/pull/10940

we deactivated a failing test, and this solves it. 

`quarto run` with dev version needs to not set `--cached-only` as we don't vendor anymore in dev version. 

BTW, I used what we consider to be the mark of dev mode which is `QUARTO_DEBUG`. See comments 
https://github.com/quarto-dev/quarto-cli/blob/5c38d77c240f3eb2a6ebf5239a31cf7c7ea9be07/src/core/quarto.ts#L64-L67
https://github.com/quarto-dev/quarto-cli/blob/5c38d77c240f3eb2a6ebf5239a31cf7c7ea9be07/src/core/quarto.ts#L94-L95

We do set `QUARTO_DEBUG` in 
- `quarto.cmd` for windows
- `quarto` for UNIX
- our `run-tests` scripts

### DIFF
--- a/src/core/run/deno.ts
+++ b/src/core/run/deno.ts
@@ -12,6 +12,7 @@ import { architectureToolsPath, resourcePath } from "../resources.ts";
 import { RunHandler, RunHandlerOptions } from "./types.ts";
 import { removeIfExists } from "../path.ts";
 import { copyTo } from "../copy.ts";
+import { quartoConfig } from "../quarto.ts";
 
 export const denoRunHandler: RunHandler = {
   canHandle: (script: string) => {
@@ -42,7 +43,8 @@ export const denoRunHandler: RunHandler = {
           "run",
           "--import-map",
           importMap,
-          "--cached-only",
+          // --cached-only can only be used in bundles as vendoring is not done anymore in dev mode
+          ...(quartoConfig.isDebug() ? [] : ["--cached-only"]),
           "--allow-all",
           "--unstable-kv",
           "--unstable-ffi",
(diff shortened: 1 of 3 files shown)


src/core/run/deno.ts
cderv, line 1: I am using `quartoConfig` from `../quarto.ts`

If this is not ok to import from this file in `deno.ts`, then we could directly read the env var
````
getenv("QUARTO_DEBUG", "false") === "true"
````

cderv:
I can't even run it on windows locally 
````
running 1 test from ./smoke/run/stdlib-run-version.test.ts
[unit] > stdlib-run-version ...
------- output -------
[verify] > stdlib-run-version
{
  result: {
    success: false,
    code: 1,
    stdout: "",
    stderr: "ERROR: Quarto requires reconfiguration to install Deno version v1.46.3. Please run `./configure.cmd` command.\n" +
      "\n"
  }
}
----- output end -----
[unit] > stdlib-run-version ... FAILED (356ms)
````

Some windows thing regarding re-configuration detection in dev more. `quarto run` works on the file, but running it from deno test is not working well. 

I'll just ignore again

So I'll just ignore.

cderv:
> But you _can_ do `quarto run docs/run/test-stdlib.ts` locally?

Yes exactly. No error when calling `quarto run docs/run/test-stdlib.ts` on Windows. 

The problem only is with `./run-tests.ps1 smoke\run\stdlib-run-version.test.ts`

PR #10974, opened 2024-10-04
Typst brand yaml: typography bells and whistles

gordonwoodhull:
This is in good shape and has basic test coverage for all supported feature-element combinations.

image

Also includes support for subtitles in typst, taking @mcanouil's contribution to the Pandoc template from upstream #9638, and styling it as a heading.

Leaving this as a draft for now because although I'd be happy merging this in its current state, there are a few last touches to make:

- [x] Remove less-useful feature-element combinations 

image

- [x] Typst `brand.typography` should allow use of named colors from `brand.color` for `color`, `background-color`
- [x] Support `rem` correctly when `brand.typography.base.size` is specified. Current blind change to `em` will be the fallback with a warning when `base.size` is unknown
- [x] Translation of CSS `padding` and `width` for logo - currently one must use Typst-compatible units.
- [x] Refactor CSS->Typst translation functions to a module used by typst brand yaml and typst css


gordonwoodhull:
Hmm, perhaps I should not use `ensureTypstFileRegexMatches` / `pdftotext` so much. 

It is helpful for testing that the font caching pipeline is working, because I can 

```typst
#set text(fallback: false)
```

and Typst won’t output fonts it doesn’t have.

But `pdftotext` (and pdf extraction in general IIUC) is somewhat non deterministic in how strings are broken into words and lines (or rather, how it regroups the letters).

So I’ve got one falling test which I give up on for today. But otherwise this is ready.

src/resources/formats/typst/pandoc/quarto/typst-template.pdf
cderv, line 1: @gordonwoodhull do we want to keep the PDF committed ? 
Is this to have an example of default template ? 

Just wondering, it case it was for testing initially

gordonwoodhull: Thanks @cderv! 

Nope, it looks like this artifact is generated when the `.typ` is edited in vscode. (Seems to happen with any valid `.typ` file.) There is no reason to compile a template on its own.

I've removed it.

PR #10975, opened 2024-10-04
Update the hash in url for deno 1.46.3 from conda-forge

cderv:
This should fix RHEL build

Let's check at: https://github.com/quarto-dev/quarto-cli/actions/runs/11185287747


cderv:
All good RHEL is building ! so merging

PR #11034, opened 2024-10-11
Add support for new Scroll View mode of Revealjs

cderv:
This is a new feature from 1.5 : https://revealjs.com/scroll-view/

This PR adds: 

- Support for configuration through YAML `scroll-view` key. `scroll-view: true` is a shortcut for 
  ```
  revealjs:
    scroll-view:
      activate: true
  ```
  meaning to start the presentation in scroll view by default.

- a menu tool item to toggle scroll view on and off, associated with new R keyboard shortcut

- Adapt our custom footer processing to work in Scroll view mode.

- tests for our Quarto specific support (not for the scroll view feature itself as this is revealjs)

There may still be some other side effect I did not find. 

We could also so some still choices by default (like hide chalkboard button by default in scroll mode 🤷 )

I think we may discover a few side effect of this new mode, possibly also because of some Revealjs bug. We'll see

## TO-DO

- [ ] Document this new feature in Quarto-web


cderv:
Let's document after merging as discussed.

PR #11069, opened 2024-10-15
brand yaml: relative logo paths

gordonwoodhull:
Couple questions for @cscheid here, so leaving this as draft:

1. This implements the relative path to the brand yaml logo resources in `core/brand/brand.ts`. I think the implementation is good but I wonder if this belongs somewhere else? since I don't find the similar implementation for `typography` in this source file (but `typography` brand.yml-relative paths do work, as tested in this PR).
3. I've been populating the `brand` key in the metadata with an object containing some of the processed brand yaml data, e.g. `brand.typography.headings.family` etc., for the template to consume.

   However, this may conflict when the metadata contains an explicit `brand: path/brand.yml` entry.

   Here I replace the string (`Inlines`) with an object, which feels problematic. Should we rename `brand` data to something else? You expressed the opinion that this data should be available to custom lua filters.


gordonwoodhull:
As discussed, we can clean up the implementation of logo paths later, and replacing `brand` in metadata with an object is fine, as probably the path should not be emitted there anyway.

PR #11074, opened 2024-10-16
typst logo padding shorthand

### LINKED ISSUE #10982: Typst logo.padding could be a shorthand
For now (#10974) it is a scalar CSS length which gets applied on all sides of the logo. (As Typst `inset`.)

But it could be upgraded in a backward-compatible way to a shorthand for `padding-top`, `padding-right`, etc. like CSS [padding](https://developer.mozilla.org/en-US/docs/Web/CSS/padding).

The code already exists e.g. for `border-width`, which is the same kind of shorthand. It just needs to be abstracted somewhat.

gordonwoodhull:
Closes #10982 

In order to provide complete control over logo placement, the padding on each side should be independent.

The right way to do this is to implement the [CSS `padding` shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/padding) along with `padding-left`, `padding-top`, ...

This required refactoring a lot more of `typst-css-property-processing.lua` to `_quarto.modules.typst.css`

Also adds

* `_quarto.modules.typst.as_typst_dictionary`
* <strike>`_quarto.modules.typst.sortedPairs`</strike> `_quarto.utils.table.sortedPairs`


gordonwoodhull:
Also merging this once tests pass...

PR #11113, opened 2024-10-18
Fix Typst table wrapping to prevent font family and size spill

### LINKED ISSUE #11093: Font family and size can spill out of Typst raw html blocks
Since we use `typst:no-figure` on our tables (implemented in jgm/pandoc#9778), Pandoc may not enclose tables in blocks.

Unfortunately, when the Pandoc Typst writer encounters `typst:text:font`or `typst:text:size` on a table, it emits `#set text(...)` before the table, which will affect the entire scope.

E.g.

````
---
format: typst
---

Normal paragraph text.

```{=html}
<table style="font-size: 8px;">
    <tr><td>A</td><td>B</td></tr>
</table>
```

Normal paragraph text.
````

produces

```typst
Normal paragraph text.

#set text(size: 6pt); #table(
  columns: 2,
  align: (auto,auto,),
  [A], [B],
)
Normal paragraph text.
```

Image

Oops!

One way to correct the Typst output is to wrap the `set text` + table in `#[...]`:

```typst
Normal paragraph text.

#[#set text(size: 6pt); #table(
  columns: 2,
  align: (auto,auto,),
  [A], [B],
)]
Normal paragraph text.
```

Plan:

- [x] Add temporary patch, wrapping in `#[]` block, to our Typst post-processing filter here:

https://github.com/quarto-dev/quarto-cli/blob/412ea639aec65a500d83134bd3202a8b799aca55/src/resources/filters/quarto-post/typst.lua#L122-L126

- [x] File bug on Pandoc illustrating the bug in pure Pandoc
- [x] Submit fix upstream to Pandoc, using brackets only when the Typst output properties are used here.
- [ ] Remove temporary patch after fix is merged upstream and we are using that version of Pandoc.

younes-io:
Fixes #11093

### DIFF
--- a/src/resources/filters/quarto-post/typst.lua
+++ b/src/resources/filters/quarto-post/typst.lua
@@ -122,13 +122,7 @@ function render_typst_fixups()
     Table = function(tbl)
       -- https://github.com/quarto-dev/quarto-cli/issues/10438
       tbl.classes:insert("typst:no-figure")
-      
-      -- Create Div wrapper and return false to prevent processing its contents
-      return pandoc.Div({
-        pandoc.RawBlock("typst", "#["),
-        tbl,
-        pandoc.RawBlock("typst", "]")
-      }), false
+      return tbl
     end,
     Para = function(para)
       if #para.content ~= 1 then
--- a/src/resources/filters/quarto-post/typst.lua
+++ b/src/resources/filters/quarto-post/typst.lua
@@ -122,7 +122,13 @@ function render_typst_fixups()
     Table = function(tbl)
       -- https://github.com/quarto-dev/quarto-cli/issues/10438
       tbl.classes:insert("typst:no-figure")
-      return tbl
+      
+      -- Create Div wrapper and return false to prevent processing its contents
+      return pandoc.Div({
+        pandoc.RawBlock("typst", "#["),
+        tbl,
+        pandoc.RawBlock("typst", "]")
+      }), false
     end,
     Para = function(para)
       if #para.content ~= 1 then
--- a/src/resources/filters/quarto-post/typst-css-property-processing.lua
+++ b/src/resources/filters/quarto-post/typst-css-property-processing.lua
@@ -284,14 +284,17 @@ function render_typst_css_property_processing()
     Table = function(tab)
       _warnings = new_table()
       local tabstyle = tab.attributes['style']
+      local has_typst_text = false
       if tabstyle ~= nil then
         for clause in tabstyle:gmatch('([^;]+)') do
           local k, v = to_kv(clause)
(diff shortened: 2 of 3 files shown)


gordonwoodhull:
Thanks for your contribution, @younes-io!

As you can see, this fix causes infinite recursion and many of the tests time out. I think this is because it substitutes the table element with a Div containing that element, then recurses in and again tries to convert the table.

I can reproduce this locally by checking out your branch and then attempting to e.g.

```
quarto render tests/docs/smoke-all/typst/gt-islands.qmd`
```

If this is the correct root cause, you could try testing if the table already has the class `typst:no-figure` and not do the substitution in that case. (There may be a more elegant solution, but this is what comes quickly to mind.)

younes-io:
image

image

src/resources/filters/quarto-post/typst.lua
```diff
@@ -120,9 +120,20 @@ function render_typst_fixups()
       return div
     end,
     Table = function(tbl)
-      -- https://github.com/quarto-dev/quarto-cli/issues/10438
+      -- Check if table has already been processed
+      if tbl.classes:includes("typst-processed") then
+        return tbl
```
cscheid, line 125: I believe this is still going to fail. You need to stop the topdown traversal with a `return tbl, false` statement. Refer to the Pandoc docs for more: https://pandoc.org/lua-filters.html#topdown-traversal

cscheid:
> If this is the correct root cause, you could try testing if the table already has the class `typst:no-figure` and not do the substitution in that case. (There may be a more elegant solution, but this is what comes quickly to mind.)

FWIW, I think this still doesn't work on topdown traversals.

cscheid (commented):
With the `return foo, false` pattern, we don't need the `typst-processed` class, right?

gordonwoodhull:
I've squashed these commits and moved the fix the specific place and condition where it is needed, and added a test

As noted in #11093, this is a temporary change until we get the fix into Pandoc, and a couple of tests were failing due to unneeded changes to the Typst output. I'd rather avoid the churn of changing the tests twice.

Let's see if this passes now!

gordonwoodhull:
Merging. Thanks @younes-io!

younes-io:
Thank you @gordonwoodhull for your help & the fix!

BTW, I was trying to set up a `.devcontainer` to make sure I could run tests in a clean environment, but ran into lots of errors during setup... After a couple hours, I had to put it aside. I might pick it up another day! Was wondering if you could share any docs or links (beyond what's in the README) that would help set up a dev environment quickly - without all the hassle of debugging dependencies and version requirements.

Thank you again! And thanks to @cscheid for being so patient as well! 😄

PR #11187, opened 2024-10-24
Better Handle new chrome

cderv:
## New headless mode in chrome

Chrome has new headless mode (https://developer.chrome.com/docs/chromium/headless) activated as default since v128. This PR set back to `--headless=old` to revert back to old mode we know is working ok. There is a new `QUARTO_CHROMIUM_HEADLESS_MODE` env var to configure this. This is to fix #11135. 

Tested with 
````markdown
---
title: "Test"
format: typst
---


```{mermaid}
flowchart LR
  A[Hard edge] --> B(Round edge)
  B --> C{Decision}
  C --> D[Result one]
  C --> E[Result two]
```
````

This now renders ok under this PR. 

If `QUARTO_CHROMIUM_HEADLESS_MODE=new` is set and a chrome browser is already opened. Then #11135 will reproduce. close any browser open and keep `QUARTO_CHROMIUM_HEADLESS_MODE=new`, and issue disappear. So definitely some adaptations to make for new headless mode. As long as we can use, `old` then we can wait. 

## Avoid blocking problem by allowing selecting the chromium binary to use

This PR introduce also a new `QUARTO_CHROMIUM` to specify which binary to use and not let Quarto detection applies. closes #10170 (which closes #2980 and closes #4648) 

Currently I am testing our bundled `quarto install chromium` version to see if this still work and if it is ok with this change. This is quite an old pinned version. 

Also, I noticed on windows that chrome opened for screenshot does not close... So investigating that too.

PR #11195, opened 2024-10-25
ensure that a chromium process started by Quarto is always closed

cderv:
Follow up on #11135 and #11187 as they surfaced a problem where Chromium processed where not closed after rendering. 

The real issue is there 
https://github.com/quarto-dev/quarto-cli/blob/c8a3e7c046884f8cb4bac8d96d73204baaa6328a/src/core/cri/cri.ts#L129-L130

where somehow `await client.close()` does not return and so `browser.close()` never happens. This was confirmed by debugging interactively in VSCODE on Windows. 

**This PR is not solving the issue which is tracked at https://github.com/quarto-dev/quarto-cli/issues/11196** - it is only improving current logic for 1.6 releases to make sure we do clean after ourselves in all cases, like we do with other processed. 

So for that aim, this PR adds `browser` Deno process to the same cleanupHandler for processes for quarto `exitWithCleanup()` logic. This cleanupHandler has all processes started with `execProcess()` already, but we can't use `execProcess()` here as we want to start the process in background, do HTML printing, then close.


cscheid (approved):
LGTM!

PR #11263, opened 2024-10-31
tests environment - pin pipenv

cderv:
it seems new version has broke our tests suites so lets try back the previous one

The new 2024.3.1 was released 15hours ago, which match our workflow failing


cderv:
@cscheid this is fixing stuff and all tests are passing.

PR #11280, opened 2024-11-01
typst css: div color, background-color; translation of named brand colors

gordonwoodhull:
This allows

```markdown
:::{style="background-color: var(--brand-dark-grey); color: var(--brand-blue)"}
This div is blue on dark grey.
:::
```

producing

```typst
#block(fill: brand-color.dark-grey)[
#set text(fill: brand-color.blue); This div is blue on dark grey.

]
```

or produces a warning if the named brand color does not exist.

Parsing variable references isn't worse than parsing rgb, right?

cc @cwickham

### DIFF
--- a/src/resources/filters/modules/brand/brand.lua
+++ b/src/resources/filters/modules/brand/brand.lua
@@ -38,6 +38,7 @@ local function get_typography(fontName)
   local typography = brand.typography and brand.typography[fontName]
   if not typography then return nil end
   local typsted = {}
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
Odd, this test

```
tests/docs/smoke-all/2024/10/28/brand-typography-family-string
```

 from Monday a97fcc04173aabb75 was just failing today, `get_typography` in `brand.lua` needs to upgrade family string to typography object.

PR #11337, opened 2024-11-07
Add QUARTO_KNITR_RSCRIPT_ARGS to pass some flags to RScript for knitr engine

cderv:
e.g `--max-connections=258` available in R 4.4, or `--vanilla`

Args should be passed a comma separated list of flags.

This is for expert use and there is not check done on the argument being support by RScript or not.


cscheid (commented):
Setting environment variables like this is risky - environment variables are global options and we can have tests running in parallel. In that case, the environment variables might end up being changed for tests other than this one.

cderv:
So let's just remove those tests then. 

I didn't find a great way to tweak environment variable for a single test only.

cderv:
They could be added back as part of tests that we run with bundled versions. We know we need to run some already. 

Or something I did not try... leveraging Quarto project and `_environment` file ! I'll try that maybe before we merge this.

PR #11445, opened 2024-11-14
implement brand yaml logo for dashboard, revealjs

gordonwoodhull:
Allowing base logo override with path or resource name.

Also allowing logo object with `path` containing one of those and optional `alt` text.

First row of this ff matrix:

image


gordonwoodhull:
Does not implement dark mode, but would be easy to do, I just haven't looked.

gordonwoodhull:
Replaced by #11446

PR #11639, opened 2024-12-09
move fix to #10567 into website postprocessor

### LINKED ISSUE #11596: New title canonicalization for h1 is breaking custom rendering (using custom template and no boostrap)
New title canonicalization from 
- https://github.com/quarto-dev/quarto-cli/pull/11224

is a bit to eager for custom rendering lik **pkgdown** is doing. This has been found while investigating 
- https://github.com/r-lib/pkgdown/issues/2823

This example recreates what **pkgdown** is doing

* Take this custom template `template.html`
  ````html
  <!DOCTYPE html>
  <html lang="$lang$">
  <body>
  
  <div class="meta">
  $if(title)$
    <h1>$title$</h1>
  $endif$
  </div>
  
  <main>
  $body$
  </main>
  
  </body>
  </html>
  ````
  
  It does not use any of quarto partial and fully custom. **pkgdown** use something like this by taking the `$title$` to put it in a specific div it can retrieve from

* Now use it a simple .qmd document like this 
  ````markdown
  ---
  title: TESTING
  format: 
    html:
      template: template.html
      minimal: true
      theme: none
  ---
  
  # Heading
  ````

The output will be the following after `quarto render`

````html
<!DOCTYPE html>
<html lang="en"><head></head><body>

<div class="meta">
  
</div>

<main><header id="title-block-header" class="quarto-title-block"><h1 class="title">TESTING</h1></header>
<h1 id="heading">Heading</h1>
</main>



</body></html>
````

Previous with Quarto 1.5 
````html
<!DOCTYPE html>
<html lang="en"><head></head><body>

<div class="meta">
  <h1>TESTING</h1>
</div>

<main>
<h1 id="heading">Heading</h1>
</main>



</body></html>
````

New addition from https://github.com/quarto-dev/quarto-cli/pull/11224 is applying here because all the checks are triggered 
https://github.com/quarto-dev/quarto-cli/blob/39baf4081f184fe090974295b48d1beab32cd1e4/src/format/html/format-html-title.ts#L195-L241

https://github.com/quarto-dev/quarto-cli/pull/11224 was supposed to solve #10567 which seemed to be quite specific to [breadcrumbs](https://quarto.org/docs/websites/website-navigation.html#breadcrumbs). So maybe the post processing applies to eagerly here. 

- Should it apply on Project type website rendering only ? 
- Should it apply only when no custom template as been passed ? 

I see that this processing checks that `.quarto-title-block` exists and add it otherwise. I believe class is not always added in HTML document. 
It is from this specific partial (and others) added only for bootstrap output
https://github.com/quarto-dev/quarto-cli/blob/39baf4081f184fe090974295b48d1beab32cd1e4/src/format/html/format-html-title.ts#L19

I mean when `minimal: true` or `theme: pandoc` or `theme: none`, I don't think there will be this class however the processing will add it. 

There are many variation of `title block` also and this check 
https://github.com/quarto-dev/quarto-cli/blob/39baf4081f184fe090974295b48d1beab32cd1e4/src/format/html/format-html-title.ts#L198

cderv:
Fixes #11596

This PR does the `canonicalizeTitlePostprocessor()` into website post processor only which avoid any undesired DOM modification on non website project. 

I think we should look into `canonicalizeTitlePostprocessor()` because it does think for website without checking some assumption

- it fixes breadcrumbs that only are done in bootstrap format. So possibly the fix should be done in `websiteNavigationExtras()` which applies only in bootstrap
- It does try to create a title block and move the title into it, but it does not really do it considering the check on the title. Also it used classes (like `quarto-title-banner`) that are only used in some specific situation (in the example when `title-block-banner` is set). 

I'll suggest we look in all that more closely for 1.7 and adapt the fix for #10567 in a more generic way (make h1 and no title equivalent to title).


cderv:
Changelog done in 1.6 for regression fix through
- https://github.com/quarto-dev/quarto-cli/pull/11640

The change for the regression have been simplified

cderv:
Let's merge so that **pkgdown** problem can be solved

PR #11676, opened 2024-12-13
typst column layout: interpret unitless image widths as pixels and convert to inches

gordonwoodhull:
Follow-up to #11658.

Currently, the following example will fail because it generates Typst output with unitless lengths:

````
---
title: Unitless image widths in column layout
format: typst
---

::: {.callout-note}

## Plots

::: {layout-ncol=5}

![]({{< placeholder format=svg >}})

![]({{< placeholder format=svg >}})

![]({{< placeholder format=svg >}}){width=100}

![]({{< placeholder format=svg >}})

![]({{< placeholder format=svg >}}){width=25}

:::
:::
````

For the images, we already interpret unitless image widths as pixels and convert to inches:

https://github.com/quarto-dev/quarto-cli/blob/46a7609975b14f5d983fd791d7e8e3d425c522ad/src/resources/filters/quarto-post/typst.lua#L101-L110

But we also need to use the converted widths for the layout.

Alternatively / in the the long run, we could contribute the fix to Pandoc instead.

https://github.com/jgm/pandoc/issues/9945


gordonwoodhull:
Sure. LaTeX works mostly the same - looks like it needs some margin/padding on its images, and it will not accept SVG placeholders but one can recognize that LaTeX output

image

is the same intent as Typst output

image

`fig-width` is new to me, will check.

cscheid:
Great, thanks. I don't think we need to (or can) do a perfect job here because units are also not a format-agnostic concept, but it will help us to know a bit more of the current behavior.

PR #11732, opened 2024-12-23
fix: change code block syntax as no longer rendered properly

mcanouil:
For some reason, the code block is no longer rendered properly in issue template.
This PR switches to use the indented code block syntax.

| Before | After |
|--------|--------|
| image | image |


cderv (approved):
Thanks

PR #11760, opened 2024-12-30
fix: Markdown support in listings templates

### LINKED ISSUE #11758: Markdown is no longer supported in listings (>= 1.7.5)
### Bug description

The addition of `{=html}` (to fix https://github.com/quarto-dev/quarto-cli/issues/11701) had an unfortunate and bad side effect which is that none of the field used in the default listings templates support markdown anymore.

https://github.com/quarto-dev/quarto-cli/commit/5d6c846bffd683cac417ae6aa9dc783eb07fb043

### Steps to reproduce

Using a simple website from `quarto create project blog`.
Modifying one of the post with:

````qmd
---
title: "Welcome To My Blog with `code`"
description: |
  This is a text with **bold** and *italic* formatting.
author: "Tristan O'Malley"
date: "2024-12-27"
categories: [news]
---

This is the first post in a Quarto blog. Welcome!

![](thumbnail.jpg)

Since this post doesn't specify an explicit `image`, the first image in the post will be used in the listing page of posts.
````

### Expected behavior

Any version up to 1.7.4

Image

### Actual behavior

From 1.7.5

Image

### Your environment

_No response_

### Quarto check output

```txt
Quarto 99.9.9
      Pandoc version 3.4.0: OK
      Dart Sass version 1.70.0: OK
      Deno version 1.46.3: OK
      Typst version 0.11.0: OK
      Version: 99.9.9
      commit: 241155a28bbfe8ffb5e9fd53f7396b516a7a8deb
      Path: /Users/mcanouil/Projects/quarto/quarto-cli/package/dist/bin

      TinyTeX: v2024.11
      Chromium: (not installed)

      Using: TinyTex
      Version: 2024

      Version: 3.13.1
      Path: /Users/mcanouil/Projects/quarto/quarto-playground/.venv/bin/python3
      Jupyter: 5.7.2
      Kernels: julia-1.11, python3

(|) Checking R installation...........ℹ R version 4.4.2 (2024-10-31)
! Config '~/.Rprofile' was loaded!
      Version: 4.4.2
      LibPaths:
      knitr: 1.46
      rmarkdown: 2.26

```

mcanouil:
The changes restore Markdown support in the default listings templates after the introduction of `{=html}` in version 1.7.5. This update ensures that Markdown formatting is correctly rendered in listings.

Fixes #11758.

This PR also adds a check for `subtitle` field in the default listing. Previously it was adding subtitle even when not provided as long as `title` was.

Tested on: https://m.canouil.dev/quarto-issues-experiments/issue11758/
Source: https://github.com/mcanouil/quarto-issues-experiments/tree/main/quarto-cli-11758

To do:

- [x] check that the fix for #11701 still works.

### DIFF
--- a/src/resources/projects/website/listing/_filter.ejs.md
+++ b/src/resources/projects/website/listing/_filter.ejs.md
@@ -23,7 +23,7 @@ const sortUi = listing['sort-ui'];
       <option
         value="<%- sortData.listingSort.field %>"
         data-direction="<%- sortData.listingSort.direction %>">
-        <%- sortData.description %>
+        <%= sortData.description %>
       </option>
       <% } %>
     </select>
--- a/src/resources/projects/website/listing/_filter.ejs.md
+++ b/src/resources/projects/website/listing/_filter.ejs.md
@@ -10,32 +10,29 @@ const sortUi = listing['sort-ui'];
   <% const sortableFields = listing.utilities.sortableFieldData(); %>
   <% if (sortUi && sortableFields.length > 0) { %>
    <div class="input-group input-group-sm quarto-listing-sort">
-     <span class="input-group-text"><i class="bi bi-sort-down"></i></span>
-     <select
+    <span class="input-group-text"><i class="bi bi-sort-down"></i></span>
+    <select
       id="listing-<%- listing.id %>-sort"
       class="form-select"
-      aria-label="<%- listing.utilities.localizedString("listing-page-order-by")%>"
+      aria-label="<%- listing.utilities.localizedString("listing-page-order-by") %>"
       onChange="window['quarto-listings']['listing-<%- listing.id %>'].sort(this.options[this.selectedIndex].value, { order: this.options[this.selectedIndex].getAttribute('data-direction')})"
     >
-       <option value="" disabled selected hidden><%- listing.utilities.localizedString("listing-page-order-by")%></option>
-       <option value="index" data-direction="asc"><%- listing.utilities.localizedString("listing-page-order-by-default")%></option>
-       <% for (const sortData of sortableFields) { %>
-         <option
-          value="<%- sortData.listingSort.field %>"
-          data-direction="<%- sortData.listingSort.direction %>">
-          <%= sortData.description %>
-        </option>
-       <% } %>
-     </select>
+      <option value="" disabled selected hidden><%- listing.utilities.localizedString("listing-page-order-by") %></option>
+      <option value="index" data-direction="asc"><%- listing.utilities.localizedString("listing-page-order-by-default") %></option>
+      <% for (const sortData of sortableFields) { %>
+      <option
+        value="<%- sortData.listingSort.field %>"
+        data-direction="<%- sortData.listingSort.direction %>">
+        <%- sortData.description %>
+      </option>
+      <% } %>
+    </select>
   </div>
-
-
   <% } %>
-
   <% if (filterUi) { %>
     <div class="input-group input-group-sm quarto-listing-filter">
       <span class="input-group-text"><i class="bi bi-search"></i></span>
-      <input type="text" class="search form-control" placeholder="<%- listing.utilities.localizedString("listing-page-filter")%>" />
+      <input type="text" class="search form-control" placeholder="<%- listing.utilities.localizedString("listing-page-filter") %>" />
     </div>
   <% } %>
 </div>
--- a/src/resources/projects/website/listing/_filter.ejs.md
+++ b/src/resources/projects/website/listing/_filter.ejs.md
@@ -14,11 +14,11 @@ const sortUi = listing['sort-ui'];
     <select
       id="listing-<%- listing.id %>-sort"
       class="form-select"
-      aria-label="<%- listing.utilities.localizedString("listing-page-order-by") %>"
+      aria-label="<%= listing.utilities.localizedString("listing-page-order-by") %>"
       onChange="window['quarto-listings']['listing-<%- listing.id %>'].sort(this.options[this.selectedIndex].value, { order: this.options[this.selectedIndex].getAttribute('data-direction')})"
     >
-      <option value="" disabled selected hidden><%- listing.utilities.localizedString("listing-page-order-by") %></option>
-      <option value="index" data-direction="asc"><%- listing.utilities.localizedString("listing-page-order-by-default") %></option>
+      <option value="" disabled selected hidden><%= listing.utilities.localizedString("listing-page-order-by") %></option>
+      <option value="index" data-direction="asc"><%= listing.utilities.localizedString("listing-page-order-by-default") %></option>
       <% for (const sortData of sortableFields) { %>
       <option
         value="<%- sortData.listingSort.field %>"
@@ -32,7 +32,7 @@ const sortUi = listing['sort-ui'];
   <% if (filterUi) { %>
     <div class="input-group input-group-sm quarto-listing-filter">
       <span class="input-group-text"><i class="bi bi-search"></i></span>
-      <input type="text" class="search form-control" placeholder="<%- listing.utilities.localizedString("listing-page-filter") %>" />
+      <input type="text" class="search form-control" placeholder="<%= listing.utilities.localizedString("listing-page-filter") %>" />
     </div>
   <% } %>
 </div>
--- a/src/resources/projects/website/listing/_pagination.ejs.md
+++ b/src/resources/projects/website/listing/_pagination.ejs.md
@@ -1,8 +1,5 @@
-::: {.listing-no-matching .d-none}
-<%- listing.utilities.localizedString("listing-page-no-matches") %>
-:::
-
 ```{=html}
+<div class="listing-no-matching d-none"><%= listing.utilities.localizedString("listing-page-no-matches") %></div>
 <% if (listing["page-size"] < items.length) { %>
 <nav id="<%- listing.id %>-pagination" class="listing-pagination" aria-label="Page Navigation">
   <ul class="pagination"></ul>
--- a/src/resources/projects/website/listing/_pagination.ejs.md
+++ b/src/resources/projects/website/listing/_pagination.ejs.md
@@ -2,12 +2,10 @@
 <%- listing.utilities.localizedString("listing-page-no-matches") %>
 :::
 
-<% if (listing["page-size"] < items.length) { %>
-
 ```{=html}
+<% if (listing["page-size"] < items.length) { %>
 <nav id="<%- listing.id %>-pagination" class="listing-pagination" aria-label="Page Navigation">
   <ul class="pagination"></ul>
 </nav>
-```
-
 <% } %>
+```
--- a/src/resources/projects/website/listing/_pagination.ejs.md
+++ b/src/resources/projects/website/listing/_pagination.ejs.md
@@ -1,10 +1,13 @@
-```{=html}
-<div class="listing-no-matching d-none">
+::: {.listing-no-matching .d-none}
 <%- listing.utilities.localizedString("listing-page-no-matches") %>
-</div>
+:::
+
 <% if (listing["page-size"] < items.length) { %>
+
+```{=html}
 <nav id="<%- listing.id %>-pagination" class="listing-pagination" aria-label="Page Navigation">
   <ul class="pagination"></ul>
 </nav>
-<% } %>
 ```
+
+<% } %>
--- a/src/resources/projects/website/listing/_pagination.ejs.md
+++ b/src/resources/projects/website/listing/_pagination.ejs.md
@@ -1,8 +1,10 @@
-<div class="listing-no-matching d-none"><%- listing.utilities.localizedString("listing-page-no-matches")%></div>
-<% if (listing["page-size"] < items.length) { %>
 ```{=html}
+<div class="listing-no-matching d-none">
+<%- listing.utilities.localizedString("listing-page-no-matches") %>
+</div>
+<% if (listing["page-size"] < items.length) { %>
 <nav id="<%- listing.id %>-pagination" class="listing-pagination" aria-label="Page Navigation">
   <ul class="pagination"></ul>
 </nav>
-```
 <% } %>
+```
--- a/src/resources/projects/website/listing/item-default.ejs.md
+++ b/src/resources/projects/website/listing/item-default.ejs.md
@@ -31,44 +31,64 @@ return value;
 
 let value = readField(item, field);
 if (value !== undefined) {
-print(`<div class="metadata-value listing-${field}">${listing.utilities.outputLink(item, field, value)}</div>`);  
+print(`<div class="metadata-value listing-${field}">${listing.utilities.outputLink(item, field, value)}</div>`);
  }
 }
 %>
 
-<div class="quarto-post image-<%= imageAlign %>" <%= listing.utilities.metadataAttrs(item) %>>
+::: {.quarto-post .image-<%= imageAlign %> <%= listing.utilities.metadataAttrs(item) %>}
+
 <% if (fields.includes('image')) { %>
-<div class="thumbnail">
-<a href="<%- item.path %>" class="no-external">
+`<div class="thumbnail"><a href="<%- item.path %>" class="no-external">`{=html}
 <% if (item.image) { %>
 <%= listing.utilities.img(itemNumber, item.image, "thumbnail-image", item['image-alt'], item['image-lazy-loading'] ?? listing['image-lazy-loading']) %>
 <% } else { %>
 <%= listing.utilities.imgPlaceholder(listing.id, itemNumber, item.outputHref) %>
 <% } %>
-</a>
-</div>
+`</a></div>`{=html}
 <% } %>
-<div class="body">
-<% if (fields.includes('title')) { %>
-<h3 class="no-anchor listing-title"><a href="<%- item.path %>" class="no-external"><%= item.title %></a></h3>
-<div class="listing-subtitle"><a href="<%- item.path %>" class="no-external"><%= item.subtitle %></a></div>
-<% } %>
-<% if (fields.includes('categories') && item.categories) { %> 
-<div class="listing-categories">
-<% for (const category of item.categories) { %>
-<div class="listing-category" onclick="window.quartoListingCategory('<%=utils.b64encode(category)%>'); return false;"><%= category %></div>
-<% } %>
-</div>
-<% } %> 
-<% if (fields.includes('description')) { %>
-<div class="delink listing-description"><a href="<%- item.path %>" class="no-external"><%= item.description %></a></div>
-<% } %>
-</div>
-<div class="metadata"><a href="<%- item.path %>" class="no-external">
-<% if (fields.includes('date') && item.date) { %><div class="listing-date"><%= item.date %></div><% } %>
-<% if (fields.includes('author') && item.author) { %><div class="listing-author"><%= item.author %></div><% } %>
-<% if (fields.includes('reading-time') && item['reading-time']) { %> <div class="listing-reading-time"><%= item['reading-time'] %></div> <% } %>
-<% for (const field of otherFields) { %>
-<% outputMetadata(item, field) %><% } %></a></div>
 
-</div>
+::: {.body}
+
+<% if (fields.includes('title')) { %>
+`<h3 class="no-anchor listing-title"><a href="<%- item.path %>" class="no-external">`{=html} <%= item.title %> `</a></h3>`{=html}
+<% if (fields.includes('subtitle')) { %>
+`<div class="listing-subtitle"><a href="<%- item.path %>" class="no-external">`{=html} <%= item.subtitle %> `</a></div>`{=html}
+<% } %>
+<% } %>
+
+<% if (fields.includes('categories') && item.categories) { %>
+`<div class="listing-categories">`{=html}
+<% for (const category of item.categories) { %>
+`<div class="listing-category" onclick="window.quartoListingCategory('<%- utils.b64encode(category) %>'); return false;">`{=html} <%= category %> `</div>`{=html}
+<% } %>
+`</div>`{=html}
+<% } %>
+
+<% if (fields.includes('description')) { %>
+`<div class="delink listing-description"><a href="<%- item.path %>" class="no-external">`{=html} <%= item.description %> `</a></div>`{=html}
+<% } %>
+
+:::
+
+::: {.metadata}
+`<a href="<%- item.path %>" class="no-external">`{=html}
+<% if (fields.includes('date') && item.date) { %>
+`<div class="listing-date">`{=html} <%= item.date %> `</div>`{=html}
+<% } %>
+
+<% if (fields.includes('author') && item.author) { %>
+`<div class="listing-author">`{=html} <%= item.author %> `</div>`{=html}
+<% } %>
+
+<% if (fields.includes('reading-time') && item['reading-time']) { %>
+`<div class="listing-reading-time">`{=html} <%= item['reading-time'] %> `</div>`{=html}
+<% } %>
+
+<% for (const field of otherFields) { %>
+<% outputMetadata(item, field) %>
+<% } %>
+`</a>`{=html}
+:::
+
+:::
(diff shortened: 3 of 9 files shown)


src/resources/projects/website/listing/item-default.ejs.md
```diff
@@ -31,44 +31,64 @@ return value;
...
 let value = readField(item, field);
 if (value !== undefined) {
-print(`<div class="metadata-value listing-${field}">${listing.utilities.outputLink(item, field, value)}</div>`);  
+print(`<div class="metadata-value listing-${field}">${listing.utilities.outputLink(item, field, value)}</div>`);
  }
 }
 %>
 
-<div class="quarto-post image-<%= imageAlign %>" <%= listing.utilities.metadataAttrs(item) %>>
+::: {.quarto-post .image-<%= imageAlign %> <%= listing.utilities.metadataAttrs(item) %>}
```
mcanouil, line 39: Switching to div as raw inline/block content leads to parsing issues.

src/resources/projects/website/listing/item-default.ejs.md
```diff
@@ -31,44 +31,64 @@ return value;
...
+<% } %>
 <% } %>
-<% if (fields.includes('categories') && item.categories) { %> 
-<div class="listing-categories">
+
+<% if (fields.includes('categories') && item.categories) { %>
+`<div class="listing-categories">`{=html}
 <% for (const category of item.categories) { %>
-<div class="listing-category" onclick="window.quartoListingCategory('<%=utils.b64encode(category)%>'); return false;"><%= category %></div>
+`<div class="listing-category" onclick="window.quartoListingCategory('<%- utils.b64encode(category) %>'); return false;">`{=html} <%= category %> `</div>`{=html}
```
mcanouil, line 63: I don't think `<%=` should be used here, mostly semantic as base64 should not have any markdown syntax in it.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,101 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
 });
 %>
 
-<div class="g-col-1" <%= listing.utilities.metadataAttrs(item) %>>
+::: {.g-col-1 <%= listing.utilities.metadataAttrs(item) %> }
```
mcanouil, line 40: Same as for the other file.
Using div syntax to avoid parsing issues with the use of. `` ` `` either in raw inline or block.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,101 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
+
+```{=html}
 <a href="<%- item.path %>" class="quarto-grid-link">
-<div class="quarto-grid-item card h-100 <%-`card-${align}`%><%= hideBorders ? ' borderless' : '' %>">
+<div class="quarto-grid-item card h-100 <%- `card-${align}` %><%- hideBorders ? ' borderless' : '' %>">
```
mcanouil, line 44: Switching to `<%-`.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,101 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
 <% } %>
 
-</div>
+`</div>`{=html}
 <% } %>
 
 <% if (fields.includes('filename') || fields.includes('file-modified')) { %>
 
-<div class="card-footer">
+`<div class="card-footer">`{=html}
```
cderv, line 123: Could be a problem here to wrap in raw inline as our Lua filter does check for div of class `card-footer`
https://github.com/quarto-dev/quarto-cli/blob/247513ca1c072567f44132de6aee3c474c5f5afc/src/resources/filters/modules/dashboard/card.lua#L83-L85

Though from quick search and check, it seems we use Lua processing only in `modules/dashboard/*`, so could be safe in listing page.

Just sharing in case it rings any bells to anyone.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,101 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
 <% if (fields.includes('categories') && item.categories) { %>
+`<div class="listing-categories">`{=html}
 
-<div class="listing-categories">
-  <% for (const category of item.categories) { %>
-<div class="listing-category" onclick="window.quartoListingCategory('<%=utils.b64encode(category)%>'); return false;"><%= category %></div>
-  <% } %>
-</div>
+<% for (const category of item.categories) { %>
+`<div class="listing-category" onclick="window.quartoListingCategory('<%- utils.b64encode(category ) %>'); return false;">`{=html} <%= category %> `</div>`{=html}
```
cderv, line 77: Do you think we need to change to `<%-` to escape content ? 
I would probably use `<%` but `%<=` seems to be fine and working right now.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,101 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
 <% } %>
 
 <% } %>
 <% if (showField('title') || showField('subtitle') || showField('description') || showField('author') || showField('date') || otherFields.length > 0) { %>
 
-<div class="card-body post-contents">
-<% if (showField('title')) { %><h5 class="no-anchor card-title listing-title"><%= item.title %></h5><% } %>
-<% if (showField('subtitle')) { %><div class="card-subtitle listing-subtitle"><%= item.subtitle %></div><% } %>
-<% if (showField('reading-time')) { %><div class="listing-reading-time card-text text-muted"><%= item['reading-time'] %></div> <% } %>
+`<div class="card-body post-contents">`{=html}
```
cderv, line 59: `card-body` seems to be checked on divs in Lua but only in `modules/dashboard/*` so probably ok.

cderv (commented):
Thanks a lot !

Using raw inlines for each raw HTML seems the right call if we need to avoid markdown in HTML parsing. 

Only impact I can think of is that `<div>` in markdown leads to `Div` in AST, while using raw inline leads to RawInline. It has impact on Lua filter processing. 

````powershell
❯ pandoc -t native
<div class="thumbnail">**STRONG**</div>
^Z
[ Div
    ( "" , [ "thumbnail" ] , [] )
    [ Plain [ Strong [ Str "STRONG" ] ] ]
]

~ took 12s
❯ pandoc -t native
`<div class="thumbnail">`{=html}**STRONG**`</div>`{=html}
^Z
[ Para
    [ RawInline (Format "html") "<div class=\"thumbnail\">"
    , Strong [ Str "STRONG" ]
    , RawInline (Format "html") "</div>"
    ]
]
````

So we probably need to make sure we have no Lua processing applying on any Div we create in Listing. I check our Lua code using search. 

See more in comment about some example of potential issue - I think we are safe as they seems to apply in Dashboard format only. 

Regarding test, it seems something needs to be adapted as a test seems to be broken. Didn't look carefully yet. 

Thanks again for the PR !

File: src/resources/projects/website/listing/item-grid.ejs.md (hunk 4 above)
Comment (mcanouil) [reply src/resources/projects/website/listing/item-grid.ejs.md:44]: Mostly for clarify as to what is expected.
I'm expecting `<%=` to be further processed while `<%-` taken as is.

File: src/resources/projects/website/listing/item-default.ejs.md (hunk 1 above)
Comment (mcanouil) [reply src/resources/projects/website/listing/item-default.ejs.md:39]: The line is not taken as a line and is broken when using raw inline or block.
It mostly breaks the filter/sort feature which is the most obvious issue.
I'm am not yet sure to what's causing this.

File: src/resources/projects/website/listing/item-default.ejs.md (hunk 2 above)
Comment (mcanouil) [reply src/resources/projects/website/listing/item-default.ejs.md:63]: We need to do more testing as we might fix the following issue which seems related to how categories, etc. are handled.

- #11752

File: src/resources/projects/website/listing/item-grid.ejs.md (hunk 4 above)
Comment (cderv) [reply src/resources/projects/website/listing/item-grid.ejs.md:44]: > I'm expecting <%= to be further processed while <%- taken as is/escaped.

Regarding this, to avoid any confusion, 
* `<%-` escapes
* `<%=` just interpolates

This is some oddity with lodash framework which is used here, as it is reversed with other ejs framework. It was discussed previous in https://github.com/quarto-dev/quarto-cli/issues/7185#issuecomment-1850229793 

Docs for the syntax is at https://lodash.com/docs/4.17.15#template

so when you say `<%-` "as escaped", do you mean it should be escaped in output ?

Just trying to be on the same page, as this was confusing for me first time I worked on those template 😅 

> I need to do some more checks, because I personally don't like the use of that much raw inline/block.
If the div markdown syntax also avoid the memory issue, I believe it would make the EJS more readable/maintainable.µ

👍

File: src/resources/projects/website/listing/item-default.ejs.md (hunk 2 above)
Comment (cderv) [reply src/resources/projects/website/listing/item-default.ejs.md:63]: Oh I missed that. I can have a look as I think I fixed something related, after we switch to base64 encoding for category. 

I would probably do a separate PR and not fix everything in one single PR.

Comment (mcanouil) [reply src/resources/projects/website/listing/item-default.ejs.md:63]: Yes, if it's in the JS script it makes sense to make a separate PR.

File: src/resources/projects/website/listing/_pagination.ejs.md [hunk 8]
```diff
@@ -1,8 +1,10 @@
-<div class="listing-no-matching d-none"><%- listing.utilities.localizedString("listing-page-no-matches")%></div>
-<% if (listing["page-size"] < items.length) { %>
 ```{=html}
+<div class="listing-no-matching d-none">
+<%- listing.utilities.localizedString("listing-page-no-matches") %>
```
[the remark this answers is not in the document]
mcanouil: 0f779cf95845d3cd550d6adaa9b6f03ee12a0da9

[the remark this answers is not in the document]
mcanouil: Looking at other part of the code, I'm not sure we want `listing.utilities.localizedString("listing-page-no-matches")` to be evaluated as Markdown.
For instance, it's not possible for `_filter.ejs.md`.

src/resources/projects/website/listing/item-grid.ejs.md
mcanouil: > so when you say <%- "as escaped", do you mean it should be escaped in output ?

Well, there should be nothing to be escaped at all in the string as classes and attributes won't like non ascii characters.
Let's keep `<%=`.

src/resources/projects/website/listing/item-grid.ejs.md
mcanouil: It actually does not matter has base64 encoding does not produce anything that needs escaping.

mcanouil:
I'm a bit stuck for the remaining raw HTML code.

Using the following without raw code block works as expected for having the item being read/evaluated as markdown.

```html
<h5 class="no-anchor card-title listing-title"><%= item.title %></h5>
```

Using raw block will lead to the item being placed in a `<p>` tag leading to extra styling which is really unwanted inside `h5`.

````md
```{=html}
<h5 class="no-anchor card-title listing-title">
```
<%= item.title %>
```{=html}
</h5>
```
```

This happens for other parts with `<div>` (grid and default listings) and `<td>` (table listing and grid listing for metadata).

File: src/resources/projects/website/listing/listing-table.ejs.md [hunk 9]
```diff
@@ -84,11 +84,13 @@ return listing.utilities.outputLink(item, field, value, `listing-${field}`);
 <tr <%= listing.utilities.metadataAttrs(item) %><%= onclick(item) %>>
 <% for (const field of fields){ %>
 <td>
+```
 <%= outputValue(itemNumber, field) %>
+```{=html}
```
mcanouil, line 89: This leads to extra `<p>` being added as the line is read as Markdown.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,147 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
 <% if (showField('title') || showField('subtitle') || showField('description') || showField('author') || showField('date') || otherFields.length > 0) { %>
 
-<div class="card-body post-contents">
-<% if (showField('title')) { %><h5 class="no-anchor card-title listing-title"><%= item.title %></h5><% } %>
-<% if (showField('subtitle')) { %><div class="card-subtitle listing-subtitle"><%= item.subtitle %></div><% } %>
-<% if (showField('reading-time')) { %><div class="listing-reading-time card-text text-muted"><%= item['reading-time'] %></div> <% } %>
+::: {.card-body .post-contents}
+
+<% if (showField('title')) { %>
+<h5 class="no-anchor card-title listing-title"><%= item.title %></h5>
```
mcanouil, line 71: No raw block to avoid `<p>` being added.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,147 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
+<% } %>
+
+<% if (showField('subtitle')) { %>
+<div class="card-subtitle listing-subtitle"><%= item.subtitle %></div>
```
mcanouil, line 75: No raw block to avoid `<p>` being added.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,147 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
-<div class="card-attribution card-text-small <%-flexJustify%>">
-<% if (showField('author')) { %><div class="listing-author"><%= item.author %></div><% } %>
-<% if (showField('date')) { %><div class="listing-date"><%= item.date %></div><% } %>
+
+```{=html}
+<div class="card-attribution card-text-small <%- flexJustify %>">
+```
+
+<% if (showField('author')) { %>
+<div class="listing-author"><%= item.author %></div>
```
mcanouil, line 123: No raw block to avoid `<p>` being added.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,147 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
+<% } %>
+
+<% if (showField('date')) { %>
+<div class="listing-date"><%= item.date %></div>
```
mcanouil, line 127: No raw block to avoid `<p>` being added.

src/resources/projects/website/listing/item-default.ejs.md
```diff
@@ -31,44 +31,94 @@ return value;
...
+
 <% } %>
-<div class="body">
+
+::: {.body}
+
 <% if (fields.includes('title')) { %>
 <h3 class="no-anchor listing-title"><a href="<%- item.path %>" class="no-external"><%= item.title %></a></h3>
+<% if (fields.includes('subtitle')) { %>
 <div class="listing-subtitle"><a href="<%- item.path %>" class="no-external"><%= item.subtitle %></a></div>
```
mcanouil, line 60: No raw block to avoid `<p>` being added.

src/resources/projects/website/listing/item-default.ejs.md
```diff
@@ -31,44 +31,94 @@ return value;
...
+<a href="<%- item.path %>" class="no-external">
+```
+
+<% if (fields.includes('date') && item.date) { %>
+<div class="listing-date"><%= item.date %></div>
+<% } %>
+
+<% if (fields.includes('author') && item.author) { %>
+<div class="listing-author"><%= item.author %></div>
+<% } %>
```
mcanouil, line 104: No raw block to avoid `<p>` being added.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,147 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
...
+let value = readField(item, field);
 %>
 <tr>
 <td><%= listing.utilities.fieldName(field) %></td>
-<td class="<%-field%>"><%= listing.utilities.outputLink(item, field, value) %></td>
+<td class="<%- field %>"><%= listing.utilities.outputLink(item, field, value) %></td>
 </tr>
 <% } %>
 </table>
+```
```
mcanouil, line 149: No markdown supports in metadata with this.
Splitting this to allow markdown for `listing.utilities.fieldName(field)` and `listing.utilities.outputLink(item, field, value)` will lead to `<p>` being added.

mcanouil:
In the current state, I have a constant/mitigate memory consumption when rendering the project from #11701

src/resources/projects/website/listing/item-grid.ejs.md
mcanouil: For consistency until there is a decision, I switched to using raw blocks except for the lines with the variables to allow markdown.

9c43d630bc7a34f3dcf7f7465380420b2890d030

This means, there is still inline HTML.

src/resources/projects/website/listing/listing-table.ejs.md
mcanouil: refactor a bit by making `<td>` outside the raw block.

9c43d630bc7a34f3dcf7f7465380420b2890d030

src/resources/projects/website/listing/item-default.ejs.md
cderv: New PR is up for this.

src/resources/projects/website/listing/item-grid.ejs.md
```diff
@@ -37,83 +37,151 @@ return !["title", "image", "image-alt", "date", "author", "subtitle", "descripti
 });
 %>
 
-<div class="g-col-1" <%= listing.utilities.metadataAttrs(item) %>>
+::: {.g-col-1 <%= listing.utilities.metadataAttrs(item) %> }
```
cscheid, line 40: There's a subtle change in behavior here. `metadataAttrs` was written with the HTML attribute syntax in mind, and now we're using it inside a fenced div.

These are not equivalent because of `data-{ATTR}` vs {ATTR}`. Consider:

````
$ pandoc -f html -t native
<div data-foo="bar">Ok</div>
[ Div
    ( "" , [] , [ ( "foo" , "bar" ) ] ) [ Plain [ Str "Ok" ] ]
]
$ pandoc -f markdown -t native
::: {data-foo="bar"}
Ok
:::
[ Div
    ( "" , [] , [ ( "data-foo" , "bar" ) ] )
    [ Para [ Str "Ok" ] ]
]
````

Note the difference in the attribute name in the result. Now, the bug might be harmless, because Pandoc "back-converts" attributes named data-ATTR to ATTR when writing to HTML:

````
$ pandoc -f markdown -t html
::: {data-foo="bar"}
Hello
:::

::: {foo="bar"}
Hello
:::
^D
<div data-foo="bar">
<p>Hello</p>
</div>
<div data-foo="bar">
<p>Hello</p>
</div>
````

The change in behavior is that if we have Lua filters handling the listing document, the attributes generated by metadataAttrs will have changed from `foo` to `data-foo`. I'm not sure this is a problem, but I wanted to make sure this is documented somewhere.

File: tests/docs/listings/_quarto.yml [hunk 18]
```diff
@@ -5,4 +5,4 @@ website:
   title: "Listing Test"
 format:
   html:
-    theme: [litera, styles.scss]
\ No newline at end of file
+    theme: [litera]
```
cscheid, line 8: I don't think this makes a difference either way, but why have this change here?

cscheid (approved):
This looks great!

I wanted to leave one comment for future reference because of a subtle change in behavior, but I think it's ready to merge.

src/resources/projects/website/listing/item-grid.ejs.md
mcanouil: As currently some HTML is still use inline, we could do that here as well.
It will require to test again rendering on the repository from <https://github.com/quarto-dev/quarto-cli/issues/11701> to be sure reverting this to inline HTML does not lead to the memory issue by Pandoc.

tests/docs/listings/_quarto.yml
mcanouil: To remove all the warnings from the log because the file was not found.
It makes it harder to focus on what matters when testing this.

src/resources/projects/website/listing/item-grid.ejs.md
cscheid: I really think no further action is required. By "I wanted to make sure this is documented somewhere", I meant "I want to leave this comment here so that we can find it ourselves in the future".

mcanouil: Ok, feel free to merge when you want 😉

src/resources/projects/website/listing/_filter.ejs.md
cderv, line 1: `localizedString()` usage here will insert a user provided value. I think it is safer to keep using `<%-` here, especially when used in HTML attribute, as we can't know which character would be used. 

Maybe there won't be any `<` used for `listing-page-order-by` but if so, then I think it could break HTML. 

That is my thinking. If there is a specific reason, I am missing it, so I would really like to understand and learn something new. Why replacing ? 

My first though is that using `%<-` for escaping is safer. 

Thank you

src/resources/projects/website/listing/_pagination.ejs.md
```diff
@@ -1,8 +1,8 @@
-<div class="listing-no-matching d-none"><%- listing.utilities.localizedString("listing-page-no-matches")%></div>
-<% if (listing["page-size"] < items.length) { %>
 ```{=html}
+<div class="listing-no-matching d-none"><%= listing.utilities.localizedString("listing-page-no-matches") %></div>
```
cderv, line 2: Same here: `localizedString()` is for user input value, and we don't know for sure that no special char won't be used. Isn't it safer to keep `<%-` ? 🤔 
is this to support HTML content for `listing-page-no-matches` ?

cderv (commented):
This is to improve my knowledge and remove some concerns due to interrogation I have. I prefer to be sure we avoid possible regression. 

Those EJS template where using `<%-` and not `<%=` in some case to be sure that content is encoded, when used in attributes for example. 

Maybe there was issue to fix and we wrongly encoded some of them, though I am not sure all replacement of `<%-` are justified below, specifically those with `localizedString()`. 

I may be missing something though, so asking for precision again to fully understand those EJS template. 

Thanks for your patience :)

src/resources/projects/website/listing/_filter.ejs.md
mcanouil: As all the occurrences are in an HTML raw block, I think we can go (back) with `<%-` as no markdown will ever work there.

Should I do the PR?

cderv: To keep on the purpose of learning, as this is still unclear, can you explain to me 

> no markdown will ever work there.

Or rather you'll confirm if I understand correctly

- This ejs template is in fact a markdown file
- So we should be carefull when not in a raw block because Markdown in HTML will be interpreted (which is the initial problem)
- And in this case, escaping could be harmful because it could break the markdown syntax. 
- And so what I shared above is really valid only inside raw block, or inside a template file that is not interpreted as .md file. 

I think this is what I was missing, right ?

mcanouil: We should probably update the documentation about all this story:
- using raw block (Pandoc memory issue and possible parsing issue that could happen as without Pandoc has to guess it's HTML)
- EJS templates in Quarto are read as Mardown files

https://quarto.org/docs/websites/website-listings-custom.html

mcanouil: I went ahead and made a PR:

- https://github.com/quarto-dev/quarto-cli/pull/11817

mcanouil: Also opened a documentation issue:

- https://github.com/quarto-dev/quarto-cli/issues/11818

PR #11875, opened 2025-01-16
try to add duplicate warning definition only for knitr

cderv:
In knitr, `warning: 'NA'` is supported. How to allow that for auto completion ? 

````markdown
---
title: "test"
format: html
---

```{r}
#| warning: NA
warning("Test")
```
````

### DIFF
--- a/src/resources/editor/tools/vs-code.mjs
+++ b/src/resources/editor/tools/vs-code.mjs
@@ -8216,6 +8216,24 @@ var require_yaml_intelligence_resources = __commonJS({
           }
         },
         {
+          name: "warning",
+          tags: {
+            engine: "knitr"
+          },
+          schema: {
+            enum: [
+              true,
+              false,
+              "NA"
+            ]
+          },
+          default: true,
+          description: {
+            short: "Include warning in rendered output.",
+            long: "Include warnings in rendered output. Possible values are `true`, `false`, or `NA`. \nIf `true`, messages are included in the output. If `false`, messages are not included. \nIf `NA`, messages are not included in output but shown in the knitr log to console.\n"
+          }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by cderv** [comment]
@cscheid Do you know if our schema would allow to get a specific YAML validation for an option in cell 
image

while this option is also used as `document-execute`. 

The way I tried here, seems to have remove the execute part
image
while correctly added under cell 
image

Normaly warning is also in execute
image


So do we have a way to have two `warning` definition with our schema ? 

Thank you !

**Comment by cderv** [comment]
A note that the document in example
````
---
title: "test"
format: html
---

```{r}
#| warning: NA
warning("Test")
```
````

Is not supposed to be rendered and should have a validation error at rendering time. 

But this is not happening because of 
https://github.com/quarto-dev/quarto-cli/blob/46d71b0fd440dbae2a0d044c95f52622724a3939/src/core/schema/validate-document.ts#L113-L116

End line problem 😁 `\r\n` on Windows, so `start` should be `+ 7` not 6.

Fixing this will probably make a lot of document fails, at least those using valid **knitr** option but unknown to validation. We'll see. 

I'll probably open a new PR to solve this issue.

cderv:
#12151 will fix the YAML validation, but this means that `warning: NA` won't work in .qmd anymore, until this issue of scoped definition is solved.

cderv:
For reference, similar problem with `echo` and `eval`. Our schema focument more than just boolean, but we don't only it
https://github.com/quarto-dev/quarto-cli/blob/e8eefd98f816622a13aba818d404b98c00c8cfb6/src/resources/schema/cell-codeoutput.yml#L17-L35

PR #11928, opened 2025-01-22
Bump QuartoNotebookRunner to 0.12.0

MichaelHatherly:
Upgrades to the latest QNR.jl version. This version now supports running different versions of Julia in individual notebooks. Previously we required all the Julia versions to match the version used by the Julia server process.

### DIFF
--- a/.github/workflows/test-smokes.yml
+++ b/.github/workflows/test-smokes.yml
@@ -170,7 +170,7 @@ jobs:
       - name: Setup Julia
         uses: julia-actions/setup-julia@v2
         with:
-          version: "1.11"
+          version: "1.11.3"
 
       - name: Cache Julia Packages
         uses: julia-actions/cache@v2
--- a/src/resources/julia/Project.toml
+++ b/src/resources/julia/Project.toml
@@ -2,4 +2,4 @@
 QuartoNotebookRunner = "4c0109c6-14e9-4c88-93f0-2b974d3468f4"
 
 [compat]
-QuartoNotebookRunner = "=0.11.6"
+QuartoNotebookRunner = "=0.12.0"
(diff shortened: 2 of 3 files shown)


MichaelHatherly:
Looks like some errors related to running Julia 1.11.3 with a Manifest.toml that is resolved for Julia 1.11.2. QNR detects that case and will now throw an error, since we should really always be running with a correctly resolved Manifest.toml to reproducibility. The options are:

- update the Manifest.toml to `1.11.3`
- pin the version of `julia` installed in CI to `1.11.2` to match the current manifest
- pin it to `1.11.3` and update the committed Manifest.toml
- don't commit a Manifest.toml, instead instantiate it during CI so that it always matches the Julia version that CI is using.

cderv:
> don't commit a Manifest.toml, instead instantiate it during CI so that it always matches the Julia version that CI is using.

I think we commit the Manifest to do the same way we commit `renv.lock` for R, and `uv.lock` for python. 
This insure that CI is always run with a known set of dependencies, and then we control the update of the engine's dependencies. We did that so that our current work is not impacted by an update of any R, python or julia dependencies happening because we install latest version of everything. 

Is this correct usage of the `Manifest.toml` ? 

> * pin the version of julia installed in CI to 1.11.2 to match the current manifest
> * pin it to 1.11.3 and update the committed Manifest.toml

I believe we should do one or the other. We do pin minor version for other tools, I should have done it for Julia too 🤦

MichaelHatherly:
> Is this correct usage of the Manifest.toml ?

That's fine.

> I believe we should do one or the other.

I will do the second of those so we are on the most up to date stable release.

MichaelHatherly:
CI is all green now @cderv, should be good to merge.

cderv:
Thanks !

PR #12093, opened 2025-02-14
py-shiny - Use explicit = syntax for option

### LINKED ISSUE #12059: can't import custom module utils.py (Error: Got unexpected extra argument (utils.py)) when using quarto preview and `server: shiny`
### Bug description

I am experimenting with Quarto to create a dashboard with shiny. 
My dashboard runs well when I have all the code in the same file. 
However, since I refactored the code and moved functions and classes into the script utils.py, I have encountered some errors.

In the documentation, I found this solution: https://quarto.org/docs/computations/render-scripts.html. 

So I created the following _quarto.yml file:
```yaml
project: 
  title: "dashboard_quarto"
  render:
    - "*.qmd"
    - "!utils.py"
author: "x"
format: 
  dashboard:
    theme: morph
    nav-buttons: [github]
server: shiny
```

Now I always have the same error : 

Image

Thank you for your help :) 

### Steps to reproduce
main  file 
```{python}
from shiny import render, reactive, ui
from shinywidgets import output_widget, render_widget
import pandas as pd
import numpy as np
import plotly.express as px
import plotly.io as pio
pio.renderers.default = "notebook"

from utils import plot_x_y

np.random.seed(0)
n = 10  
salaries = np.random.randint(3000, 12000, size=n)
expenses = np.random.randint(1000, 8000, size=n)
data = {
    'Salaries': salaries,
    'Expenses': expenses
}
df = pd.DataFrame(data)

# Study
@render.data_frame
def dataview():
    return render.DataGrid(df, width="100%",height="100%")


@render_widget
def plot():
    fig = plot_x_y(df, "Salaries","Expenses")
    return  fig 
```

file utils.py 
```{python}
import pandas as pd
import plotly.express as px
def plot_x_y(
    df: pd.DataFrame, x_col: str, y_col: str
):
    return px.scatter(df, x_col, y_col)
```

file _quarto.yml
```
project: 
  title: "dashboard_quarto"
  render:
    - "*.qmd"
    - "!utils.py"
author: "x"
format: 
  dashboard:
    theme: morph
    nav-buttons: [github]
server: shiny
```

### Expected behavior

I will finally this dashboard : 

Image

### Actual behavior

_No response_

### Your environment

- quarto Version: 1.5.57
- pyhton 3.10

### Quarto check output

Image

cderv:
Otherwise, `quarto preview` does not work on Windows as quoting is not enough to avoid shell expension https://github.com/pallets/click/issues/2359

Could be revisited if shiny for python opt-out shell expansion

closes #12059 

Not sure if we can add test.

### DIFF
--- a/news/changelog-1.7.md
+++ b/news/changelog-1.7.md
@@ -47,6 +47,10 @@ All changes included in 1.7:
 - ([#11835](https://github.com/quarto-dev/quarto-cli/issues/11835)): Take markdown structure into account when detecting minimum heading level.
 - ([#11964](https://github.com/quarto-dev/quarto-cli/issues/11964)): Using panel layout without a crossref label now correctly do not add an empty `#block[]` that was leading to an unnecessary space in output.
 
+## Interactive Shiny Document
+
+- ([#12059](https://github.com/quarto-dev/quarto-cli/issues/12059)): `quarto preview` now correctly run on Windows an interactive Quarto doc with Jupyter engine and py-shiny, when using a `*.py` file as resource.
+
 ## Lua Filters and extensions
 
 - ([#11526](https://github.com/quarto-dev/quarto-cli/pull/11526)):
--- a/src/execute/jupyter/jupyter.ts
+++ b/src/execute/jupyter/jupyter.ts
@@ -515,8 +515,7 @@ export const jupyterEngine: ExecutionEngine = {
     ];
     if (options.reload) {
       cmd.push("--reload");
-      cmd.push(`--reload-includes`);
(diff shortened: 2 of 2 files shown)


cderv:
> This is a good fix, thanks. We should really try to avoid shell expansion anywhere we possibly can.

Totally ! 

For context, I discussed this with @gadenbuie - this is not expected really, it is specific to Windows due to how the CLI tool used in Python for py-shiny works. 

For reference, 
- https://github.com/pallets/click/issues/2359

Ideally, shiny for Python should probably opt-out this default shell expansion on Windows.

Issue #8170, opened 2024-01-08
Review Julia action in CI workflow as it seems caching does not work as we expect


cderv:

cderv:
I was looking into that in 
- https://github.com/quarto-dev/quarto-cli/pull/12108

It seems to be working, but with new V2 cache API in Github Action, it is creating lots of warning

- Cache is now immutable
- All matrix jobs is trying to save a cache with same key but only one succeed, other errors (which is expected). 
- We don't want to save a cache by buckets run because it would be too much (a cache for Julia is 300MB it seems). 


Maybe not using the `julia-actions/cache` and do our own caching will be a solution. 

I'll think about it.

PR #12108, opened 2025-02-18
gha - adapt cache saving to new actions/cache@v4

cderv:
`save-always` is deprecated, and now it needs to be a two step process: restore and save

based on adaptation at r-lib/actions/setup-renv in https://github.com/r-lib/actions/commit/108c7a6b3ecd9f71dbd0ce7d9e6ca9f57e2d244f

### DIFF
--- a/.github/actions/cache-typst/action.yml
+++ b/.github/actions/cache-typst/action.yml
@@ -9,7 +9,7 @@ outputs:
     value: ${{ steps.cache-typst-restore.outputs.cache-primary-key }}
   cache-matched-key:
     description: "Key of the cache found and used."
-    value: ${{ steps.cache-typst-restore.outputs.cache-primary-key }}
+    value: ${{ steps.cache-typst-restore.outputs.cache-matched-key }}
   cache-path:
     description: "where is the packages cache for typst ?"
     value: ${{ steps.cache-typst-path.outputs.TYPST_CACHE }}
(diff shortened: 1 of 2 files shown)


cderv:
It seems new cache API has changed a few things as we are hitting this issue 
- https://github.com/actions/cache/issues/1541

with

- my first attempt : [`7aad8ed`](https://github.com/quarto-dev/quarto-cli/pull/12108/commits/7aad8ed12569e8566dc8059c1f4930dd93f5bae6)
- Julia cache action

I subscribed to upstream issue and we'll see. 

Restoring and saving cache seems to have become more complicated, especially when a new cache key will be necessary. As we are running multiple workflow in parallel, in case of new cache
- They all have a cache not found at restore 
- So they will all try to save cache at save state but...
- ...cache can be created once and otherwise it will fail with warning. 

Not a problem to have a working workflow but a lot of noise as Warning in log 😞 

Let's see if they can improve this - I may find a workaround otherwise.

cderv:
This PR also try to fix Typst caching. The new cache API in github does not allow to save always the cache as before. It is now immutable and can't be updated. 

So this needs to be improved for more efficiency. 

Or we need to delete cache more regularly. I'll see later.

cderv:
So we get those warning 
image

quite verbose because of what is discussed at https://github.com/actions/cache/issues/1541

We'll live with that until I managed to improve it 
- https://github.com/quarto-dev/quarto-cli/issues/8170

PR #12121, opened 2025-02-19
Update `QuartoNotebookRunner.jl` to `0.13.1`

MichaelHatherly:
Updates the backend to the latest minor release.

This includes support for running `{python}` code cells (in addition to the already supported `{julia}` and `{r}`). It uses https://github.com/JuliaPy/PythonCall.jl to manage the `python` process for this, and allows passing values via `$` interpolation between all three languages.

The announcement post can be found at https://discourse.julialang.org/t/ann-native-julia-engine-for-quarto-using-quartonotebookrunner-jl/112753/56, and includes an example of using this new feature.

### DIFF
--- a/news/changelog-1.7.md
+++ b/news/changelog-1.7.md
@@ -69,7 +69,7 @@ All changes included in 1.7:
 ### `julia`
 
 - ([#11659](https://github.com/quarto-dev/quarto-cli/pull/11659)): Fix escaping bug where paths containing spaces or backslashes break server startup on Windows.
-- ([#12092](https://github.com/quarto-dev/quarto-cli/pull/12092)): Update QuartoNotebookRunner to 0.12.3.
+- ([#12121](https://github.com/quarto-dev/quarto-cli/pull/12121)): Update QuartoNotebookRunner to 0.13.1. Support for evaluating Python cells via [PythonCall.jl](https://github.com/JuliaPy/PythonCall.jl) added.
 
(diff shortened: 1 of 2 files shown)


cderv (commented):
This is interesting addition !! 

Should we document these new features with some links toward your docs at https://prerelease.quarto.org/docs/computations/julia.html#using-the-julia-engine ? 

At least, this probably needs to be added to the changelog as a more major change than the current addition, isn't it ? 
https://github.com/quarto-dev/quarto-cli/blob/ec8cd58685a4906f708c5d009117d1cb2e96e6f8/news/changelog-1.7.md#L72

MichaelHatherly:
> Should we document these new features with some links toward your docs at

Yes, I can add some updated docs covering the areas that are missing from there.

> At least, this probably needs to be added to the changelog as a more major change than the current addition, isn't it ?

Agreed, changelog should be updated as well. @cderv going forward I can just do the updates to the changelog directly in these version update PRs if you'd prefer? Or should it happen in follow up PRs?

cderv:
> I can just do the updates to the changelog directly in these version update PRs if you'd prefer? Or should it happen in follow up PRs?

Yes definitely, you can do it in the PR. I have been doing it after by convenience when I saw it. 

> I can add some updated docs covering the areas that are missing from there.

Awesome thanks. Please target the `prerelease` branch when you'll work on this. This is the version of the website for anything that is only 1.7 related. Thanks !

PR #12151, opened 2025-02-25
Fix YAML validation in cell metadata

cderv:
On windows, this does not fail and should

````markdown
---
title: "YAML intelligence should fail with knitr engine"
---

```{r}
#| echo: 123
1 + 1
```
````

And possibly using more backtick should fail always, and it is not
`````markdown
---
title: "YAML intelligence should fail with knitr engine"
engine: julia
---

This code cell has more than default 3 backticks

````{julia}
#| echo: 123
1 + 1
````
`````

### DIFF
--- a/news/changelog-1.7.md
+++ b/news/changelog-1.7.md
@@ -13,6 +13,8 @@ All changes included in 1.7:
 ## YAML validation
 
 - ([#11654](https://github.com/quarto-dev/quarto-cli/issues/11654)): Allow `page-inset` as value in `column` key for code cells.
+- ([#12151](https://github.com/quarto-dev/quarto-cli/issues/12151)): Fix YAML validation in computations cell on Windows.
+- ([#12151](https://github.com/quarto-dev/quarto-cli/pull/12151)): Basic YAML validation is now active in cell for document using Julia engine.
 
 ## Website projects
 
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by cderv** [comment]
@cscheid This PR also add some basic YAML validation for `engine: julia` doc. It seemed we did not have some, even basic. 

@MichaelHatherly @jkrumbiegel Just a heads up that when this PR is merged, YAML option syntax in cell will be now validated 

Document with non valid option like
````
---
title: "YAML intelligence should fail with knitr engine"
---
```{julia}
#| echo: 123
1 + 1
```
````
will now error. 

Other engine have some specific schema tag that allows to filter out specific schema. Example

* `tidy` option is specific to knitr engine
	https://github.com/quarto-dev/quarto-cli/blob/22b6c5d84a9ec2585b060a1d374f6e2e8a85b7d8/src/resources/schema/cell-codeoutput.yml#L117-L125

* `tags` is specific to Jupyter engine
	https://github.com/quarto-dev/quarto-cli/blob/22b6c5d84a9ec2585b060a1d374f6e2e8a85b7d8/src/resources/schema/cell-attributes.yml#L13-L18

If there are some specific options to support, this needs to happen in those schema files - Open a PR and I'll help.

PR #12277, opened 2025-03-13
Feature: cell renderings, light and dark

gordonwoodhull:
Enables the cell attribute

```
#| renderings: [light, dark]
```

If the cell contains two `div.cell-output-display`, the content will be put in a `div.light-content` and a `div.dark-content` instead.

Then `.dark-content` is hidden when `body` has `.quarto-light`, and vice-versa.

To do:
 - [ ] <strike>Warn if caption or label on the same cell, as they won't work right</strike>
 - [x] Add `has_renderings` flag in `flags.lua` and check in `main.lua`
 - [x] Use `.quarto-dark` instead of `.quarto-light` if dark theme is default

### DIFF
--- a/news/changelog-1.7.md
+++ b/news/changelog-1.7.md
@@ -43,6 +43,7 @@ All changes included in 1.7:
 
 ## `html` format
 
+- ([#12277](https://github.com/quarto-dev/quarto-cli/pull/12277)): Provide light and dark plot and table renderings with `renderings: [light,dark]`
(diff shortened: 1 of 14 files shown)


gordonwoodhull:
It's not as easy as I thought to detect captions and labels, due to the same complexities that make them difficult to handle properly. I think it could be done succinctly with walks, but I will leave it out and document it for now.

gordonwoodhull:
The filter is unintentionally discarding other output like `echo: true` code blocks.

It should retain any content before the first `.cell-output-display`.

Will fix and test before merging.

PR #12319, opened 2025-03-18
feature: dark brand

### LINKED ISSUE #12159: dark mode and brand yml
From https://github.com/posit-dev/brand-yml/issues/68

@Aariq reports

> I'm not sure if this belongs here or in the Quarto repo, but it appears like using theme: brand doesn't work for dark mode, only light mode. Perhaps this is just not implemented yet, but I was expecting something like this would work:
> 
> ```
> format:
>   html:
>     theme:
>       light: brand
>       dark:
>         - brand
>         - dark.scss
> ```
> where dark.scss only overrides background and foreground colors. However, when I do this, I see that in dark mode the fonts and everything else changes. Even just using light: brand and dark: brand makes it clear that _brand.yml is only applied to light mode.

Tracking here for Quarto 1.7

gordonwoodhull:
Fixes #12159 (in a more general way).

Addresses #11487

Example:
````yaml
brand:
  dark: dark-brand.yml
  light: light-brand.yml
````

Collate dark brand in brand sass bundles and use it for creating dark sass bundles.

`resolveBrand` now returns `{light?:, dark?:}`, basic downstream changes have been made.[^1] May need adjustment.

Enable toggling when brand is multi and includes dark (as previously for theme).

Includes some smoke tests; probably some edge cases need to be tested further.

[^1]: e.g. to still use `.light` for all the places we are pulling logo. No attempt yet to apply dark logo where needed: that would be a separate effort.

### DIFF
--- a/src/command/render/pandoc-html.ts
+++ b/src/command/render/pandoc-html.ts
@@ -88,28 +88,45 @@ export async function resolveSassBundles(
     // the brand bundle itself doesn't have any 'brand' entries;
     // those are used to specify where the brand-specific layers should be inserted
     // in the final bundle.
-    const brandLayersMaybeBrand = bundlesWithBrand.find((bundle) =>
+    const maybeBrandBundle = bundlesWithBrand.find((bundle) =>
       bundle.key === "brand"
-    )?.user || [];
-    assert(!brandLayersMaybeBrand.find((v) => v === "brand"));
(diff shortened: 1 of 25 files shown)


gordonwoodhull:
I've dealt with the major issues and I am merging this as-is once tests pass.

There are still many repercussions to deal with, including the lua brand metadata change, and this needs comprehensive tests via playwright. Will do all that next week.

PR #12400, opened 2025-03-28
do not emit dark stylesheets when there is no dark brand

gordonwoodhull:
Properly handle absent light or dark brand; do not emit empty light/dark brand bundles.

This fixes the syntax highlighting issue we were seeing. (But note the issue can still be seen with arrow syntax highlighting when JavaScript is disabled; see #12399.)

### DIFF
--- a/tests/docs/smoke-all/dark-mode/no-brand.qmd
+++ b/tests/docs/smoke-all/dark-mode/no-brand.qmd
@@ -1,0 +1,15 @@
+---
+format: html
+_quarto:
+  tests:
+    html:
+      ensureHtmlElements:
+        - []
+        - ["link[href*=\"-dark-\"]"]
+---
+
+## Hello
+
+```markdown
+![asdf](asd.png)
+```
--- a/src/command/render/pandoc-html.ts
+++ b/src/command/render/pandoc-html.ts
@@ -123,12 +123,12 @@ export async function resolveSassBundles(
       }
       return bundle as SassBundle;
     });
-    if (!foundBrand.light || !foundBrand.dark) {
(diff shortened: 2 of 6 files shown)


tests/docs/smoke-all/dark-mode/no-brand.qmd
cderv, line 1: Just sharing an interrogation that this test gives me. 

Our Quarto default theme (`theme: default`) does not work with brand ? 

I would have expected `brand: false` for a no-brand case, but it seems just no passing any theme works. So I wondered why. 

I guess I thought our `theme: default` was equivalent to `theme.light = default`, but it is not. Is that it? 

Sorry - not a review but I take the opportunity to clarify my understanding - hope you don't mind 😄

gordonwoodhull: I'm confused about this question. This is really a no-brand, no-theme test - it's the one we discussed yesterday.

`brand: false` will disable any brand specified in the project.

I don't find any difference between `theme: default` and `theme.light = default`, but I must be missing something.

My understanding is that we don't produce any warning for invalid theme names, and these are the same as `theme: foobar`, doing nothing.

Please let me know if these PRs have changed theme behavior in any way. They should not!

cderv: My questions shows I am still lacking a good understanding of how brand works internally with our initial theming system. 

I think I overthink it, and just that when there is no `_brand.yml` or no `brand` config, this is not brand. 

I think I was confused by previous discussion about brand, dark-light, etc... 

Thanks for your explanation.

PR #12406, opened 2025-03-28
remove julia test for #12271

cderv:
follow up of #12271 to fix some test failure with Julia

Something is different in CI than locally as I cannot reproduce. 

This is testing Jupyter and Knitr too. 

I'll check Julia later.


gordonwoodhull:
Ahhh, I think this explains a mysterious failure on #11918

cderv:
Sorry for this 😞

PR #12426, opened 2025-04-01
`respect-user-color-scheme` enables checking of prefers-color-scheme media query

### LINKED ISSUE #1470: FR: Websites automatically detect user preference for dark mode from OS

gordonwoodhull:
Fixes #1470 

When `respect-user-color-scheme` is enabled (default `false`), user preference for color scheme will be read from the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media query.

This mean that the author preference (via order of light and dark theme or brand in YAML) is unused, as long as JS is enabled. NoJS still uses stylesheet ordering based on author preference. [^1]

If the UI dark mode toggle is used, it gets saved to local storage and that take precedence. Currently the user can delete their local storage for the site to go back to their OS default, but in the future we might consider e.g. deleting the local storage setting if it matches the OS setting.

This doesn't do dynamic changes if the OS setting changes, so people who use Auto will have to refresh the page for now.

<strike>This tests correctly manually across browsers, but unfortunately it breaks automated tests in unexpected ways, and I haven't figured out how to fix those yet.

In particular, I have tried [Changing Color Scheme in Playwright Test](https://ray.run/questions/how-do-i-change-the-color-scheme-during-a-playwright-test) to no avail.</strike>

[^1]: We might consider a grand refactor using proper media queries, but I am just finding my way around the code and haven't considered that yet.


gordonwoodhull:
🗝️   When we test `prefers-color-schema`, we need to set it globally for those tests.

We don't have dynamic switching, so

```ts
await page.emulateMedia({ colorScheme: "dark" });
```
within a test won't work.

gordonwoodhull:
Seems to be dependency failure

```
    ! package or namespace load failed for 'flextable':
     package 'openssl' does not have a namespace
    Backtrace:
      1. global .main()
      2. execute(...)
      3. rmarkdown::render(...)
         ...
           at evaluate/R/evaluate.R:141:5
     17. base (local) withRestartList(expr, restarts)
     18. base (local) withOneRestart(withRestartList(expr, restarts[-nr]), restarts[[nr]])
     19. base (local) docall(restart$handler, restartArgs)
     21. evaluate (local) fun(base::quote(`<smplErrr>`))
```

cderv:
Yes possibly a conflict between package that would require update. I'll look at it.

gordonwoodhull:
Still seeing those flextable / openssl errors. I'm going to merge - pretty sure my tests are not on any of those buckets.

PR #12611, opened 2025-04-23
bugfix: giscus light dark toggle on load

gordonwoodhull:
Simpler alternative to #12608

Wait for giscus to load and then set its light/dark toggle immediately.

This is failing on Safari. May still need a Mutation Observer for the iframe.

### DIFF
--- a/src/resources/formats/html/giscus/giscus.ejs
+++ b/src/resources/formats/html/giscus/giscus.ejs
@@ -22,7 +22,7 @@
           if(giscusIframe) {
             giscusIframe.addEventListener("load", function() {
               window.setTimeout(() => {
-                window.toggleGiscusIfUsed(window.hasAlternateSentinel(), authorPrefersDark);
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Okay, it works with a Mutation Observer waiting for giscus to load. 

Although this is a long chain of callbacks, and we did not manage to avoid `MutationObserver`, IMO this is simpler and more straightforward than dynamically loading the giscus script.

gordonwoodhull:
Thanks for the report, @feinmann! What version of Quarto are you using?

gordonwoodhull:
Thanks @feinmann, I am able to reproduce with a site with Giscus but no dark mode.

I'll look into a fix for this.

gordonwoodhull:
@feinmann, this is fixed in 1.7.31

feinmann:
Thank you for the great work! It is already deployed :)

PR #12657, opened 2025-05-01
simple dynamic load of giscus

gordonwoodhull:
Solves wrong-theming of Giscus in Chrome when Giscus is scrolled off-screen.

Generate script tag based on user preference at page load.
No need for mutation observer because class is already set.

This is the v1.7 backport of #12659

### DIFF
--- a/src/resources/formats/html/giscus/giscus.ejs
+++ b/src/resources/formats/html/giscus/giscus.ejs
@@ -1,37 +1,33 @@
-<script src="https://giscus.app/client.js"
-        data-repo="<%- giscus.repo %>"
-        data-repo-id="<%- giscus['repo-id'] %>"
-        data-category="<%- giscus.category %>"
-        data-category-id="<%- giscus['category-id'] %>"
-        data-mapping="<%- giscus.mapping %>"
-        data-reactions-enabled="<%- giscus['reactions-enabled'] ? 1 : 0 %>"
-        data-emit-metadata="0"
-        data-input-position="<%- giscus['input-position'] %>"
-        data-theme="<%- giscus.theme %>"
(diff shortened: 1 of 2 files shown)


src/resources/formats/html/giscus/giscus.ejs
```diff
@@ -1,37 +1,33 @@
...
-<input type="hidden" id="giscus-alt-theme" value="<%- giscus.altTheme %>">
\ No newline at end of file
+<input type="hidden" id="giscus-alt-theme" value="<%- giscus.altTheme %>">
+<script>
+  function loadGiscus() {
+    // Function to get the theme based on body class
+    const getTheme = () => {
+      let baseTheme = document.getElementById('giscus-base-theme').value;
+      let altTheme = document.getElementById('giscus-alt-theme').value;
+      if (authorPrefersDark) {
```
amitness, line 9: Where is this `authorPrefersDark` variable defined? Utterances has stopped working for me because of this. Chrome raises an error due to undefined variable.

If I add the variable to the top of the page, it works.

gordonwoodhull: Thanks for reporting, @amitness! The bug with Giscus has already been reported and fixed (#12701) and will be released in the next patch release of Quarto. Sorry for the inconvenience.

I don't think you mean Utterances, because the `loadGiscus()` function is not included for Utterances.

amitness: Great to know, thank you! And yes I meant giscus.

gordonwoodhull: Fix released in 1.7.31

PR #12659, opened 2025-05-01
simple dynamic load of giscus

gordonwoodhull:
Solves wrong-theming of Giscus in Chrome when Giscus is scrolled off-screen.

Generate script tag based on user preference at page load.
No need for mutation observer because class is already set.

This PR is for 1.8, 1.7 backport is #12657


cderv:
Really cool! This is the logic I had in mind, I think. Awesome. Thank you Gordon !

PR #12667, opened 2025-05-02
feat: enhance category selection and filtering with multiple selection

mcanouil:
Introduce support for multiple category selection and filtering in the listing interface. This change allows users to select multiple categories simultaneously using Ctrl or Cmd.

https://github.com/user-attachments/assets/e8c41377-655a-48ac-944a-6be146d003cf


src/resources/projects/website/listing/quarto-listing.js
```diff
@@ -66,9 +81,14 @@ window.document.addEventListener("DOMContentLoaded", function (_event) {
     const category = decodeURIComponent(
       atob(categoryEl.getAttribute("data-category"))
     );
-    categoryEl.onclick = () => {
+    categoryEl.onclick = (e) => {
+      // Allow holding Ctrl/Cmd key for multiple selection
+      // Clear other selections if not using Ctrl/Cmd
+      if (!e.ctrlKey && !e.metaKey) {
+        selectedCategories.clear();
+      }
```
mcanouil, line 89: I don't know if that is a good idea as it makes the multiple selection feature "hidden".
I thought this was less disruptive than checkboxes everywhere.
This was the only idea I came up with to keep previous visual appearance.

ea-dc:
I was actually looking for a way to select multiple categories, but my need seems to be the opposite: I want to combine categories using an AND condition. That is, when selecting two categories, I’d like to show only the posts that belong to both.

I’m using Quarto 1.7.31. Is this new feature available (with AND or OR)? I couldn’t find anything about it in the documentation.

jandermoreira:
I was actually looking for a way to select multiple categories, but my need seems to be the opposite: I want to combine categories using an AND condition. That is, when selecting two categories, I’d like to show only the posts that belong to both.

I’m currently using Quarto 1.7.31. Is this new feature available (with AND or OR)? I couldn’t find anything about it in the documentation.

mcanouil:
Added tests, changelog, and fixed a hash check duplication.

PR #12683, opened 2025-05-05
configure ruff for project

cderv:
This is a new PR for #12446 

This PR complements the one for R 
- #12405 

Ruff VSCODE extension is now configuredto format all `*.py`

`*.ipynb` are skipped because of `#|` to `# |` change for cell options. 

configuration is done on `.ruff.toml` so that `ruff` works on CLI too.


.ruff.toml
```diff
@@ -0,0 +1,14 @@
...
+  "**/renv",
+  "**/.venv",
+  ".github",
+  ".vscode",
+  "**/node_modules",
+  "**/__pycache__",
+  "**/build",
+  # file with cell magic not parsed by ruff
+  "src/resources/jupyter/lang/python/cleanup.py",
+  "src/resources/jupyter/lang/python/setup.py",
```
cderv, line 11: @cscheid we already don't process those files as they don't play well with **ruff** as you guessed. This why they are not formatted in this PR.

PR #12752, opened 2025-05-16
typst-css - correctly handle unsupported font-weight values

cderv:
> [!NOTE]
> This PR solves the problem with 1.8.5 that prevent quarto-web to built

This is follow up from 
- #12554

Unsupported font-weight are translated to `nil` with a warning, but `quote()` does not handle `nil`

for example in the great_tables example `font-weight: initial` 

This PR handles the error by handling `nil` and do nothing


gordonwoodhull:
Thanks @cderv! LGTM

PR #12870, opened 2025-06-02
Update QuartoNotebookRunner to `0.17.3`

MichaelHatherly:
Includes two bug fixes, added a changelog entry about them. These would be safe to backport to the 1.7 release.

### DIFF
--- a/news/changelog-1.8.md
+++ b/news/changelog-1.8.md
@@ -69,6 +69,10 @@ All changes included in 1.8:
 
 - ([#12753](https://github.com/quarto-dev/quarto-cli/issues/12753)): Support change in IPython 9+ and import `set_matplotlib_formats` from `matplotlib_inline.backend_inline` in the internal `setup.py` script used to initialize rendering with Jupyter engine.
 
+### `julia`
+
+- ([#12870](https://github.com/quarto-dev/quarto-cli/pull/12870)): Update `julia` engine from `0.17.0` to `0.17.3` to improve `juliaup` detection on Windows systems and correctly set `Base.source_path()` output to match REPL and script usage.
+
 ## Other fixes and improvements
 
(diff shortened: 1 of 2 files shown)


cderv:
Thanks @MichaelHatherly ! 

> These would be safe to backport to the 1.7 release.

@cscheid, this is an interesting question. How do we consider update to QuartoNotebookRunner? 

In contrast to **knitr** or Jupyter, `engine: julia` is "tied" to a version of `QuartoNotebookRunner`, in the sense we ship the version information automatically installed within Quarto CLI. 

Do we consider it safe to "backport" an update of `QuartoNotebookRunner` ? 

@MichaelHatherly, for more context, our current policy with backport is that only regression (bugs introduced in 1.7 that were not in 1.6). For other bug fixes, we do not backport them. 

I see `QuartoNotebookRunner 0.17.3` is a bug fix only - are they related to regressions?

When would you switch to 0.18?  Could we consider it safe to backport any minor version once we decide 0.17 is the version shipped with Quarto? 

Just curious to have this discussion. Thanks

MichaelHatherly:
> are they related to regressions?

They are not fixes for regressions in the strict sense of the word I think, though they are non-breaking bug foxes for the `juliaup` support that has been in since 0.12.0 that broke in certain situations on Windows systems.

> When would you switch to 0.18?

When something gets added that is a definite feature addition, such as supporting a new field and the YAML front matter for example. (We should probably consider a v1 soon since it's all been pretty stable recently and I don't foresee and major changes to it going forward.)

MichaelHatherly:
Friendly bump, could this be merged?

cderv:
@MichaelHatherly I am merging for 1.8

We can discuss with Carlos the backport in another PR to v1.7 branch anyway. 

Sorry for the delay

PR #12895, opened 2025-06-05
Make Julia env copying robust against read-only scenarios

jkrumbiegel:
I noticed that I got failures on this line when using `quarto_jll` in Julia to run quarto. This happens because `copyFileSync` copies permissions of the source file as well, and all jlls in Julia have their source files under read-only permissions. This causes the Project.toml in the quarto cache directory to become read-only as well, which will make the next server startup fail because quarto attempts to overwrite the file.


cderv:
Thank you !

PR #12967, opened 2025-06-20
create - force `--no-open` when inside Posit  Workbench

### LINKED ISSUE #12965: quarto create should not automatically open new positron window when inside workbench
Shared by @MariaSemple and @jthomasmock on Slack

`quarto create` does try to be clever by detecting editors, and opening the new created project inside the same editor in a new window

This is a problem inside Posit Workbench where Positron can be opened. And from @jthomasmock
> On Workbench, inside a Positron terminal, $ quarto create with a new project will create the files and then relaunch the Positron session into that folder without prompting the user.

and from @MariaSemple
> the behaviour in Workbench feels unexpected for a user and could lead to some undesirable circumstances we're trying to avoid

Supported editors
https://github.com/quarto-dev/quarto-cli/blob/f2bb8943c7927fe0dbf320fae4ab7bb6b97301c2/src/command/create/editor.ts#L43-L47

Each editor defines a open function like
https://github.com/quarto-dev/quarto-cli/blob/f2bb8943c7927fe0dbf320fae4ab7bb6b97301c2/src/command/create/editor.ts#L153-L169

and quarto tries to check for editor and if detected, use the open function 
https://github.com/quarto-dev/quarto-cli/blob/f2bb8943c7927fe0dbf320fae4ab7bb6b97301c2/src/command/create/cmd.ts#L105-L114

https://github.com/quarto-dev/quarto-cli/blob/f2bb8943c7927fe0dbf320fae4ab7bb6b97301c2/src/command/create/cmd.ts#L199-L203

So we need to detect and make an exception inside workbench that we can already detect
https://github.com/quarto-dev/quarto-cli/blob/f2bb8943c7927fe0dbf320fae4ab7bb6b97301c2/src/core/platform.ts#L37-L41

cderv:
closes #12965 

@jthomasmock, @MariaSemple, this is simple tweak to force `--no-open` when inside workbench. 
I added info message in case a user is explicitly setting `--open <editor name>` to explain this is not permitted.  

I'll manually test it somewhere I can access. 

Is there any other specific behavior it should have in the Workbench context? Should we open automatically or prompt to open, but in a different way? 

You can also test this PR by cloning, switching branch and installing from source Quarto dev version

### DIFF
--- a/src/command/create/cmd.ts
+++ b/src/command/create/cmd.ts
@@ -8,7 +8,10 @@ import { extensionArtifactCreator } from "./artifacts/extension.ts";
 import { projectArtifactCreator } from "./artifacts/project.ts";
 import { kEditorInfos, scanForEditors } from "./editor.ts";
 
-import { isInteractiveTerminal } from "../../core/platform.ts";
+import {
+  isInteractiveTerminal,
+  isRStudioWorkbench,
+} from "../../core/platform.ts";
 import { runningInCI } from "../../core/ci-info.ts";
 
 import { Command } from "cliffy/command/mod.ts";
@@ -55,6 +58,16 @@ export const createCommand = new Command()
         // Compute a sane default for prompting
         const isInteractive = isInteractiveTerminal() && !runningInCI();
         const allowPrompt = isInteractive && !!options.prompt && !options.json;
+
+        // Specific case where opening automatically in an editor is not allowed
+        if (options.open !== false && isRStudioWorkbench()) {
+          if (options.open !== undefined) {
+            info(
+              `The --open option is not supported in Posit Workbench - ignoring`,
+            );
+          }
+          options.open = false;
+        }
 
         // Resolve the type into an artifact
         const resolved = await resolveArtifact(
--- a/src/command/create/cmd.ts
+++ b/src/command/create/cmd.ts
@@ -10,7 +10,7 @@ import { kEditorInfos, scanForEditors } from "./editor.ts";
 
 import {
   isInteractiveTerminal,
-  isRStudioWorkbench,
+  isPositWorkbench,
 } from "../../core/platform.ts";
 import { runningInCI } from "../../core/ci-info.ts";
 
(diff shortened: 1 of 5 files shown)


cderv:
Tested manually on Posit workbench - No new window opened.

```
$ ./package/dist/bin/quarto create 
Check file:///mnt/home/cderv/quarto-cli/src/quarto.ts
? Create › project
? Type › default
? Directory › ../testing-3
? Title (../testing-3) › Test
Creating project at /mnt/home/cderv/testing-3:
  - Created _quarto.yml
  - Created Test.qmd
````

````
$ ./package/dist/bin/quarto create --open vscode
The --open option is not supported in Posit Workbench - ignoring
? Create › project
? Type › default
? Directory › ../testing-4
? Title (../testing-4) › test
Creating project at /mnt/home/cderv/testing-4:
  - Created _quarto.yml
  - Created test.qmd
````

File: src/command/create/cmd.ts [hunk 1]
```diff
@@ -56,6 +59,16 @@ export const createCommand = new Command()
         const isInteractive = isInteractiveTerminal() && !runningInCI();
         const allowPrompt = isInteractive && !!options.prompt && !options.json;
 
+        // Specific case where opening automatically in an editor is not allowed
+        if (options.open !== false && isRStudioWorkbench()) {
```
jthomasmock: Minor nit and this might be an older boolean, but `isPositWorkbench` would be more appropriate.

jthomasmock (commented):
LGTM other than a minor nit about RStudio Workbench vs Posit Workbench.

src/command/create/cmd.ts
cderv: Yes this is an older function
https://github.com/quarto-dev/quarto-cli/blob/24e9355dec8f7fd7e5e2d474b7cd2774a0a4f46a/src/core/platform.ts#L37-L41

I did not bother renaming, but I can for sure.

cderv: Done in 18c12f5

Probably best to also change in quarto-dev/quarto ? 

https://github.com/quarto-dev/quarto/blob/57ed6fb2ab6066a6a4f0333c0c968586701e006a/apps/vscode/src/core/platform.ts#L22-L26

PR #13057, opened 2025-07-10
lua - use `cli_path()` information instead of building the path with the environment variable

cderv:
Follow up on https://github.com/quarto-dev/quarto-cli/pull/13053#issuecomment-3057244940

We have `quarto.config.cli_path()` set to the Quarto Binary path to used in filter. 

This ensures that the same Path as the quarto binary is used. 

I just wonder if `io.popen` to work correctly with `quarto.cmd` on Windows. 

Let's see if the CI passes.


cderv:
Tests are passing, so this should be ok even on windows. It means the typst tests are passing.

PR #13082, opened 2025-07-15
[create release] Add a step in each jobs to prevent re-running the job

cderv:
if it has already been executed once, we don't want to re-run it again as this would mess up version.txt and the release process. New builds needs to be triggered manually.


cderv:
We will now have an error in case of re-run
image

So this will prevent the problem we encountered.

PR #13108, opened 2025-07-21
typst logo partial

### LINKED ISSUE #13107: Hi @alinacherkas, you are understanding correctly. We do not currently have any general facilities for accessing brand metadata in templates.
Hi @alinacherkas, you are understanding correctly. We do not currently have any general facilities for accessing brand metadata in templates.

Given that templates are the same level of abstraction as brands, it would make sense to expose brand metadata directly to templates. 

But currently you can only access the entire brand metadata from Lua or via the [brand shortcode](https://quarto.org/docs/authoring/brand.html#shortcode---.quarto-shortcode__-escaped-data-is-shortcode1-data-value-brand), neither of which help you with templates. 

The background logo with positioning that we currently support in Typst

https://quarto.org/docs/authoring/brand.html#document-logo-customization

is implemented in this Lua filter:

https://github.com/quarto-dev/quarto-cli/blob/bf038503cd0290ef040288e763174497f5bd7d0b/src/resources/filters/quarto-post/typst-brand-yaml.lua#L217-L305

We have discussed separately that the logo should be implemented as a partial instead of this Lua filter directly generating the 

```typst
#set page(background: ...)
```

Exposing brand metadata to templates would be a similar effort, but more general, because it would expose all of the brand logos to the template, not just the chosen one. 

It could be done either through definition of more constants in Typst [as is done for colors](https://quarto.org/docs/authoring/brand.html#typst-2), or through more metadata exposed to the template as the Typst `brand.typography` metadata discussed here.

_Originally posted by @gordonwoodhull in https://github.com/quarto-dev/quarto-cli/discussions/11616#discussioncomment-13833001_

gordonwoodhull:
Fixes #13107 

This PR

* Moves the Typst logo declaration to a new partial.

   The `typst-brand-yaml.lua` filter populates the `logo` object in the pandoc metadata.

   So one can replace the partial, or change the `logo` properties after the Lua filter.

* Declares new `brand-logo` and `brand-logo-images` dictionaries in the Typst header, for access to all the resolved logos for the chosen `brand-mode`

Not sure if it's too late to add new Typst partials for 1.8. If not, we might also consider
* #11710

### DIFF
--- a/src/core/brand/brand.ts
+++ b/src/core/brand/brand.ts
@@ -10,6 +10,7 @@ import {
   BrandColorLightDark,
   BrandFont,
   BrandLogoExplicitResource,
+  BrandLogoResource,
   BrandLogoSingle,
   BrandLogoUnified,
   BrandNamedLogo,
@@ -143,11 +144,7 @@ export class Brand {
       }
     }
     for (const [key, value] of Object.entries(data.logo?.images ?? {})) {
-      if (typeof value === "string") {
-        logo.images[key] = { path: value };
-      } else {
-        logo.images[key] = value;
-      }
+      logo.images[key] = this.resolvePath(value);
     }
 
     return {
@@ -240,11 +237,7 @@ export class Brand {
     return fonts ?? [];
   }
 
-  getLogoResource(name: string): BrandLogoExplicitResource {
-    const entry = this.data.logo?.images?.[name];
-    if (!entry) {
-      return { path: name };
-    }
+  resolvePath(entry: BrandLogoResource) {
     const pathPrefix = relative(this.projectDir, this.brandDir);
     if (typeof entry === "string") {
       return { path: join(pathPrefix, entry) };
@@ -255,6 +248,13 @@ export class Brand {
     };
   }
 
+  getLogoResource(name: string): BrandLogoExplicitResource {
+    const entry = this.data.logo?.images?.[name];
+    if (!entry) {
+      return { path: name };
+    }
+    return this.resolvePath(entry);
+  }
   getLogo(name: BrandNamedLogo): BrandLogoExplicitResource | undefined {
     const entry = this.data.logo?.[name];
     if (!entry) {
--- a/src/format/typst/format-typst.ts
+++ b/src/format/typst/format-typst.ts
@@ -92,6 +92,7 @@ export function typstFormat(): Format {
         partials: [
           "definitions.typ",
           "typst-template.typ",
+          "logo.typ",
           "typst-show.typ",
           "notes.typ",
           "biblio.typ",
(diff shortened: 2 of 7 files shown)


gordonwoodhull:
@mcanouil, does this suit the use case you had in mind? 

IIRC you ran into conflicts customizing `page.background`.

mcanouil:
It seems so, I'll check more thoroughly tomorrow to see how this can work with the workaround I've put in place to add side text in margins.

My first thought is that I'm not sure we want to name that partial logo.
For instance, having to modify the logo partial in order to add background elements is a bit weird.
Since the partial is about setting the page background to set the logo but is not limited to the logo, maybe a name closer to what the partial set would be better: `page-background`?

mcanouil:
I confirm having the `set page()` has a partial would allow users to set/reset background with more than just the logo and because the logic would be in the template, it also allows to place the logo anywhere with any kind of transformation allowed by Typst.

Another thought, If the other `set page()` code from the article function were to be placed alongside the setting for the logo, it could be a general partial named `page-layout`.

- https://github.com/quarto-dev/quarto-cli/blob/9a0a0736d6f42c95dac455f8bae1c551dd503e5b/src/resources/formats/typst/pandoc/quarto/typst-template.typ#L32-L36

```typ
#set page(
  paper: $if(papersize)$"$papersize$"$else$"us-letter"$endif$,
  margin: $if(margin)$($for(margin/pairs)$$margin.key$: $margin.value$,$endfor$)$else$(x: 1.25in, y: 1.25in)$endif$,
  pagenumbering: $if(page-numbering)$"$page-numbering$"$else$none$endif$,
  background: $if(logo)$align($logo.location$, box(inset: $logo.inset$, image("$logo.path$", width: $logo.width$$if(logo.alt)$, alt: "$logo.alt$"$endif$)))$else$none$endif$,
)
```

(not 100% sure of the syntax)

gordonwoodhull:
Thanks @mcanouil!

I'm definitely on board with `page-background`. 

I'm a bit wary of diverging further from the Pandoc template, and any unforeseen consequences of moving the other page stuff, but that is a neat idea as well. I know users have asked for more control over page numbering. Hmm.

mcanouil:
My main concern is the partial will be used so it will be harder to change any decisions without making a breaking change, regardless of doing any of the proposals.

I do like the partial idea but I'm wondering if this should be a bigger work design not focusing on logo. Maybe a PR to Pandoc.

- https://github.com/quarto-dev/quarto-cli/issues/11709 (https://github.com/quarto-dev/quarto-cli/pull/11710)

gordonwoodhull:
Yeah... I think the fact that there are two `#set page(...)`, and we don't really know if they should be combined, indicates we should discuss further and maybe submit upstream. And it should be way more general than just for logos.

The title block partial is better defined, maybe we should do that first.

I'll convert this back to a draft for now.

gordonwoodhull:
Replaced by #13120

PR #13120, opened 2025-07-23
typst: page partial

gordonwoodhull:
The page partial takes page layout and numbering set-rules out of the `article()` function and puts them at top-level.

It also takes the logo set page background rule, which was previously generated in `typst-brand-yaml.lua`.

The partial gives users an easier way to customize page and logo options, or combine them with other page background properties.

### DIFF
--- a/src/format/typst/format-typst.ts
+++ b/src/format/typst/format-typst.ts
@@ -92,6 +92,7 @@ export function typstFormat(): Format {
         partials: [
           "definitions.typ",
           "typst-template.typ",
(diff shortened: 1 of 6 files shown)


gordonwoodhull:
As proposed by @mcanouil in https://github.com/quarto-dev/quarto-cli/pull/13108#issuecomment-3106371606, with the minor difference that we don't set `page.background` if we don't have a logo. (We don't want to remove background that may have been added elsewhere.)

Issue #11585, opened 2024-11-30
code-line-numbers and fragments not working as expected/hoped in revealjs


rpruim:
### Bug description

I would like to use `#| code-line-numbers and custom fragments to annotate revealjs slides, but I haven't been able to get it to work.  Is this possible?

Here is the kind of thing I have in mind.

````qmd
---
title: "code-line-numbers"
sbustitle: "and fragments"
format: revealjs
execute: 
  echo: true
---

## A slide

```{r}
#| code-line-numbers: "|1|2"
x <- 1
y <- 2
```

:::{.fragment fragment-index=1}
Comment on first line of code
:::

:::{.fragment fragment-index=2}
Comment on second line of code
:::
````

But when I view, the code highlighting is not synced up with the fragment indexing.

### Your environment

R version 4.4.1 (2024-06-14)
Platform: aarch64-apple-darwin23.4.0
Running under: macOS Sonoma 14.4
RStudio: 2024.09.1+394 (2024.09.1+394)

### Quarto check output

Quarto 1.5.56
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.2.0: OK
      Dart Sass version 1.70.0: OK
      Deno version 1.41.0: OK
      Typst version 0.11.0: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.5.56
      Path: /Applications/quarto/bin

[✓] Checking tools....................OK
      TinyTeX: (not installed)
      Chromium: (not installed)

[✓] Checking LaTeX....................OK
      Using: Installation From Path
      Path: /opt/homebrew/bin
      Version: undefined

[✓] Checking basic markdown render....OK

[✓] Checking Python 3 installation....OK
      Version: 3.12.4
      Path: /Users/rpruim/.virtualenvs/r-reticulate/bin/python3
      Jupyter: 5.7.2
      Kernels: python3

[✓] Checking Jupyter engine render....OK

[✓] Checking R installation...........OK
      Version: 4.4.1
      Path: /opt/homebrew/Cellar/r/4.4.1/lib/R
      LibPaths:
        - /opt/homebrew/lib/R/4.4/site-library
        - /opt/homebrew/Cellar/r/4.4.1/lib/R/library
      knitr: 1.49
      rmarkdown: 2.29

[✓] Checking Knitr engine render......OK

mcanouil:
[...]

Issue #7378, opened 2023-10-27
Feature request: fragment-index for code-line-numbers


maxpel:
One possibility to address the problem of setting fragment order easily in Rmarkdown for code-line-numbers, discussed here:

https://github.com/hakimel/reveal.js/issues/3055#issuecomment-1477534636

Include a new option `#| fragment-index: ""`

````
```{r, echo=T}
#| fragment-index: "0"
#| code-line-numbers: "|1|2|3"
x <- 1:10
x
LETTERS[x]
```
````

This would set the "start" index of the first code highlighting fragment, which would already solve my use case, which would be going through code in one chunk with highlight fragments in one column, after all those fragments showing the plot output as a fragment in the second column and then displaying some text fragment in the first column below the code chunk.

Issue #4899, opened 2023-03-20
Clearer argument name for Reveal.js highlighting with `code-line-numbers` 


malcolmbarrett:
`code-line-numbers` is an unclear name for using with Reveal.js to do highlighting, particularly since `code-line-numbers` does other important things to the output. IMO this deserves its own argument, e.g. `code-line-highlighting`, or at least an alias along those lines

mine-cetinkaya-rundel:
Just chiming in to say I like the idea of giving this a different name!

cderv:
Just adding a bit of context regarding why this name 

`code-line-numbers` is not specific to **revealjs** format. This is the configuration to trigger Pandoc to add numbered line to code chunk.  This is used in [html format](https://quarto.org/docs/output-formats/html-code.html#line-numbers) for example. 

In bare Pandoc, this is with `.number-line` class on chunk ([Pandoc Manual](https://pandoc.org/MANUAL.html#extension-fenced_code_attributes))

Revealjs has a feature [to highlight lines but for Highlight.js](https://revealjs.com/code/#line-numbers-%26-highlights) and the idea was to port this feature in Quarto to work with Pandoc syntax highlighter. Though in Revealjs, the feature is triggered by adding `data-line-numbers` to code part. 

This is the reason why when we added the feature in #147, we thought that reusing the `code-line-numbers` used in other format to add line numbers to revealjs with the specific addition of highlighting was better than adding a new specific option like `code-line-highlighting` specific to revealjs format

I am not against adding a new options but this means there will be two options 
* `code-line-numbers` which is TRUE by default for revealjs format 
* `code-line-highlighting` which will be unset by default, and used only for revealjs special feature 
* If `code-line-numbers` is FALSE, then `code-line-highlighting` would have no effect (or takes precedence - choice to make). 

Hopefully this clarify the initial choice of  keeping the same option name, while adding the feature.

cderv:
From a discussion with the team, idea to add also `code-line-numbers-start` instead of `startFrom` and this for all format, not just revealjs. 

And also proposal for new name: `code-line-numbers-highlight`

PR #13125, opened 2025-07-24
feat(reveal.js): add support for fragment indices in line highlighting

mcanouil:
Introduce support for fragment indices to enhance line highlighting functionality, allowing for more precise control over highlighted code segments in Reveal.js.

This adds a `code-line-fragment-indices` attribute to offer control over fragment.
When set, the same number of fragment indices as there are steps are required.

I'm not sure of the other changes that this new feature should be coming with.

````
---
format: revealjs
---

:::: {.fragment fragment-index="3"}
Step 3
:::

:::: {.columns}

:::: {.column width="50%"}
```{.r code-line-numbers="|2|3" code-line-fragment-indices="1,2,4"}
par(mfrow = c(1, 2))
plot(cars)
plot(cars, col = "red")
par(mfrow = c(1, 1))
```
:::

:::: {.column width="50%"}
```{.r code-line-numbers="|2|3" code-line-fragment-indices="1,2,4"}
par(mfrow = c(1, 2))
plot(cars)
plot(cars, col = "red")
par(mfrow = c(1, 1))
```
:::

:::

:::: {.fragment fragment-index="1"}
First step
:::
````


cderv:
Thanks for initiating ! 

Could be also related to 
- https://github.com/quarto-dev/quarto-cli/issues/7378
- https://github.com/quarto-dev/quarto-cli/issues/11585 (maybe ?)
 
Also, do we have an issue about this new feature with a discussion or only the Slack question? 
Discussing the issue, the design is sometimes easier, and I may have missed. So maybe the one above covers it. 

And regarding naming, I think I would like to handle this one too before adding new options
- https://github.com/quarto-dev/quarto-cli/issues/4899

Just listing all this as thoughts for later.

mcanouil:
Thanks for linking. The changes would indeed fix both issues. (I'm not linking them formally at this point to avoid risking closing the issues)

The PR is mostly here to not lose the code.

mcanouil:
I updated the PR for it to be more complete: changelog, test, etc.

PR #13193, opened 2025-08-07
Brand extension

### LINKED ISSUE #12559: Need section on how to share multiformat branding between Quarto projects

Report an issue on quarto.org

### Description

Using the [multiformat branding](https://quarto.org/docs/authoring/brand.html) section of the Quarto docs together with the [brand-yml](https://posit-dev.github.io/brand-yml/) docs, I have been able to put together a company branding that I am happy with.

The next step is to share that branding with other projects, and I find the documentation lacking in this regard. My intuition tells me it could be achieved with a Quarto extension (which one?), instead of redundant copy-pasting, but the bridge between these 2 sections is not an easy cross.

I believe a "Sharing brand" section (echoing the "Applying brand" one) would be very welcome. Unless I missed something else, which is quite possible.

Cheers.

gordonwoodhull:
Fixes #12559

This extension builds on the metadata extension, providing a sample brand.yml

To do this, we need to look at the project's `project.brand` and not just `brand`. Also includes @cscheid's fix merging the extension metadata during construction of the project context.

Posting as draft:
- [x] Needs tests
- [x] Currently distributes a sample `_quarto.yml` in order for the example in the extension to work — shouldn't the extension work without a project?


src/extension/extension.ts
```diff
@@ -808,8 +808,14 @@ async function readExtension(
           [],
         );
         if (resolved.include.length > 0) {
-          (object.project as Record<string, unknown>)[key] = resolved
-            .include;
+          if (key === "brand") {
+            (object.project as Record<string, unknown>)[key] = relative(
+              join(extensionDir, "..", ".."),
```
gordonwoodhull, line 813: This is somewhat fishy - is the extension directory always the grandchild of the project directory?

gordonwoodhull:
Do we test commands like `quarto create extension brand`?

Includes tests for brand extension _consumers_:
- logo (dashboard, website sidebar, typst). 
- typography (typst)

HTML format does not yet support file-based fonts #12501; we'll have to test them with brand extension when we implement them.

I couldn't get revealjs to render in this project so no test for that one
```
ERROR: NotFound: No such file or directory (os error 2): rename
'/Users/gordon/src/quarto-cli/tests/docs/smoke-all/brand/logo/logo-extension/test.html' ->
'/Users/gordon/src/quarto-cli/tests/docs/smoke-all/brand/logo/logo-extension/_site/test.html'
```

@cscheid, I think this is ready to merge, but please review https://github.com/quarto-dev/quarto-cli/pull/13193/commits/bc31efdebeb615883be9a764565ed5c68f8716d4 and the note above about grandparent directory.

gordonwoodhull:
Hmm, we're getting relative `img` paths with backslashes in HTML on Windows. Not quite there, changing back to draft.

gordonwoodhull:
Yeah, the change I made to `singleFileProjectContext` is too broad — this enables project mode for individual files when there is an extension, causing files to be rendered to `_site/` and I'm sure there are other side effects.

Let's review tomorrow!

gordonwoodhull:
I will submit a separate PR extending metadata extensions to single-file projects, which we can consider in the future. For now, we need a `_quarto.yml` in order to use brand or metadata extensions.

gordonwoodhull:
Last thing I'll mention here: post-processing of `href` paths under `_extensions` could not have been working on Windows prior to this PR.

It was looking for paths starting with `_extensions/`. Now it looks for either `_extensions/` or `_extensions\`.

I can't find any previous issues referencing this, but logo paths were not resolved without https://github.com/quarto-dev/quarto-cli/pull/13193/commits/135b59cb5ffeb4d57796bb67901d33bbf94f9930, and it's too risky to change Windows to use forward slashes for HTML output so late in the release.

src/extension/extension.ts
gordonwoodhull: Indeed, it's not! This breaks e.g. when an extension is installed from GitHub, because the path will be of the form `_extensions/github-org/extension-name`. Checking...

gordonwoodhull: Resolution: search upward in path to find `_extensions`. Not pretty but safe for our purpose.

PR #13218, opened 2025-08-14
fix: code-annotation not detected when non-R code blocks are present

### LINKED ISSUE #13216: Code annotations break with code linking AND embedded HTML
### I have:

### Bug description

Code annotation support is broken in HTML documents with embedded HTML and `code-link: true`. 

Removing the `{=html}` code block (and keeping `code-link: true`) returns styling to the code annotation. Alternatively, setting `code-link: false` (and keeping the `{=html}` code block) also returns styling to the code annotation. Behavior is the same whether `code-annotations` is set to `below`, `hover`, or `select`.

The issue seems related to https://github.com/quarto-dev/quarto-cli/issues/3945 where the solution was to automatically disable code linking when code annotations are enabled. In this case, adding an `{=html}` code block disrupts the exception.

### Steps to reproduce

````qmd
---
code-annotations: below
code-link: true
---

```{=html}
Embedded HTML breaks the styling of code annotations when <tt>code-link</tt> is true.
```

```{r}
2 + 2 # <1>
```

1. This should be styled as a popup.

````

### Actual behavior

The annotation marker in the code is not displayed. Code annotations are presented without styling in a list after the code. 

Image

### Expected behavior

The annotation marker in the code should be displayed. Code annotations should be presented in a styled list after the code. 

Image

### Your environment

- IDEs: Positron 2025.08.0 build 130 and RStudio 2025.05.0 Build 496
- OS: MacOS Sequoia 15.4.1

### Quarto check output

```bash
Quarto 1.7.33

      Quarto cache location: ~/Library/Caches/quarto

      Pandoc version 3.6.3: OK
      Dart Sass version 1.85.1: OK
      Deno version 1.46.3: OK
      Typst version 0.13.0: OK

      Version: 1.7.33

      TinyTeX: v2023.01
      Chromium: (not installed)

      Using: TinyTex
      Path: ~/Library/TinyTeX/bin/universal-darwin
      Version: 2022

      Using: Chrome found on system
      Source: MacOS known location

      Version: 3.9.6
      Path: ~/.virtualenvs/r-reticulate/bin/python3
      Jupyter: 5.7.2
      Kernels: python3

      Version: 4.5.0
      LibPaths:
      knitr: 1.50
      rmarkdown: 2.29

```

mcanouil:
Code annotation support was not correctly detecting annotations when non-R code blocks were present. 

Reformat via Air. (format on save in the repository)

Fixes #13216.

- [x] add test
- [x] add changelog entry


mcanouil:
I don't understand why [`docs/smoke-all/2025/05/19/issue-12775.qmd`](https://github.com/quarto-dev/quarto-cli/blob/4ef2c65a27a9f9df13877e8545ae14c14d86b061/tests/docs/smoke-all/2025/05/19/issue-12775.qmd) is failing as it's completely unrelated to the changes made here.

tests/docs/smoke-all/2025/08/14/issue13216.qmd
```diff
@@ -0,0 +1,24 @@
...
+
+```
+hello
+```
+
+```{r}
+2 + 2 # <1>
+```
+
+1. This should be styled as a popup.
```
cderv: By popup did you meant the style image
 ? 
 
 `code-annotations: below` and not `hover` is used, so I wasn't sure the meaning. 

Maybe this would be interesting to clarify this for future us (and future quarto developer). 

The current look is ok: 
image


Thanks for fixing this !

src/resources/rmd/rmd.R
```diff
@@ -112,7 +119,7 @@
 
           # look for at least one annotations that is in a code chunk
           for (a in annotations) {
-            if (chunkMap[a] == TRUE) {
+            if (chunkMap[a]) {
```
cderv: Just curious: Why did you remove `== TRUE` exactly ? Is this because you found this is unnecessary, so why not remove it ?  

I try usually to avoid changing code line that is not directly related to the PR and commit to have cleaner blame in history. So just wanted to confirm my guess. 

For practice sharing purposes: 

Usually I would do `isTRUE(chunkMap[a])` but `chunkMap[a] == TRUE` is the same. 
This is to protect again potential issue in `if` in case content is `NULL`

```
> x <- NULL
+ if (x) print("hello")
Error in `if (x) ...`:
! argument is of length zero
````

As we do `chunkMap <- rep(FALSE, lastLine)`, we can probably be sure that this won't happen here, but I find it is usually safer to not account for that. If for some reason the line is changed in the future, or deleted this would prevent issues. 

Anyhow, not important, this is a coding style and practice thing, and it is opinionated. So I don't mind changing.

**Review by cderv — APPROVED** [review]
Thanks a lot for fixing this !

File: src/resources/rmd/rmd.R (hunk 2 above)
Comment (mcanouil) [inline src/resources/rmd/rmd.R]:
```suggestion
            if (chunkMap[a] == TRUE) {
```

Revert

mcanouil: That was the idea, but I reverted that change to keep the PR's focus clean.

tests/docs/smoke-all/2025/08/14/issue13216.qmd
mcanouil: That was the example from the issue, but the content is actually indeed confusing. What matters is that code-annotation should work regardless of the mode.

PR #13224, opened 2025-08-15
feature: metadata extension works for single file

gordonwoodhull:
Continuation of
* #13193 
* #13221

This extends `singleFileProjectContext` to ingest metadata extensions.

This completely works and includes tests to prove it (with an extension but without `_quarto.yml`).

However, I was concerned that it has unknown side effects:
* It structures the output directories differently.
* I think I saw it change the output directory (but maybe I was mistaken).
* What else does it do? We don't know!

It is a clear generalization from what we currently have:
1. `singleFileProjectContext` takes `RenderOptions` instead of `RenderFlags`.
2. Export`mergeExtensionMetadata` from `project-context.ts`.
3. Create an empty `config.project`. (It could have `type: default`, which should do the same thing.)
4. `mergeExtensionMetadata` into the project.


cscheid:
This feels super risky now. I'm punting to 1.10 but would be ok with dropping and rethinking it through.

gordonwoodhull:
This was an early draft of functionality that later went in with engine extensions. We do support metadata extensions without `_quarto.yml` now.

PR #13288, opened 2025-08-28
possible bug in fs-extra v11.3.1

cderv:
Possibly related to a change in fs-extra so trying to pin a version.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -291,9 +291,10 @@ jobs:
         shell: bash
         run: |
           if [ ! -f "share/preview/quarto-preview.js" ]; then
-            echo "::error file=share/preview/quarto-preview.js::Required file share/preview/quarto-preview.js not found. Check the windows build configure step."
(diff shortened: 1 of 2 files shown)


cderv:
So this is tested with https://github.com/quarto-dev/quarto-cli/actions/runs/17296021231

It shows that this is building now. Basically, the update to `fs-extra@11.3.1` fails. 
It happens because of new release of **fs-extra** used by our old version of **vite-plugin-static-copy** (we use 0.13.1 - we should update)

It does not happen with prerelease version probably because npm is used now https://github.com/quarto-dev/quarto-cli/commit/e5ca74fa3c08c1c00504266946eecc0ad6a1bc8e

Still this quarto-preview.js needs to be improved. 

I noticed we build it twice 
 - One at configure time
 - Another at prepareDist time

cderv:
For reference, this is indeed a problem with fs-extra 11.3.1 but only with Deno due to limitation with compatibility with Node. See explanation at:
- https://github.com/denoland/deno/issues/30570#issuecomment-3253898638

PR #13439, opened 2025-09-26
bugfix: minimal: true would assert when combined with brand

### LINKED ISSUE #13383: `minimal: true` does fail when brand is found
````markdown
---
title: "Reproducible Quarto Document"
format:
  html:
    minimal: true
---

This is minimal
````

Have a `_brand.yml` in the project too. 

This leads to obscure error with 1.8
````
❯ quarto render index.qmd   
ERROR: AssertionError:
````

Stack trace
```
Stack trace:
    at Module.assert (https://jsr.io/@std/assert/0.224.0/assert.ts:18:11)
    at assert (https://jsr.io/@std/testing/0.224.0/asserts.ts:608:11)
    at resolveSassBundles (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/pandoc-html.ts:332:7)
    at eventLoopTick (ext:core/01_core.js:178:7)
    at async resolveExtras (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/pandoc.ts:1431:14)
    at async runPandoc (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/pandoc.ts:549:20)
    at async renderPandoc (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render.ts:202:24)
    at async Object.onRender (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-files.ts:729:30)
    at async renderFileInternal (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-files.ts:691:9)
    at async renderFiles (file:///C:/Users/chris/Documents/DEV_R/quarto-cli/src/command/render/render-files.ts:331:9)
```

This fails at 
https://github.com/quarto-dev/quarto-cli/blob/3c9950c77b952a3f2b72ba296c36c30f8cf4b47e/src/command/render/pandoc-html.ts#L332

We should probably either error better, by adding a message to the assertion error, 
but should probably make sure to not error in `resolveSassBundles` when `minimal: true` is set.

This is a regression from 1.7.34 as it renders correctly with it. Opting out brand yaml does avoid the error
````markdown
---
title: "Reproducible Quarto Document"
format:
  html:
    minimal: true
brand: false
---

This is minimal
````

@gordonwoodhull as discussed in Quarto workshop, we need to decide what to do in that specific case.

gordonwoodhull:
fixes #13383

v1.7 introduced this code implementing the light, dark, light stylesheet trick for syntax highlighting css.

it should not try to do this if there are no stylesheets!

we did not see the bug trigger until v1.8 which treats brand data as possibly implementing both light and dark
decided when we see theme

PR for CI; want to investigate a little further whether this means we're passing around  `hasDark` / `hasDarkStyles` that are always true.

### DIFF
--- a/src/command/render/pandoc-html.ts
+++ b/src/command/render/pandoc-html.ts
@@ -325,11 +325,11 @@ export async function resolveSassBundles(
       "dark",
       defaultStyle,
     );
-    if (defaultStyle === "light") {
+    if (defaultStyle === "light" && lightEntry) {
       const dep2 = extras.html?.[kDependencies]?.find((extraDep) =>
         extraDep.name === kQuartoHtmlDependency
       );
-      assert(dep2?.stylesheets && lightEntry);
+      assert(dep2?.stylesheets);
       dep2.stylesheets.push({
         ...lightEntry,
         attribs: {
--- a/tests/docs/smoke-all/brand/minimal-brand.qmd
+++ b/tests/docs/smoke-all/brand/minimal-brand.qmd
@@ -1,0 +1,11 @@
+---
+title: "Reproducible Quarto Document"
+format:
+  html:
+    minimal: true
+brand:
+  color:
+    background: '#cfc'
+---
+
+This is minimal


gordonwoodhull:
Uncovered #13450 while testing this.

This code shouldn't be hit, but for now let's make sure it doesn't crash.

I'll proceed with backport and changelog on 1.8.

Issue #13474, opened 2025-09-30
Typst brand settings always triggered for creating title


christopherkenny:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

The current setup for Typst has incorrect logic for checking for brand metadata when creating the title heading block.

Specifically, the template currently sets the brand family to [Libertinus Serif](https://github.com/quarto-dev/quarto-cli/blob/fbfe3aa7a601f8cb2e7796d2506dddff0e902b6c/src/resources/formats/typst/pandoc/quarto/typst-template.typ#L17C1-L17C38) by default. This means that when checking if any of the brand components are set, it will always run [this if for the header](https://github.com/quarto-dev/quarto-cli/blob/fbfe3aa7a601f8cb2e7796d2506dddff0e902b6c/src/resources/formats/typst/pandoc/quarto/typst-template.typ#L38).

### Steps to reproduce

Simply set `mainfont` without using a `brand.yml` file.

````md
---
title: "Header logic"
format: typst
mainfont: Comic Sans MS
---

Some text.

# Header logic so we have the same letters

More text.
````

Everything should be in the whimsical comic sans font.

### Actual behavior

<img width="448" height="182" alt="Image" src="https://github.com/user-attachments/assets/1d0d1b24-3a48-48ac-91a1-ac0fa6b589dd" />

### Expected behavior

Everything should be in Comic Sans (for better or for worse).

### Your environment

- IDE: RStudio 2025.09.0 Build 387
- OS: Windows 11

### Quarto check output

```bash
Quarto 1.8.24
[>] Checking environment information...
      Quarto cache location: C:\Users\chris\AppData\Local\quarto
[>] Checking versions of quarto binary dependencies...
      Pandoc version 3.6.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.3.1: OK
      Typst version 0.13.0: OK
[>] Checking versions of quarto dependencies......OK
[>] Checking Quarto installation......OK
      Version: 1.8.24
      Path: C:\Users\chris\AppData\Local\Programs\Quarto\bin
      CodePage: 1252

[>] Checking tools....................OK
      TinyTeX: (external install)
      Chromium: (not installed)

[>] Checking LaTeX....................OK
[...]

PR #13475, opened 2025-09-30
heading-family should also default to mainfont

gordonwoodhull:
fixes
* #13474

`mainfont` has not been applied to the title since the implementation of typst brand yml in 1.6.

This fix prefers `brand.typography.headings.family` but falls back to `mainfont`

Perhaps we should follow the same precedence  for `brand.typography.base.family` vs `mainfont` a few lines up

https://github.com/quarto-dev/quarto-cli/blob/d5110ff7023e0ea8c80b3d26ca04c80b283acabd/src/resources/formats/typst/pandoc/quarto/typst-show.typ#L32-L36

Filing this made me wonder if we should handle `mainfont` in `typst-brand-yaml.lua` instead, so filing this as draft and I think I prefer the next PR over this one.


gordonwoodhull:
This stays closer to the Pandoc partial, and keeps logic there. 

It's slightly messier logic than treating `mainfont` exactly like `brand.typography.base.family` as tried in
* #13476

but probably easier to maintain anyway.

Will merge after #13470

PR #13476, opened 2025-09-30
heading-family should default to mainfont in typst-brand-yaml.lua

gordonwoodhull:
fixes
* #13474

Alternative PR to #13475.

A simpler approach, but one that diverges further from the Pandoc `default.typst`.

Process `mainfont` as if it sets `brand.typography.base.family`.

I think this is cleaner and safer, and doesn't have any questions of precedence.

But I'll sleep on it before merging this or the other one.


gordonwoodhull:
I'm closing this in favor of #13475.

Despite this being a cleaner approach in terms of treating `mainfont` exactly the same as `brand.typography.base.family`, it is not as safe, and it's probably a good practice to keep logic in the templates where can.

PR #13507, opened 2025-10-06
Correctly check for sidebar.logo in EJS template 

### LINKED ISSUE #13506: Active nav is no more the right one when the page has a sidebar
In Quarto website and probably book, when there is navbar for a page with a sidebar, the active nav is not the right one. 

See below for Quarto Web - the CSS for active link being blue shows the overview is always selected

Image

This was reported by @ashleyhenry15

This was tricky to debug, but this is because: 

We do set the active nav based on some sidebar finding and href comparaison
https://github.com/quarto-dev/quarto-cli/blob/13fd02c059ac49e57822c3bdfdb65fe518bf6bb5/src/project/types/website/website-navigation.ts#L555-L572

but this has been broken by the fact the sidebar may have a logo with a link:
https://github.com/quarto-dev/quarto-cli/blob/13fd02c059ac49e57822c3bdfdb65fe518bf6bb5/src/resources/projects/website/templates/sidebar.ejs#L35-L46

This part of the template is activated now always because sidebar.logo has been normalized in 
- https://github.com/quarto-dev/quarto-cli/pull/12996
and it is no more undefined
````
{
  light: undefined,
  dark: undefined,
}
````

So `if(sidebar.logo)` is always true and this creates some HTML that are not expected in the page 
````
<!-- sidebar -->
  <nav id="quarto-sidebar" class="sidebar collapse collapse-horizontal quarto-sidebar-collapse-item sidebar-navigation floating overflow-auto">
    <div class="pt-lg-2 mt-2 text-left sidebar-header sidebar-header-stacked">
      <a href="./index.html" class="sidebar-logo-link">
      </a>
      </div>
````

which messes up with the search 
https://github.com/quarto-dev/quarto-cli/blob/13fd02c059ac49e57822c3bdfdb65fe518bf6bb5/src/project/types/website/website-navigation.ts#L561-L563

hence the `.active` class not being set on the right nav. 

So 

- We can't test `if(sidebar.logo)` anymore, as it seems it won't be undefined but `sidebar.logo.light` and `sidebar.logo.dark` will
- We probably need to protect the search for avoid selecting the link inside the logo part. 

This is a regression and we'll need to backport.

cderv:
sidebar.logo is now normalized as an object with light and dark properties

So `sidebar.logo` is never undefined - it may have light or dark undefined, but the object itself exists.

fix #13506

### DIFF
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -5,6 +5,7 @@ All changes included in 1.9:
 - ([#13396](https://github.com/quarto-dev/quarto-cli/issues/13396)): Fix `quarto publish connect` regression.
 - ([#13441](https://github.com/quarto-dev/quarto-cli/pull/13441)): Catch `undefined` exceptions in Pandoc failure to avoid spurious error message.
 - ([#13046](https://github.com/quarto-dev/quarto-cli/issues/13046)): Use new url for multiplex socket.io server <https://multiplex.up.railway.app/> as default for `format: revealjs` and `revealjs.multiplex: true`.
+- ([#13506](https://github.com/quarto-dev/quarto-cli/issues/13506)): Fix navbar active state detection when sidebar has no logo configured. Prevents empty logo links from interfering with navigation highlighting.
 
 ## Dependencies
 
--- a/src/core/brand/brand.ts
+++ b/src/core/brand/brand.ts
@@ -344,9 +344,14 @@ export function resolveLogo(
     return logo;
   };
   if (!spec) {
+    const lightLogo = findLogo("light", order);
+    const darkLogo = findLogo("dark", order);
+    if (!lightLogo && !darkLogo) {
+      return undefined;
+    }
     return {
-      light: findLogo("light", order) || findLogo("dark", order),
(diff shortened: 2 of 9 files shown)


cderv:
While trying to add some tests, I am now wondering how we are supposed to pass logo when we do hybrid navigation: https://quarto.org/docs/websites/website-navigation.html#hybrid-navigation

- one logo per sidebar ? 
- one logo for all sidebar ? 

🤔

gordonwoodhull:
As discussed, we are only resolving the logos in `sidebars[0]`, but we should probably resolve logos in all sidebars:

https://github.com/quarto-dev/quarto-cli/blob/42623202862d46086153fefee404d8ed163c878a/src/project/types/website/website-shared.ts#L174-L208

cderv:
Thanks. That is explaining it ! Surely not clear from the doc ! 

I think I'll handle it in another issue

cderv:
setting `sidebar.logo` to undefined required to properly handle the logo in Typst Lua brand support. Our test caught it which is cool!

Issue #13515, opened 2025-10-07
Document `page-numbering` for Typst


cwickham:
Since https://github.com/quarto-dev/quarto-cli/commit/5c56f77c47381c9491639507a3f739da04a19405 you can do:

````markdown
---
title: "Untitled"
author: Charlotte Wickham
format: 
  typst:
    page-numbering: ""
---

## Introduction

{{< lipsum 2 >}}

## Section

{{< lipsum 3 >}}

````

But we don't document it, either in the [Reference](https://quarto.org/docs/reference/formats/typst.html#numbering) or [Guide > Documents > Typst > Typst Basics](https://quarto.org/docs/output-formats/typst.html)

xx02al:
Hi @cscheid  and @cwickham 

I hope you're open to take a moment, even though that late, to discuss the default value of `page-numbering`.

The current implementation defaults to the hard-coded `"1"`.

I would argue `page-numbering` should default to `none` (or `false` if Quarto uses boolean values only). This allows us to use Typst's `page()` function with the default `numbering: none` **without explicitly setting `page-numbering: none`**. What actually happens is that Quarto displays a page number. Thus, if Quarto were to default to `none`(`false`), (IMO) extension code would be cleaner and more transparent.

[Quarto's own templates](https://github.com/quarto-ext/typst-templates) illustrate this. Take the `letter` as an example. It has `set page(paper: "us-letter", margin: (top: 2cm))` suggesting an intent to use the default `numbering: none` but instead, Quarto's default `page-numbering: true` takes over and displays a page number. This intent is supported by the `.png` shown. (I am aware that it was generated with a Quarto version that did not have the `page-numbering` option, but it still illustrates the point). As said, we would need to explicitly set `numbering: none` in the template to achieve the intended result, which is not ideal.

mcanouil:
@xx02al Please open a ["feature request" discussion](https://github.com/orgs/quarto-dev/discussions/new?category=feature-requests). Closed issues are not suited for such discussion.

xx02al:
https://github.com/orgs/quarto-dev/discussions/14102

PR #13516, opened 2025-10-07
Add `page-numbering` to schema

cwickham:
## Description

Add `page-numbering` to the schema for `format: typst`

Required for https://github.com/quarto-dev/quarto-cli/issues/13515


cderv:
Was this added in v1.8? 

It would mean we need to backport the schema change to v1.8 too so that it works in auto-completion and YAML validation

cwickham:
I think 1.7 😬

cderv:
backporting to 1.8 seems needed then. Too late for 1.7 😓

cderv:
Sorry, I forgot to merge this. I fixed the conflict and the typo. I'll open the backport too

PR #13524, opened 2025-10-09
Support inserted plausible snippet from website config 

cderv:
This is to have a simple built-in support for a `plausible-analytics` key, like we have for `google-analytics`.

I say simple because what this does it only accepting the HTML snippet that a user is supposed to get from its plausible dashboard: https://plausible.io/docs/plausible-script

````
website:
  title: "Plausible Custom Script Test"
  plausible-analytics: |
    <script async src="https://plausible.io/js/pa-TESTCODE123.js"></script>
    <script>
      window.plausible=window.plausible||function(){(plausible.q=plausible.q||[]).push(arguments)};
      plausible.init()
    </script>
````

It seems there are ways to use a generic `https://plausible.io/js/script.js` and pass domain as attributes. However, this is not documented and could be harder to maintain. 
Also the user workflow will always be to go to its plausible dashboard to activate, and so copy pasting the snippet is easy enough. 

More complex config would allow something like `google-analytics`

```
website:
  plausible-analytics:
    domain: example.com
    ... 
```

but again, not well documented and so probably not easy to maintain and test.


cwickham:
I like this balance: a dedicated key gives us an obvious place to document it (alongside `google-analytics`), while relying on a copy-paste reduces the maintenance burden.

cderv:
Great! 

I'll make a PR to the doc too.

cderv:
We could also do this to allow less ugly configuration

````
plausible-analytics: 
  path: _plausible.html
````

in addition of 

````
website:
  plausible-analytics: |
    <script async src="https://plausible.io/js/pa-XXXXXX.js"></script>
    <script>
      window.plausible=window.plausible||function(){(plausible.q=plausible.q||[]).push(arguments)},plausible.init=plausible.init||function(i){plausible.o=i||{}};
      plausible.init()
    </script>
````

cderv:
@cscheid I have added the `path` support as we discussed. I think this is better to have it now.

PR #13525, opened 2025-10-09
Fix cookie-consent opt-out and algolia insights

cderv:
Add correct cookie consent handling to search options and Algolia initialization

when `website.cookie-consent: false` we need to make sure that Algolia Insigth does not use cookie

Previous we had some logic to deferred loading when scripts where hidden behind `cookie-consent`, but would still use cookie always even when `cookie-consent: false`, the default. 

@cwickham @cscheid this should solve the problem on quarto-web were we default to `cookie-consent: false`

Doing this change will mean that `quarto` CLI will now default to not using cookie ever. And if a users wants to make a website that use cookie, then it needs to be behind `cookie-consent: true`

Is this expected behavior ?  This makes our default compliant to places where there are regulation like GDPR 

Or do we want to have a way to set `useCookie: true` explicitly even with `cookie-consent: false` for website that don't care 🤷‍♂️ 

I would think not, but I prefer to ask

## Test comment

I did only static test to check we do insert the thing we expect. 

I would like to do playwright test, to check for cookie or no cookie (https://playwright.dev/docs/api/class-browsercontext#browser-context-cookies) but this would require a valid way to connect to Algolia API.


cwickham:
I'm in favor of forcing the use of `cookie-consent: true` if someone wants Algolia to use cookies:

* It prevents people from unintentionally violating things like GDPR
* It signals we (as a team/company) think not violating things like GDPR is important

We may surprise anyone who currently has `analytics-events: true` and not `cookie-consent: true` with breaking some aspect of search analytics they use, but:

1. I'm not sure there are many people in this situation. A GitHub search revealed most examples with `analytics-events: true` are forks of `quarto-web`. We might give the folks at [ibis](https://github.com/ibis-project/ibis/tree/main) a heads up.

2. Most of these people are genuinely unaware that Algolia is using a cookie, and a change that forces compliance with a way to opt back in to cookies is probably better than a change that doesn't alert them to the problem.

cderv:
Ok great ! that confirms my thinking. Thanks

> We might give the folks at [ibis](https://github.com/ibis-project/ibis/tree/main?rgh-link-date=2025-10-09T18%3A30%3A41Z) a heads up.

Good call! I'll open an issue in their repo once merged, ahead of release.

cderv:
BTW regarding default cookie behavior, from the document at https://quarto.org/docs/websites/website-tools.html#cookie-consent I see the cookie consent feature was implemented with an implied default. 
Meaning user can change preference, but cookie will be used by default. 

Example used with GA is
````yaml
website:
  cookie-consent:
    type: express
    style: headline
    palette: dark
  google-analytics:
    tracking-id: "G-XXXXXXX"
    anonymize-ip: true
````

And in this case, this more aligned with express content to use cookie in GA. 

So this raises question of the default we want for cookie, and also if using `google-analytics` should set `cookie-consent: true` by default...

cwickham:
IIUC we currently have the behavior:

Algolia Analytics: 

- `cookie-consent: false` Compliant, no cookie
- `cookie-consent: true` Compliant, has cookie unless viewer opts-out
- `cookie-consent: type: express` Compliant, no cookie unless viewer opts-in

Google Analytics:

- `cookie-consent: false` Not compliant, has cookie and user cannot opt-out
- `cookie-consent: true` Compliant, has cookie unless viewer opts-out
- `cookie-consent: type: express` Compliant, no cookie unless viewer opts-in

I think the two questions are (edit: upon re-reading your comment ☝🏻 I don't think you are actually asking 2.):

1. Should using Google Analytics default to `cookie-consent: true`?
2. ~Should `cookie-consent` default to `type: express`?~

For 1, I'm inclined to say, "yes", it then means the behavior is consistent between Algolia and GA, and encourages people to be compliant. It will surprise some people, but I think it's OK for the same reasons as mentioned above for Algolia.

~For 2, I'm less sure. Personally, I prefer to have to opt-in to cookies rather than opt-out, but this would be a big visual change for any site currently using the default settings of  `cookie-consent: true`. So, I'm inclined to say, "no".~

cwickham:
We might get more pushback from people wanting to use GA without `cookie-consent: true`. I think it's OK for us to say we don't support that (if they really want to do it, they can set up GA without using our `google-analytics` helper)

cwickham:
Actually, now I'm not sure about our default for `cookie-consent`. My impression is GDPR requires **opt-in**, which means if we take a stance on encouraging compliance `cookie-consent: type: express` should be the default.

cderv:
Thanks a lot on the feedback. 

I had a call with Carlos so sharing also what we discussed. 

> Should using Google Analytics default to cookie-consent: true?

we should do that - but still allow `cookie-consent: false` if set explicitly. I think we can do that. 
So users that want GA without cookie-consent could still do it. 

> Actually, now I'm not sure about our default for cookie-consent. My impression is GDPR requires opt-in, which means if we take a stance on encouraging compliance cookie-consent: type: express should be the default.

This is a good call ! We should probably set the `type: express` to be the default when `cookie-consent: true`. 

I'll do this in another PR though.

PR #13547, opened 2025-10-13
`cookie-constent: { type: express }` is now the default.

cderv:
This means loading cookie is now opt-in instead of opt-out.

This follows the change and discussion in 
- https://github.com/quarto-dev/quarto-cli/pull/13525

@cwickham @cscheid as discussed. This is changing the default for an opt-in behavior now related to cookie on website. Users needs to configure explicitly for different beheavior


cscheid:
Thanks! This is important and welcome. We will need a careful update of our documentation for it, though.

cderv:
OK. I forgot to change value in the schema too. Done now. 

> We will need a careful update of our documentation for it, though.

I'll do a complement to this one: 
- https://github.com/quarto-dev/quarto-web/pull/1806

Issue #12803, opened 2025-05-22
typst format's callouts show different text alignment with Prefix


jeremy886:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

After upgrading to the latest version of Quarto, callouts are rendered differently with the prefix #tip-. All text is centred instead of left-aligned.

If `#tip-xxx` is removed, then the callout will be displayed normally.

### Steps to reproduce

````qmd
---
format:
  typst:
    papersize: a4
---


::: {.callout-tip #tip-abc }  
## Tip with Title

This is an example of a callout with a title.

This is an example of a 'folded' caution callout that can be expanded by the user. You can use `collapse="true"` to collapse it by default or `collapse="false"` to make a collapsible callout that is expanded by default.
:::

::: {.callout-tip}  
## Tip with Title

This is an example of a callout with a title.

This is an example of a 'folded' caution callout that can be expanded by the user. You can use `collapse="true"` to collapse it by default or `collapse="false"` to make a collapsible callout that is expanded by default.

:::

````

### Actual behavior

The text is centered.

### Expected behavior

The text is left-aligned

### Your environment

- IDE: Positron
- MacOS: 15.5

### Quarto check output

Quarto 1.7.31
[✓] Checking environment information...
      Quarto cache location: /Users/jeremy/Library/Caches/quarto
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.6.3: OK
      Dart Sass version 1.85.1: OK
      Deno version 1.46.3: OK
      Typst version 0.13.0: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.7.31
      Path: /Applications/quarto/bin

[✓] Checking tools....................OK
      TinyTeX: (not installed)
      Chromium: (not installed)

[✓] Checking LaTeX....................OK
      Tex:  (not detected)

[✓] Checking Chrome Headless....................OK
      Using: Chrome found on system
      Path: /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
      Source: MacOS known location

[✓] Checking basic markdown render....OK

[✓] Checking Python 3 installation....OK
      Version: 3.12.10
[...]

PR #13590, opened 2025-10-21
Add support for icon=false in typst callout

cderv:
closes #13555 

This PR brings support for `icon=false` in callouts

image

(centering problem for referenced callouts is to be dealt with in https://github.com/quarto-dev/quarto-cli/issues/12803)

@gordonwoodhull could you review this PR for thoughts maybe ? Thank you.


gordonwoodhull:
LGTM! Surprises me that we support false but not values here, but that’s how it is across formats. 🤷‍♂️

gordonwoodhull (approved):

cderv:
Thanks for the review !

> but not values here,

What do you mean? Values as choosing the icon to use for a callout?

I guess this relates to Custom Callout supports in a way, and we should maybe work on this wished feature:
- https://github.com/quarto-dev/quarto-cli/issues/844

To be discussed there.

PR #13619, opened 2025-10-27
Add automatic detection update of new patterns in tinytex

cderv:
Quarto LaTeX engine can automatically install missing packages and fonts because it parsed log file. 

The parsing happens through known error using regex patterns. 

Quarto leverages TinyTeX mainly, and experience in **tinytex** R package. 

On tinytex side, we did export the patterns through `regex.json` file so we can sync between project. We haven't done so. 

It is like the `tinitex` CLI tool that is supposed to have all the clever latexmk logic in one tool instead of duplicating in here and in tinytex. 

First step is using the `regex.json`. This PR adds a workflow file to check daily for new entries, and open an issue for us to look at so that we don't miss any update in the future


cderv:
Test on my fork and leads to create issue like this 
image

On first run, there will be no diff and an initialization leading to some summary step
image

PR #13682, opened 2025-11-13
gha - separate cloudsmith publishing in a new workflow

cderv:
this way wa can manually call it to publish release

Aim is to start publishing v1.8 patch release there


cderv:
This require we do add rpm and deb assets to 1.8 though..

PR #13701, opened 2025-11-17
feature: engine extensions and quarto api

gordonwoodhull:
I still want to add some specific tests, but it's time to make this a draft PR instead of testing directly in CI.

I've addressed the concern about `output-dir` by
* Warning if an extension contributes `output-dir`.
* Adding the `forceClean` flag in this case, as we do with `--output-dir`.

I also searched for extensions that use this, and only found three.

Unfortunately, I was not able to squash the commits thematically... way too much churn and merge conflicts.

So they are squashed chronologically into 10 commits, following the plan which added some temporary infrastructure and flags, and then removed it. 

The commit of concern, enabling metadata and engine extensions for single files, is separate in 3a813d5, in case we need to revert it.


gordonwoodhull:
Wow! Building the package and `tinitex` opened up new dependency hells.

The most noteworthy one is that external engines depend on import maps, not bundled dependencies, at least in the current iteration. Will discuss internally.

This still needs more tests, but maybe it's closer to merge, we'll see.

gordonwoodhull:
Closing due to GH refusing to run tests due to spurious merge conflicts. Will replace

PR #13718, opened 2025-11-20
feature: engine extension

gordonwoodhull:
replaces #13701

Working branch for CI while I work out dependency and build issues.


gordonwoodhull:
I think this PR has passed through bundler hell and could be ready to merge.

It required some dynamic imports due to new dependencies from check and command to render modules using module level await, triggering esbuild bundler bugs in test-bundle

Then those fixes caused spurious eslint variables not found, disabled in validate-bundle.

So, yeah! I’m not excited about merging it with problems I fundamentally don’t understand. I’m going to sleep on it.

gordonwoodhull:
I’ve resolved the circular dependencies and other build issues with reasonable elegance.

Since I’m out for the next week, I propose to merge on Dec 11.

Still need to clean up the commit history but I think this is solid.

We are hitting what appears to be an esbuild bug, necessitating marking some minified symbols as external for eslint. Seems to be harmless. https://github.com/evanw/esbuild/issues/4348

gordonwoodhull:
The only really ugly thing in here is the commands that need the list of engines:

```
quarto check
quarto call engine
quarto create-project
```

Often we need the list really early, when parsing command-line arguments.

Previously this was known statically, actually initialized at module level. But we can't do that anymore: we need a project context and we need to load extensions.

So there's a new utility `initializeProjectContextAndEngines` [in a new`src/command/command-utils.ts`](https://github.com/quarto-dev/quarto-cli/pull/13718/files#diff-d114f17533038c29becc37a54a1721bbdaeed443ee23d845111564f95f617b1e), which also invents a minimal `zeroFileProjectContext` to make this possible.

I tried to isolate the badness... we can revisit the implementation later, or feel free to review here!

PR #13747, opened 2025-12-01
claude: Make Windows smoke tests fail-fast like Linux

gordonwoodhull:
This is a small quality-of-life change to the Windows tests, that I've been using in my branch for a while.

I find it convenient to see the error immediately when checking results in CI, rather than it printing a bunch of successes afterward. (And possibly more failures, I acknowledge.)

I think I got distracted one evening and asked Claude if it could figure out why this is, and it diagnosed it as "failing fast" in Linux but not on Windows. So I tried its suggestion.

@cderv, your choice whether to take or leave it. Maybe there are other reasons I don't know for logging errors and then continuing?

### Description

When a test fails on Windows, exit immediately instead of continuing through remaining tests in the bucket. This matches Linux behavior (bash -e flag) and keeps test failures at the end of the log where they're easy to find, instead of buried under 1000+ lines of subsequent test output.

### DIFF
--- a/.github/workflows/test-smokes.yml
+++ b/.github/workflows/test-smokes.yml
@@ -276,23 +276,15 @@ jobs:
           # Useful as TinyTeX latest release is checked in run-test.sh
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          $haserror=$false
           foreach ($file in ('${{ inputs.buckets }}' | ConvertFrom-Json)) {
             Write-Host ">>> ./run-tests.ps1 ${file}"
             ./run-tests.ps1 $file
-            $status=$LASTEXITCODE
-            if ($status -eq 1) {
-              Write-Host ">>> Error found in test file"
-              $haserror=$true
-            } else {
-              Write-Host ">>> No error in this test file"
+            if ($LASTEXITCODE -ne 0) {
+              Exit 1
             }
+            Write-Host ">>> No error in this test file"
           }
-          if ($haserror) {
-            Exit 1
-          } else {
-            Write-Host ">>> All tests have passed"
-          }
+          Write-Host ">>> All tests have passed"
         working-directory: tests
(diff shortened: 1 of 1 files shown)


cderv (commented):
Thanks. 

Just to be clear, are linux tests failing faster ? 

I don't see exactly the same as claude saying:
> "failing fast" in Linux but not on Windows.

Code is 
https://github.com/quarto-dev/quarto-cli/blob/af0bb42f9c96a52ff44d18137df4ef1bd224525c/.github/workflows/test-smokes.yml#L260-L269

We do 
```
[ $status -eq 0 ] && echo ">>> No error in this test file" || haserror=1
```
so if failing tests, then we set `haserror` to 1 and we continue the loop

Then when done, we exit if there was one error
````
[ $status -eq 0 ] && echo ">>> No error in this test file" || haserror=1
````

Do I understand things wrongly? 

To me this PR makes Linux and Windows behave differently. 


We did Linux and Windows the same, by not failing fast **by design** for now. This is because that way we get ALL the tests that are failing on a buckets. And we can investigate locally all of them. 
Failing fast would mean, we get a failure in the bucket. We solve it locally, we trigger new CI by a commit and then we discover the next failure. It seemed to me not as convenient as knowing all the issues at once. 

Now with experience, maybe we don't have that much difference in failure each time, and in PR runs we could indeed fail fast, and then keep non fast fail when CI is running on main. 

Anyhow, we can change and fail fast if this is prefered. But we need to adapt Linux test too I believe. 

Let's discuss this.

cderv:
FWIW I am working on some improvement as we discussed (grouping and better summary) in #13787

gordonwoodhull:
As we discussed, the tests are currently written to keep going on error, but GHA executes `bash` with `-e` on Linux, so it stops immediately when any test returns a non-zero return code.

From that perspective, this isn't really the right fix, so I'll close it.

PR #13762, opened 2025-12-05
feat(lua-api): expose Typst path in Lua API

mcanouil:
Add `quarto.paths.typst()` to the Lua API to provide a consistent way to access the Typst binary path.
Include unit tests to verify the new functionality and update the changelog accordingly.

This change allows users to use Typst inside a filter or shortcode to process diagrams or other components in Typst.

### DIFF
--- a/tests/docs/smoke-all/2025/12/05/typst-path-api.qmd
+++ b/tests/docs/smoke-all/2025/12/05/typst-path-api.qmd
@@ -1,0 +1,12 @@
+---
+format: html
+shortcodes:
+  - typst-path-api.lua
+_quarto:
+  tests:
+    html:
+      ensureFileRegexMatches:
+        - ["<p>The Typst path is .+.</p>"]
+---
+
+The Typst path is "{{< typst >}}".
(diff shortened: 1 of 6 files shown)


cderv (approved):
Thanks! 

Can you create a companion doc PR in quarto-web if not already ?

tests/docs/smoke-all/2025/12/05/typst-path-api.qmd
cderv, line 1: Just a note that by doing this kind of testing in `typst-path-api.qmd`, it is not that minimal because we are also testing shortcode feature. So in a way it is good, but also this is not a test aimed to test shortcode and it would break if shortcode breaks. 

See the test for `rscript` for scoped test: Only the LUA api is tested which is what we want here.

https://github.com/quarto-dev/quarto-cli/blob/912352ece8117f3c4276c6e81f9f96feaf1bb849/tests/docs/smoke-all/2024/10/07/issue-10999.lua#L1-L5

part of #11000

cderv (approved):

tests/docs/smoke-all/2025/12/05/typst-path-api.qmd
mcanouil: fb2dd27174f08ac39297dceac86af462c9d20154

mcanouil:
> Can you create a companion doc PR in quarto-web if not already ?

- quarto-dev/quarto-web#1837

Was this the only place? (I did not find any other occurrences of `quarto.paths`, but I looked quickly)

PR #13763, opened 2025-12-06
feat(lua-api): add Lua API for retrieving Pandoc path

mcanouil:
Expose the path to the Pandoc binary through a new Lua API function `quarto.paths.pandoc()`.

This is a follow up of #13762

### DIFF
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -68,6 +68,10 @@ All changes included in 1.9:
 
 - ([#13414](https://github.com/quarto-dev/quarto-cli/issues/13414)): Be more forgiving when Confluence server returns malformed JSON response. (author: @m1no)
 
+## Lua API
+
+- Added `quarto.paths.pandoc()` to expose the path to the Pandoc binary that Quarto uses for rendering. This respects the `QUARTO_PANDOC` environment variable if set, or returns the path to the bundled Pandoc binary.
+
 ## Other fixes and improvements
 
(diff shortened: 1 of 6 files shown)


cderv (commented):
Thanks. 

Is there any use case behind this ? 

In LUA there is the Pandoc API already, so I don't see how the pandoc binary can be useful, or even prefered to using the Pandoc's LUA API. 

But maybe I am missing something. 

I share this concerns because exposing something means we think it could be useful. This is why sometimes we only do some modification when there is feature request or a use case behind. This avoid creating side effect or misusage of feature.

mcanouil:
Maybe I missed something that is covered by the Pandoc Lua API.
I need to double check as I did a lot of experiments. (switching to draft to avoid merging for now)

cderv:
> I need to double check as I did a lot of experiments.

Cool! Interested to know the use case behind using Pandoc binary inside lua filters. 

Calling Pandoc binary inside a Pandoc processing worries me, as when we do try to call quarto again with quarto. but maybe I am missing something obvious

mcanouil:
> Calling Pandoc binary inside a Pandoc processing worries me, as when we do try to call quarto again with quarto. but maybe I am missing something obvious

I'm more likely the one that missed something here, because I had the same thought on recursive calls.

mcanouil:
Closing as I don't find again the version (not on Git ...) of my experiment where it was useful.

PR #13771, opened 2025-12-09
fix(theme): `light-content` / `dark-content` SCSS rules not included in Reveal.js format

### LINKED ISSUE #13722: `light-content` / `dark-content` SCSS rule not included in Reveal.js format
### I have:

### Bug description

It appears, the SCSS from https://github.com/quarto-dev/quarto-cli/blob/45f556781e7bbe7fc80471d995d2d647dc941e50/src/resources/formats/revealjs/reveal/css/reveal.scss#L1538-L1549 has no effect and is not even included in rendered Reveal.js.

- https://github.com/quarto-dev/quarto-cli/pull/13168

### Steps to reproduce

<table>
<tr><th>Input</th><th>Output</th></tr>
<tr><td>

````qmd
---
title: "Quarto Playground"
format: revealjs
---

## Slide {.smaller}

[Light]{.light-content}

:::: {.light-content}
This is light content.
:::

[Dark]{.dark-content}

:::: {.dark-content}
This is dark content.
:::
````

</td><td>

Slide with a title 'slide' and two lines with 'Light' and 'Dark'

Slide with a title 'slide' and four lines with 'Light', 'This is light content.', 'Dark', and 'This is dark content.'

</td></tr>
</table>

### Actual behavior

Both spans/divs "Light" and "Dark" are displayed.

### Expected behavior

Only one should be displayed as with:

````qmd
---
title: "Quarto Playground"
format:
  revealjs:
    include-in-header:
      - text: |
          <style>
            body.quarto-light .dark-content {
              display: none !important;
            }

            body.quarto-dark .light-content {
              display: none !important;
            }
          </style>
---

## Slide {.smaller}

[Light]{.light-content}

:::: {.light-content}
This is light content.
:::

[Dark]{.dark-content}

:::: {.dark-content}
This is dark content.
:::
````

### Your environment

- VSCode
- Latest macOS

### Quarto check output

```bash
Quarto 99.9.9
      Pandoc version 3.6.3: OK
      Dart Sass version 1.87.0: OK
      NOTE: Deno version 2.4.5 does not strictly match 2.3.1 and strict checking is enabled. Please use 2.3.1.
      Typst version 0.13.0: OK
      Version: 99.9.9
      commit: 45f556781e7bbe7fc80471d995d2d647dc941e50
      Path: /Users/mcanouil/Projects/quarto-dev/quarto-cli/package/dist/bin

      TinyTeX: v2025.11
      Chromium: (not installed)

      Using: TinyTex
      Version: 2025

      Using: Chrome found on system
      Source: QUARTO_CHROMIUM

      Version: 3.14.0
      Path: /Users/mcanouil/Projects/quarto-dev/quarto-playground/.venv/bin/python
      Jupyter: 5.9.1
      Kernels: uv, julia-1.12, python3

(-) Checking R installation...........ℹ R version 4.5.2 (2025-10-31)
! Config '~/.Rprofile' was loaded!
      Version: 4.5.2
      LibPaths:
        - /Users/mcanouil/Projects/quarto-dev/quarto-playground/renv/library/macos/R-4.5/aarch64-apple-darwin20
        - /Users/mcanouil/Library/Caches/org.R-project.R/R/renv/sandbox/macos/R-4.5/aarch64-apple-darwin20/4cd76b74
      knitr: 1.50
      rmarkdown: 2.30

```

mcanouil:
Move the `light-content` / `dark-content` SCSS rules to the right SCCS file.

Fixes #13722


src/resources/formats/revealjs/reveal/css/reveal.scss
cderv, line 1: I can confirm that this file is the RevealJS source one that we update from upstream when we run our update logic. 

And `format: revealjs` will directly use the distribution `reveal.css`. We only use the theme layers in building theme for reveal using our SCSS layer system. 

So I guess this file could be ignored in our bundling, but we get it as we retrieve the whole revealjs source. 

Anyhow, rules should indeed not be in there. Thanks @mcanouil !

src/resources/formats/revealjs/quarto.scss
cderv, line 1: The fact we need to add this in this file, and that we have the exact same rules for HTML in 
https://github.com/quarto-dev/quarto-cli/blob/87f18c876759c619b0573105cf6ada7fab0fb6b3/src/resources/formats/html/bootstrap/dist/scss/_light-dark.scss#L1-L7

makes me think that we should be this rules in the Quarto base layer which is common to HTML format without bootstrap.
See for reveal
https://github.com/quarto-dev/quarto-cli/blob/87f18c876759c619b0573105cf6ada7fab0fb6b3/src/format/reveal/format-reveal-theme.ts#L175-L180
and see for HTML when no bootstrap
https://github.com/quarto-dev/quarto-cli/blob/87f18c876759c619b0573105cf6ada7fab0fb6b3/src/format/html/format-html.ts#L478-L490

This base layers is where rules for quarto and non bootstrap format are added

For Boostrap, the rules are usually added in https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/formats/html/bootstrap/_bootstrap-rules.scss

I see this `_light-dark.scss` does it differently though, but maybe there is a reason. 

I should document this theming logic we have currently... 

All that to say, I think this should go into the base layer **if this is supposed to also be something useful when no boostrap used**. Probably a test should be added with cell rendering, or just light / dark content and no bootstrap used (`theme: none` or `theme: pandoc`). 

Hope it makes sense.

news/changelog-1.9.md
```diff
@@ -48,6 +48,10 @@ All changes included in 1.9:
 - ([#13667](https://github.com/quarto-dev/quarto-cli/issues/13667)): Fix LaTeX compilation error with Python error output containing caret characters.
 - ([#13730](https://github.com/quarto-dev/quarto-cli/issues/13730)): Fix TinyTeX detection when `~/.TinyTeX/` directory exists without binaries. Quarto now verifies that the bin directory and tlmgr binary exist before reporting TinyTeX as available, allowing proper fallback to system PATH installations.
 
+### `revealjs`
+
+- ([#13722](https://github.com/quarto-dev/quarto-cli/issues/13722)): Improve handling of light and dark content in `revealjs` format. `light-content` and `dark-content` SCSS rules are now included in the proper SCSS file, ensuring proper styling for light and dark modes divs, mostly for multiformat documents. (author: @mcanouil)
```
cderv: > `light-content` and `dark-content` SCSS rules are now included in the proper SCSS file,

I believe changelog item should be user facing, and more about "this issue is fixed" than how it is fixed. 

Basically, it was not working previously, so handling of light and dark content in `revealjs` format is now working. This fixes cell rendering in revealjs format. 

Don't know if it makes sense; Just sharing as feedback: take it or leave it.

cderv (commented):
Some thoughts and context about how theming system works in Quarto and where it could be place. 

This PR needs testing too.

(and yes I know this is not marked for review, but I was in PR review mode and it was not draft 😅 )

src/resources/formats/revealjs/quarto.scss
mcanouil: When no bootstrap, there are no quarto-light/quarto-dark, so the rules have no effects.

mcanouil:
I've made the changes, but how can you test that compiled SCSS rules get included in the output? `ensureHtmlElements` only check the HTML and if `embed-ressources` is used, you still don't have a literal string to search in.

edit: I figured it out, there is a playwright testing suite.

src/resources/formats/revealjs/quarto.scss
cderv: We don't have boostrap in revealjs, and we have quarto-light and quarto-dark. 

So it does not mean we could had that `brand-mode` logic, or light dark support in some custom formats used without bootrap. The feature is not bootstrap dependant, otherwise we would have done it in revealjs right ?

tests/integration/playwright/tests/html-light-dark-content.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  const dark_spans = await page.locator('span.dark-content');
+
+  // Check that light content spans exist and are visible
+  await expect(light_spans.first()).toHaveCSS('display', /(block|inline)/);
+
+  // Check that dark content spans are hidden
+  await expect(dark_spans.first()).toHaveCSS('display', 'none');
+
+  const light_blocks = await page.locator('div.light-content').filter({ has: page.locator('blockquote, p, :not(div)') });
+  const dark_blocks = await page.locator('div.dark-content').filter({ has: page.locator('blockquote, p, :not(div)') });
```
cderv: This is not tested. 

Leftover or missing expectations ?

src/resources/formats/revealjs/quarto.scss
mcanouil: Current state is:

- Reveal.js does have `quarto-light` and `quarto-dark`.
- HTML + Bootstrap have `quarto-light` and `quarto-dark`.
- HTML + `theme: none` (or similar) does not have `quarto-light` and `quarto-dark`.

`brand-mode` is supported in Reveal.js which changes the class but only if there is branding, so `brand-mode` has no effect alone.

Disabling Bootstrap in HTML means no theme switching. (`brand-mode` is not a thing for HTML, and maybe it should be rather than adding a switch button)

Hope this clarifies as I'm not sure I quite get your question.

tests/integration/playwright/tests/html-light-dark-content.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    // Check that light-content and dark-content elements exist
+    const light_spans = await page.locator('span.light-content');
+    const dark_spans = await page.locator('span.dark-content');
+    const light_divs = await page.locator('div.light-content');
+    const dark_divs = await page.locator('div.dark-content');
+
+    await expect(light_spans).not.toHaveCount(0);
+    await expect(dark_spans).not.toHaveCount(0);
+    await expect(light_divs).not.toHaveCount(0);
+    await expect(dark_divs).not.toHaveCount(0);
```
cderv, line 46: FWIW We usually use `toBeVisible()` for testing existence of existence and visibility, and `toBeAttached()` if not visible. 

It is just a readibility matter I guess. I don't think those function do something differently, but in general, affirmation is easier than negation when reading code. 

Anyhow, this works as it.

tests/integration/playwright/tests/html-light-dark-content.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // Check that body has the correct theme class
+    const body = await page.locator('body').first();
+    const bodyClasses = await body.getAttribute('class');
+    expect(bodyClasses).toContain('quarto-light');
+
+    // Verify visibility based on theme
+    await expect(light_spans.first()).toHaveCSS('display', /(block|inline)/);
+    await expect(dark_spans.first()).toHaveCSS('display', 'none');
```
cderv: Though you are testing visibility after that. 

So unless you want to specifically test `display` classes, using `toBeVisible()` and `toBeHidden()` is usually better pattern for readability and intent. 

Sharing in case you didn't knew about them.

tests/integration/playwright/tests/html-light-dark-content.spec.ts
mcanouil: Are those helper testing functions documented somewhere? (I have to admit I did not go through all the playwright testing suite)

tests/integration/playwright/tests/html-light-dark-content.spec.ts
mcanouil: yes, I initially thought for testing all cases I could think of, but this was a bit of a rabbit hole.

tests/integration/playwright/tests/html-light-dark-content.spec.ts
mcanouil: > Sharing in case you didn't knew about them.

I did not (https://github.com/quarto-dev/quarto-cli/pull/13771#discussion_r2610005071)
I'll refactor to use these helpers as it's way simpler.

tests/integration/playwright/tests/html-light-dark-content.spec.ts
cderv: Documentation is here: 
- https://playwright.dev/docs/test-assertions#auto-retrying-assertions
- https://playwright.dev/docs/api/class-locatorassertions

Usually google search is helpful to find the right function. 

Maybe now with claude it can be easier. I see a claude skills: https://github.com/lackeyjb/playwright-skill 

But I haven't tried it. 

Anyhow, this comes with time as often the function are the same. And most useful are shared in best practice guide like: https://playwright.dev/docs/best-practices#use-web-first-assertions

Also, you have way to have your test be written for you: https://playwright.dev/docs/getting-started-vscode#generating-tests-with-codegen

No the best code, but quite helpful to get the structure or locators.

mcanouil: 9c690688329c24e310a3b564d5d07cae9bf5b99d

tests/integration/playwright/tests/html-light-dark-content.spec.ts
mcanouil: 9c690688329c24e310a3b564d5d07cae9bf5b99d

This was way easier with those functions ahah

tests/integration/playwright/tests/html-light-dark-content.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    await expect(dark_spans).not.toHaveCount(0);
+    await expect(light_divs).not.toHaveCount(0);
+    await expect(dark_divs).not.toHaveCount(0);
+
+    // Check that body has the correct theme class
+    const body = page.locator('body');
+    await expect(body).toHaveClass(/quarto-light/);
+
+    // Verify visibility based on theme
+    await expect(light_spans.first()).toBeVisible();
```
cderv, line 53: Just because I am in doubt, why are we only testing the first ? 🤔 
I think I saw this pattern in playwright doc, but unsure.

tests/integration/playwright/tests/html-light-dark-content.spec.ts
cderv: Great thanks ! 

I believe this does not harm but `.not.toHaveCount(0);` is no more useful if you are checking for visible or hidden after. At least, this is what I understand 🤷‍♂️

tests/integration/playwright/tests/html-light-dark-content.spec.ts
mcanouil: It's from strictness check from playwright when multiple occurrences are found: https://playwright.dev/docs/locators#strictness

I also thought it makes sense as the rule applies to everything or to nothing, testing all occurrences is a waste of resources.
It's also me being a bit lazy: I could change the example and the test to have more specific selector thus avoiding to opt-out strictness check/test.

From my understanding of the playwright documentation, the alternative to test all occurrences is to do something like:

```ts
const light_spans = page.locator('span.light-content');
await expect(light_spans).toHaveCount(3);
await expect(light_spans.nth(0)).toBeVisible();
await expect(light_spans.nth(1)).toBeVisible();
await expect(light_spans.nth(2)).toBeVisible();
```

I'm not a Playwright expert, I'm using it for the very first time here.

cderv: Ok make sense here I understand. i'll remember this `first()` usage. Thanks !

PR #13787, opened 2025-12-11
test: enhance error reporting and grouping in test execution inside Github Action

cderv:
This PR aims at improving readability of CI logs on failures. Idea is to navigate more easily to failing tests, and also have Windows and Linux behave the same. Difference in behavior is about the default for shell execution on error. See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#exit-codes-and-error-action-preference for details (basically CMD does not fail fast by default unlike bash and pwsh shells on github action)


cderv:
Triggering a test with the failing one in the middle to see if we can navigate to it easily: https://github.com/quarto-dev/quarto-cli/actions/runs/20174970499

cderv:
New run at https://github.com/quarto-dev/quarto-cli/actions/runs/20226090577

cderv:
So this is what we see now
image
And we can unfold what we want to see 
image
And at the end we have a summary of failing tests
image

This is the same on Windows. 

See detailed log at [quarto-dev/quarto-cli/actions/runs/20226090577](https://github.com/quarto-dev/quarto-cli/actions/runs/20226090577)

PR #13790, opened 2025-12-12
detect and warn about x64 R on Windows ARM

cderv:
When x64 R fails on Windows ARM, parse YAML output to detect the architecture mismatch and provide a specific error message with ARM64 R download links instead of the generic "check your R installation" message.

Also adds platform information to quarto check output with warning when x64 R is detected on ARM Windows.

This is draft PR because I need to check this work as expected in windows ARM runner

### DIFF
--- a/src/core/knitr.ts
+++ b/src/core/knitr.ts
@@ -82,7 +82,7 @@ function checkWindowsArmR(platform: string | undefined): void {
 
   const isX64R = platform.includes("x86_64") || platform.includes("i386");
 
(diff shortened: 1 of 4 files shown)


cderv:
As discussed, I think we'll start with simpler than trying to detect Windows ARM from deno x64 using a dll for Windows API. #13822 is about simplifying by checking error code.

cderv:
This is closed in favor of #13822 

But we'll keep the branch

PR #13804, opened 2025-12-16
Manage projectcontext objects correctly on preview

cscheid:
This should close a few of our `quarto preview` intermittent issues whose symptom is a missing `quarto_ipynb` file.

See, eg, #12992.

### DIFF
--- a/package/scripts/common/quarto
+++ b/package/scripts/common/quarto
@@ -187,9 +187,9 @@ QUARTO_DENO_OPTIONS="--unstable-ffi --unstable-kv --no-config --no-lock ${QUARTO
 
 # --enable-experimental-regexp-engine is required for /regex/l, https://github.com/quarto-dev/quarto-cli/issues/9737
 if [ "$QUARTO_DENO_V8_OPTIONS" != "" ]; then
-  QUARTO_DENO_V8_OPTIONS="--enable-experimental-regexp-engine,--max-old-space-size=8192,--max-heap-size=8192,${QUARTO_DENO_V8_OPTIONS}"
+  QUARTO_DENO_V8_OPTIONS="--enable-experimental-regexp-engine,--max-old-space-size=8192,--max-heap-size=8192,--stack-trace-limit=100,${QUARTO_DENO_V8_OPTIONS}"
 else
-  QUARTO_DENO_V8_OPTIONS="--enable-experimental-regexp-engine,--max-old-space-size=8192,--max-heap-size=8192"
+  QUARTO_DENO_V8_OPTIONS="--enable-experimental-regexp-engine,--max-old-space-size=8192,--max-heap-size=8192,--stack-trace-limit=100"
 fi
 
 if [ "$QUARTO_DENO_EXTRA_OPTIONS" == "" ]; then
--- a/src/command/render/render-shared.ts
+++ b/src/command/render/render-shared.ts
@@ -33,20 +33,23 @@ import { kTextPlain } from "../../core/mime.ts";
 import { normalizePath } from "../../core/path.ts";
 import { notebookContext } from "../../render/notebook/notebook-context.ts";
 import { singleFileProjectContext } from "../../project/types/single-file/single-file.ts";
-import { assert } from "testing/asserts";
+import { ProjectContext } from "../../project/types.ts";
 
 export async function render(
   path: string,
   options: RenderOptions,
+  pContext?: ProjectContext,
 ): Promise<RenderResult> {
   // one time initialization of yaml validators
   setInitializer(initYamlIntelligenceResourcesFromFilesystem);
   await initState();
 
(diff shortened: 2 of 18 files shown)


cscheid:
This last failing test is interesting. If I'm reading the messages correctly, we previously had:

- a test that used to pass but which requested invalid options
- a code path that verified the validity of options that wasn't getting called because of incorrect plumbing of the ProjectContext object

Now that things are properly connected, the test fails. But it seems that this test shouldn't really even be there?

package/scripts/common/quarto
```diff
@@ -187,9 +187,9 @@ QUARTO_DENO_OPTIONS="--unstable-ffi --unstable-kv --no-config --no-lock ${QUARTO
 
 # --enable-experimental-regexp-engine is required for /regex/l, https://github.com/quarto-dev/quarto-cli/issues/9737
 if [ "$QUARTO_DENO_V8_OPTIONS" != "" ]; then
-  QUARTO_DENO_V8_OPTIONS="--enable-experimental-regexp-engine,--max-old-space-size=8192,--max-heap-size=8192,${QUARTO_DENO_V8_OPTIONS}"
+  QUARTO_DENO_V8_OPTIONS="--enable-experimental-regexp-engine,--max-old-space-size=8192,--max-heap-size=8192,--stack-trace-limit=100,${QUARTO_DENO_V8_OPTIONS}"
```
cderv, line 190: Why do we need this flag ? 

Just to understand. Because we probably need to do the same in `quarto.cmd` for the windows version to keep both scripts in sync.

src/command/render/render-shared.ts
```diff
@@ -33,20 +33,23 @@ import { kTextPlain } from "../../core/mime.ts";
...
   await initState();
 
-  const nbContext = notebookContext();
+  const nbContext = pContext?.notebookContext || notebookContext();
 
   // determine target context/files
-  let context = await projectContext(path, nbContext, options);
+  // let context = await projectContext(path, nbContext, options);
+  let context = pContext || (await projectContext(path, nbContext, options)) ||
+    (await singleFileProjectContext(path, nbContext, options));
```
cderv: This is the change that makes a test fail. Before we add `context = null` possible, which lead to the `if (!context && options.flags?.outputDir) ` for synthetic project creation. 

Now this does not go there, and `--output-dir` can't work with `singleFileProjectContext()`

Is this expected to fix the preview problem that we fallback to singleFileProject here ? 
Asking to know what is the fix we should do here: just reorder the logic and fallback, or still allow `context = null` ?

cderv (commented):
The test failure in `render-output-dir.test.ts` (and the other playwright `docs/playwright/embed-resources/issue-11860/main.qmd`) is caused by a logic ordering issue at:

https://github.com/quarto-dev/quarto-cli/blob/e48007ddc1ad14ca997d8aaa4beeb6ab5533fb77/src/command/render/render-shared.ts#L51-L63

The change on lines 51-52 now always creates a singleFileProjectContext as fallback:
```ts
let context = pContext || (await projectContext(path, nbContext, options)) ||
(await singleFileProjectContext(path, nbContext, options));
```

This means context is never null when the --output-dir check runs on line 57:
```
if (!context && options.flags?.outputDir) {
```

The synthetic project creation for `--output-dir` is now unreachable. The fix could be to check for `--output-dir` before falling back to `singleFileProjectContext`, if we really need this fallback. I am unsure about that and its relation to preview fix which is the topic of this PR. 

For example

```diff
diff --git a/src/command/render/render-shared.ts b/src/command/render/render-shared.ts
--- a/src/command/render/render-shared.ts
+++ b/src/command/render/render-shared.ts
@@ -48,9 +48,7 @@ export async function render(
   const nbContext = pContext?.notebookContext || notebookContext();

   // determine target context/files
-  // let context = await projectContext(path, nbContext, options);
-  let context = pContext || (await projectContext(path, nbContext, options)) ||
-    (await singleFileProjectContext(path, nbContext, options));
+  let context = pContext || (await projectContext(path, nbContext, options));

   // Create a synthetic project when --output-dir is used without a project file
   // This creates a temporary .quarto directory to manage the render, which must
@@ -62,6 +60,11 @@ export async function render(
     options.forceClean = options.flags.clean !== false;
   }

+  // Only fall back to singleFileProjectContext if we don't have a context yet
+  if (!context) {
+    context = await singleFileProjectContext(path, nbContext, options);
+  }
+
   // set env var if requested
   if (context && options.setProjectDir) {
     // FIXME we can't set environment variables like this with asyncs flying around
```

package/scripts/common/quarto
cscheid: It helps with debugging and shouldn't otherwise hurt. It's not important.

cderv: OK. I'll add it. Otherwise we don't know why both script are differents.

PR #13807, opened 2025-12-17
GHA TESTS - Improve logging to group in the right place 

cderv:
This PR follows up previous one #13787 by trying to group in a more scoped place.


cderv:
Here is the new state

# Playwright tests logging 

All test is grouped, but failure is shown outside the group

image

Then inside the group we get logging about each document render in pre-test step
image

Then the playwright logging itself is at its minimum (dot reporter) with error shown after 
image

image


# smoke test logging

Each test is grouped  and error shows below the failing grouped test with summary at the end
image

Unfolding shows the details
image

Same for smoke-all document
image


Still small tweaks but this is better. #13808 will also help reduce the Deno logging about dependencies download.

PR #13824, opened 2025-12-22
properly filter bundled extensions from engine list using normalized path

gordonwoodhull:
* Filter normalized paths for printing
* Look for uniform `resources/extension-subtrees/` path
* Test

Fixes #13819


cderv (commented):
FWIW we have some path helpers to handle those Windows / Linux differences. 

One of them is about using forward slash. 

https://github.com/quarto-dev/quarto-cli/blob/b323477b133bb12222bafb2c2c213b012dc253c5/src/core/path.ts#L189-L191

Just sharing for next time. And `path.ts` has other helpers like this normalizePath one 
https://github.com/quarto-dev/quarto-cli/blob/b323477b133bb12222bafb2c2c213b012dc253c5/src/core/path.ts#L304-L318

PR #13828, opened 2025-12-24
feature: `quarto use brand`

gordonwoodhull:
Slightly different pattern from other commands:
* `--dry-run`: show every operation that would happen (overwrite, copy, remove)
*  `--force`: skip and confirm all prompts

`--no-prompt` was too ambiguous because `quarto use template` does not allow actions; `--force` is really clear.

In addition to copying and overwriting, the commands will also remove any extra files that were already in the `_brand/` folder. This also invokes a prompt listing those files; saying no to removal does not cancel the rest of the command. 

`--dry-run` and `--force` also report/force potential removals.


gordonwoodhull:
Note: I'm breaking with tradition by introducing new commands under 

```
## Commands
```

in the changelog. Previously we only put updates to commands here, but I think new commands should be at least as prominent as changes to commands and not buried.

This is partly because I didn't want to hide `quarto use brand` and partly because I didn't know where else to put it.

I also reordered sections to be more consistent with previous changelogs.

Glad to discuss and rearrange.

gordonwoodhull:
I'm going ahead and releasing this.

`quarto use brand` on christmas day, why not.

Share the brands, share the cheer!

PR #6784, opened 2023-09-09
fix: Add missing CC licences and display CC code instead of URL

mcanouil:
This PR add the missing CC licenses and also use the CC code as display test instead of the license URL.

image

Fixes #6783

PR #13832, opened 2025-12-28
fix(license.lua): correct license metadata lua processing

mcanouil:
Correct the processing of license metadata to ensure proper handling of inline license elements.

The typo prevented from using `license.text` in template/partials/shortcodes when `license: "my license"` was used.

Note that version is not added to "text" for CC0 licenses because for some reason, the URL/parsing is created in TypeScript for this license.

```qmd
---
format: html
license: "Whatever"
---

{{< meta license.text >}}
```

```qmd
---
format: html
license: "CC BY"
---

{{< meta license.text >}}
```

```qmd
---
format: html
license: "CC0"
---

{{< meta license.text >}}
```

### DIFF
--- a/tests/docs/smoke-all/2025/12/28/license-CC0.qmd
+++ b/tests/docs/smoke-all/2025/12/28/license-CC0.qmd
@@ -1,0 +1,12 @@
+---
+format: html
+license: "CC0"
+_quarto:
+  tests:
+    html:
+      ensureFileRegexMatches:
+        - 
+          - "License: CC0"
+---
+
+License: {{< meta license.text >}}
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -103,3 +103,4 @@ All changes included in 1.9:
 - ([#13528](https://github.com/quarto-dev/quarto-cli/pull/13528)): Adds support for table specification using nested lists and the `list-table` class.
 - ([#13575](https://github.com/quarto-dev/quarto-cli/pull/13575)): Improve CPU architecture detection/reporting in macOS to allow quarto to run in virtualized environments such as OpenAI's `codex`.
 - ([#13656](https://github.com/quarto-dev/quarto-cli/issues/13656)): Fix R code cells with empty `lang: ""` option producing invalid markdown class attributes.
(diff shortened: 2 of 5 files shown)


cderv (approved):
Thank you for the change! 

Honestly, I don't think this is a typo - text was just not set initially. 

Then it was added for the typescript part in https://github.com/quarto-dev/quarto-cli/pull/6784. 

So this change is an improvement to make inline CC license having a default text, exactly like with the appendix part. 

This is good ! Another oversight is the the support for CC0 that was added only on the TS part too: https://github.com/quarto-dev/quarto-cli/pull/10763

````
---
format: html
license: "CC0"
---

License: {{< meta license.text >}}
````

so it won't be `CC0 1.0` like the link 
image

Also, I checked the template for which those meta are used. I think this is fine to add such text. Only JATS seems to use this licence in template. And `<license-p>` seems to support this text for CC too https://jats.nlm.nih.gov/publishing/tag-library/1.3/element/license-p.html

mcanouil:
FWIW, the context was for my Typst template, I wanted to show license/copyright, but this prevented it and generally it prevented anyone to use it as part of their template.

cderv:
Thanks for the context, that it helpful to understand ! I'll merge this today

mcanouil:
And yes, CC0 is not accounted for as it would require more thinking and possibly refactoring to move it to Lua with other license processing. Maybe worth opening an enhancement issue for "future"?

mcanouil:
The typo I was referring to is (the brackets should not be there):

- https://github.com/quarto-dev/quarto-cli/pull/13832/changes#diff-42d0469b5f06355473cc5c74a61ee7c1888ef1a7423ace4a3bbc3d3439b8bd2aL162

The other changes are only for CC licenses.

tests/docs/smoke-all/2025/12/28/license-CC0.qmd
cderv, line 1: Just a note that as I mentioned CC0 support has not been added yet for this LUA processing. So this is testing the same thing as `license: "Whatever"` 

So I think I'll remove this test until we do implement the CC0 1.0 support

mcanouil: In this case, we need to open an issue to not forget about this.

PR #13847, opened 2026-01-05
fix(website-meta.ts): change meta selector from "name" to "property"

mcanouil:
This pull request fixes a typo. Now the code correctly targets the correct `property` attribute instead of the `name` attribute for `og:site_name` meta tags.


cderv:
Thanks! Good find ! 

Can you add a test or a check for this in an existing test for shortcode in title ? 

Thanks.

news/changelog-1.9.md
```diff
@@ -62,6 +62,7 @@ All changes included in 1.9:
 - ([#13547](https://github.com/quarto-dev/quarto-cli/issues/13547))`cookie-content: { type: express }` is now the default. Previously it was `type: implied`. It now means this will block cookies until the user expressly agrees to allow them (or continue blocking them if the user doesn't agree).
 - ([#13570](https://github.com/quarto-dev/quarto-cli/pull/13570)): Replace Twitter with Bluesky in default blog template and documentation examples. New blog projects now include Bluesky social links instead of Twitter.
 - ([#13716](https://github.com/quarto-dev/quarto-cli/issues/13716)): Fix draft pages showing blank during preview when pre-render scripts are configured.
+- ([#13847](https://github.com/quarto-dev/quarto-cli/pull/13847)): Correctly targets the `property` attribute instead of the `name` attribute for `og:site_name` meta tags. (author: @mcanouil)
```
cderv: Thanks for the changelog item. 

Usually, we aim at changelog to be user facing and not just a list of technical change. 
So when possible, it is best to have an entry that describe what the change fixed for user. 

Usually, this is easy as this is the original issue problem. Here you just have a PR and not issue, but the problem you shared was: Title is wrong in Open Graph Card for title. 

I think it would be better if the entry explain this instead of just being a description of the commit change. A user reading thing won't really know what this is about. 

What do you think ?

mcanouil: Slightly edited the entry.

> but the problem you shared was: Title is wrong in Open Graph Card for title.

The correct statement would be: open graph title was not processed as markdown as it should have because of the wrong selector being used.

cderv: > open graph title was not processed as markdown

Good root cause. Usually I try to not explain the "because" especially if this is seen in the commit itself. I try to stay user facing with no technical detail on implementation or else. Habits of following [Tidyverse Style guide](https://style.tidyverse.org/news.html) I guess

mcanouil: I'm too much in technical changelogs recently^^

mcanouil:
> Can you add a test or a check for this in an existing test for shortcode in title ?

Cannot run any test at the moment.

- [5fe4471](https://github.com/quarto-dev/quarto-cli/pull/13847/commits/5fe44713fd7748f6145e9994e11d674e31328f2e)

tests/docs/smoke-all/2026/01/06/13847/index.qmd
```diff
@@ -0,0 +1,11 @@
+---
+title: "PR 13847: open graph metadata"
+_quarto:
+  render-project: true
+  tests:
+    html:
+      ensureFileRegexMatches:
+        - 
+          - '\<meta property="og:site_name" content="PR 13847: open graph metadata – Quarto CLI 1.0.0"\>'
+        - []
```
cderv: I think it is best to use selectors when we can instead of regex. 

This is a DOM node it should be possible to check the content attribute either for no `{{<` or for the expected content.

cderv (approved):

tests/docs/smoke-all/2026/01/06/13847/index.qmd
cderv: I did the change. When tests passes, I'll merge

mcanouil: Thanks

PR #13857, opened 2026-01-08
feature: cross-format `pdf-standard` selection

### LINKED ISSUE #4426: Support PDF/A-compliant output using pdfx package
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/4424

<sup>Originally posted by **fuhrmanator** February 18, 2023</sup>
My publishing requirements ask me to generate a PDF that is PDF/A (archive) compliant. I did some googling, and found a great explanation at <https://www.mathstat.dal.ca/~selinger/pdfa/>. However, I wasn't able to get the metadata into the PDF file according to the documented process. I realized it's because it's hard to know exactly what the `.tex` file is named when `pdflatex` is invoked in Quarto. In my case, it's a book project, so I think the file was `index.tex`, even though the `keep-tex` results in a file with the name of the title of my book. I suppose it's something else when you write an article, etc.

Anyway, I read the docs for the `pdfx` package and found that you can embed the metadata into the `.tex` source, with a vanilla file name that "always" works. I tried it using `header-includes:` and it works well (or seems to). I see the metadata in the PDF, and Acrobat says my PDF is indeed PDF/A when I open it. I haven't fully validated the result with a real validation tool. But it's an encouraging start. Here's what I put in my `_quarto.yml` for a book project.

```yaml
format:
  pdf:
    header-includes: |
      \usepackage[a-1b]{pdfx}
      \begin{filecontents*}[overwrite]{\jobname.xmpdata}
      \Title{Baking through the ages}
      \Author{A. Baker\sep C. Kneader}
      \Language{en-GB}
      \Keywords{cookies\sep muffins\sep cakes}
      \Publisher{Baking International}
      \end{filecontents*}
```

The first line loads the `pdfx` package, and the following lines create (overwrite) the metadata file, which is magically created temporarily in the `pdflatex` pipeline (so you don't need to worry about the exact filename, and it stays with your document). 

Of course, you must duplicate some of the XMP info that is defined in other Quarto fields, e.g., Title, Author, etc. Maybe one day Quarto can make this happen automagically (in partials), but for now I'm happy I don't have to do a separate conversion step.</div>

gordonwoodhull:
Fixes #4426 

Also includes
* fix for #13248 for alt text from captions for PDF images not being passed through to Pandoc. 
* temporary fix for #13868 until jgm/pandoc#11394 is merged, byte-for-byte identical

Adds basic support in Typst and LaTeX for PDF standard selection. 

PDF versions and standards can be independently specified, so this takes a list. Version specifies the output format; standard specified what validation is applied to the output.

Here are the rules: you can combine standards from different families:
- PDF/A (archival) + PDF/UA (accessibility) is the common combo
- PDF version can also be included

The constraint is you can only have one from each category:
- One PDF version (1.4, 1.5, 1.6, 1.7, 2.0)
- One PDF/A standard (a-1b, a-2b, etc.)
- One PDF/UA standard (ua-1, ua-2)
- One PDF/X standard (x-4, etc.)

It would be equally valid to take `pdf-version` and `pdf-standard` as separate options, as LaTeX does, but I chose to follow Typst here and use one option for both.

Once the options have been validated, these go to the command line `--pdf-standard` for Typst, new `\DocumentMetadata{}` command for PDF.

The metadata includes `tagging=on` for any standard in LaTeX.

Auto-installs required LaTeX packages (latex-lab, colorprofiles) when missing.

This includes a test which will fail UA-1 in Typst and warn in LaTeX; these warnings are surfaced from the LaTeX log.


src/resources/formats/typst/pandoc/quarto/page.typ
```diff
@@ -2,6 +2,7 @@
   paper: $if(papersize)$"$papersize$"$else$"us-letter"$endif$,
   margin: $if(margin)$($for(margin/pairs)$$margin.key$: $margin.value$,$endfor$)$else$(x: 1.25in, y: 1.25in)$endif$,
   numbering: $if(page-numbering)$"$page-numbering$"$else$none$endif$,
+  columns: $if(columns)$$columns$$else$1$endif$,
```
mcanouil, line 11: Beware of this with the pagebreak shortcode.
The shortcode is not conditional and uses `pagebreak()`. Using columns, means to use `colbreak()` (same for margins as I believe they also are containers).

I ended up defining a function and some context variables (source: https://github.com/mcanouil/quarto-mcanouil/tree/feat/typst-template/_extensions/mcanouil/typst - FYI it's all under MIT licence):

```typst
/// Apply appropriate page or column break based on layout context
/// Uses state tracking to determine if columns are active.
/// Uses colbreak when inside columns/containers, pagebreak otherwise.
/// @param weak Whether to use weak break (default: true)
/// @return Content with the appropriate break
#let conditional-break(weak: true) = context {
  let in-columns = columns-active-state.get()

  if in-columns {
    colbreak(weak: weak)
  } else {
    pagebreak(weak: weak)
  }
}
```

```typst
/// State to track whether columns are currently active
/// Set to true when multi-column layout is enabled via set page(columns: ...)
#let columns-active-state = state("columns-active", false)
```

In `typst-show.typ` (don't min the comments, in my case columns are set after title page and all outlines/table of X):

```typst
// This ensures outlines render in single column
$if(columns)$
// Update state to indicate if columns are active (true if > 1)
#columns-active-state.update($columns$ > 1)
// Set page columns (even for columns: 1, which is equivalent to single column)
#set page(columns: $columns$$if(column-gutter)$, column-gutter: $column-gutter$$endif$)
$endif$
```

mcanouil: Unfortunately, `conditional-break()` is still not fully aware of the context in which it is called and as far as my understanding goes, there is currently no way for that at the moment in Typst (probably why, native Typst has two functions in the first place)

gordonwoodhull: Thanks @mcanouil! As you may know, this change comes from upstream jgm/pandoc@cfc5ac447, where it's a regression fix. 

As such, if we want to do what you're suggesting, we should handle it in the parent PR #13249

I'll have to stare at this and think about it a bit more to understand it fully. It seems to me that as written above, this would be equivalent to defining `conditional-break` directly based on `columns`, unless there is somewhere else that changes `columns-active-state`? Because we wouldn't be using breaks outside of the body text anyway.

I've changed the base branch to make it more clear what changes are in this PR.

mcanouil: I'm mostly highlighting the col/pagebreak issue as it happens to users regularly already.
For example, with `layout*`, my idea of the state variable is that layout or other container structure could change it via Lua.
With "columns", the issue will be more frequent I believe.

For short and nothing more, we need to keep in mind that the current implementation of the pagebreak shortcode might lead to issue for Quarto users.

---

I also noticed that using `columns()` (Pandoc template) leads to more issues than setting `page(columns: ...)`, but this would be also for upstream.

mcanouil: Simple Typst example to show what I mean about columns() vs. page.columns

```typst
#columns(
  2,
  [
    #text("This is the first column. It contains some sample text to demonstrate how columns work in Typst."),
    #colbreak(), // check the difference with #pagebreak()
    #text("This is the second column. It also contains some sample text to illustrate the column layout feature.")
  ],
)
```

```typst
#set page(columns: 2)

This is the first column. It contains some sample text to demonstrate how page columns work in Typst.

#pagebreak() // check the difference with #colbreak()

This is the second column. It also contains some sample text to illustrate the page column layout feature.
```

mcanouil: page.columns won't lead to crash with the shortcode as currently implemented, but maybe not what users might expect.
At least, it keeps flexibility to change column or change page, which is not possible with columns().

mcanouil: That's not new and the `columns()` use breaks already many things:

> error: page configuration is not allowed inside of containers

````qmd
---
format:
  typst:
    columns: 2
---

{{< lipsum 1 >}}

::: {.landscape}

{{< lipsum 1 >}}

:::
````

````qmd
---
format: typst
---

```{=typst}
#set page(columns: 2)
```

{{< lipsum 1 >}}

::: {.landscape}

{{< lipsum 1 >}}

:::
````

Comment (gordonwoodhull) [reply src/resources/formats/typst/pandoc/quarto/page.typ:11]: Got it. Pandoc moved from `columns(cols, doc)` to `#set page(columns)` in Dec 2024 and we missed it. Tests and fixes merged to #13249.

**Comment by cderv** [comment]
Following our meeting discussion, related issues on pdf accessibility 
- Pandoc: https://github.com/jgm/pandoc/issues/5409
- https://github.com/quarto-dev/quarto-cli/issues/4426

**Comment by gordonwoodhull** [comment]
Clarification: LaTeX does not do compliance validation; that's why it only produces warnings.

We'd have to integrate something like veraPDF to actually validate LaTeX PDFs.

Compliance validation is integrated for Typst.

**Comment by cderv** [comment]
Probably closing: 
- https://github.com/quarto-dev/quarto-cli/pull/13124

File: src/tools/impl/verapdf.ts
Comment (cderv) [inline src/tools/impl/verapdf.ts:1]: It does not seem to work on windows 
````
❯ ./package/dist/bin/quarto.cmd install
Check file:///C:/Users/chris/Documents/DEV_R/quarto-cli.worktrees/issue-13857/src/quarto.ts
[>] Inspecting tools
? Select a tool to install » VeraPDF (1.28.2)
````

Nothing happens - no additional message. I would have expected an issue about JAVA.

File: src/tools/impl/verapdf.ts [hunk 2]
```diff
@@ -0,0 +1,283 @@
...
+  prereqs: [{
+    check: async () => {
+      const javaVersion = await getJavaVersion();
+      return javaVersion !== undefined &&
+        kSupportedJavaVersions.includes(javaVersion);
+    },
+    os: ["darwin", "linux", "windows"],
+    message:
+      `Java is not installed or version is not supported. veraPDF requires Java 8, 11, 17, or 21.
+Please install a compatible Java version from https://adoptium.net/`,
```
cderv: is this recommended to install a JAVA from this specific place for veraPDF to work ? 

It looks as if we have a reason to push toward using this provide. JAVA can be install in many different way, and I wonder why we would push toward adoptium. 🤔

src/tools/impl/verapdf.ts
```diff
@@ -0,0 +1,283 @@
...
+}
+
+function verapdfInstallDir(): string {
+  return quartoDataDir("verapdf");
+}
+
+async function installed(): Promise<boolean> {
+  const dir = verapdfInstallDir();
+  const verapdfBin = isWindows
+    ? join(dir, "verapdf.bat")
```
cderv, line 87: Just a heads up that we know .bat file are a problem in some environment. We had some report about dart-sass not working inside Quarto (because this is a bat file). 

So a limitation to know about with this veraPdf, though quite specific.

cderv (commented):
Some quick comments for now as I tried veraPDF install on windows

src/tools/impl/verapdf.ts
gordonwoodhull: I know it looks like a scam site but it’s apparently the official OpenJDK.

Even so, we should probably not suggest any particular Java. It should not matter which one people install. 

And I’m sure companies have policies about this…

src/tools/impl/verapdf.ts
gordonwoodhull: Ok let’s talk about the right way. We can probably launch directly. Hopefully we don’t need another Rust launcher. 😉

src/tools/impl/verapdf.ts
cderv: It seems to be a cliffy issue as passing the arg works 
````
❯ ./package/dist/bin/quarto.cmd install veraPDF
Installing veraPDF
Java is not installed or version is not supported. veraPDF requires Java 8, 11, 17, or 21.
Please install a compatible Java version from https://adoptium.net/
ERROR: [non-error-thrown] undefined

Stack trace:
    at Command.handleError (https://deno.land/x/cliffy@v1.0.0-rc.3/command/command.ts:2202:11)
    at Command.parseCommand (https://deno.land/x/cliffy@v1.0.0-rc.3/command/command.ts:1782:12)
    at eventLoopTick (ext:core/01_core.js:179:7)
    at async quarto (file:///C:/Users/chris/Documents/DEV_R/quarto-cli.worktrees/issue-13857/src/quarto.ts:196:5)
    at async file:///C:/Users/chris/Documents/DEV_R/quarto-cli.worktrees/issue-13857/src/quarto.ts:228:5
    at async file:///C:/Users/chris/Documents/DEV_R/quarto-cli.worktrees/issue-13857/src/core/main.ts:45:14
    at async mainRunner (file:///C:/Users/chris/Documents/DEV_R/quarto-cli.worktrees/issue-13857/src/core/main.ts:47:5)
    at async file:///C:/Users/chris/Documents/DEV_R/quarto-cli.worktrees/issue-13857/src/quarto.ts:218:3
````

Though we should probably not error like that.

Comment (gordonwoodhull) [reply src/tools/impl/verapdf.ts:1]: Yeah I only tested it with the argument. Good catch!

Comment (cderv) [reply src/tools/impl/verapdf.ts:1]: This is a broader issue in fact. I think we encountered it already in the past. Nothing to do with veraPDF I get the same for chromium.

Comment (gordonwoodhull) [reply src/tools/impl/verapdf.ts:1]: Certainly worth fixing though!

File: src/resources/filters/quarto-post/typst.lua [hunk 4]
```diff
@@ -94,6 +94,48 @@ function render_typst_fixups()
       if image.attributes["width"] ~= nil and type(width_as_number) == "number" then
         image.attributes["width"] = tostring(image.attributes["width"] / PANDOC_WRITER_OPTIONS.dpi) .. "in"
       end
+
+      -- Workaround for Pandoc not passing alt text to Typst image() calls
+      -- See: https://github.com/jgm/pandoc/issues/XXXX (TODO: file upstream)
```
cscheid: Do we have a concrete number for this one already available?

cscheid (approved):
Looks great aside from one small question on a comment.

src/resources/filters/quarto-post/typst.lua
gordonwoodhull: Hmm, looks like a miscommunication, will correct.

This is our #13868; @mcanouil has filed jgm/pandoc#11394 upstream to address this.

The commit in this PR is character-for-character equivalent to the fix upstream. (Unless moving the mediabag processing to our side changes anything... probably worth recomparing.)

src/tools/impl/verapdf.ts
cderv: Yes this is fixed by https://github.com/quarto-dev/quarto-cli/pull/13935

src/tools/impl/verapdf.ts
cderv: > I know it looks like a scam site but it’s apparently the official OpenJDK.

I would have expected this one for official openJDK ? https://openjdk.org/

But the one at adoptium looks nice though 🤔 

Anyhow, on windows one can install with winget or scoop 

`winget search openjdk` as A LOT of results ! 

Using java bucket: https://github.com/ScoopInstaller/Java
````
❯ scoop search openjdk
Results from local buckets...

Name                Version    Source Binaries
----                -------    ------ --------
openjdk-ea          25.0.2-10  java
openjdk             25.0.2-10  java
openjdk10           10.0.2-13  java
openjdk11           11.0.2-9   java
openjdk12           12.0.2-10  java
openjdk13           13.0.2-8   java
openjdk14           14.0.2-12  java
openjdk15           15.0.2-7   java
openjdk16           16.0.2-7   java
openjdk17           17.0.2-8   java
openjdk18           18.0.2.1-1 java
openjdk19           19.0.2-7   java
openjdk20           20.0.2-9   java
openjdk21           21.0.2-13  java
openjdk22           22.0.2-9   java
openjdk23           23.0.2-7   java
openjdk24           24.0.2-12  java
openjdk25           25.0.2-10  java
openjdk7-unofficial 7u80-b32   java
openjdk8-redhat-jre 8u342-b07  java
openjdk8-redhat     8u342-b07  java
openjdk9            9.0.4-12   java
````

Should i think too many possibilities. We should probably just be sure to have JAVA in anyway you want to run .jar 😅 (and for example openjdk 🤷‍♂️ )

File: src/tools/impl/verapdf.ts (hunk 3 above)
Comment (cderv) [reply src/tools/impl/verapdf.ts:87]: If you manage to run `quarto install verapdf` on windows with java then I think we're good.

**Comment by gordonwoodhull** [comment]
Tests need some help now that I understand standards better.

Typst only supports ua-1 and LaTeX only supports ua-2 so shared tests will have to look like this:

```
format:
  pdf:
    pdf-standard: ua-2
  typst:
    pdf-standard: ua-1
```

Most people won't target both formats, so this is mainly a concern for testing, but we'll have to note the differing standards in the documentation.

File: src/tools/impl/verapdf.ts (hunk 3 above)
Comment (gordonwoodhull) [reply src/tools/impl/verapdf.ts:87]: Yes! I installed a random Java 17 in Windows 11, installed verapdf, and see a nice big validation failure for an example

```
[verapdf]: Validating ua2-missing-title.pdf against PDF/UA2...FAILED

WARN: PDF validation failed for ua-2:
The Metadata stream as specified in ISO 32000-2:2020, 14.3 in the document catalog dictionary shall contain a dc:title entry
Output created: ua2-missing-title.pdf
```

Still want to get verapdf testing in windows ci, but I think it can wait.

Comment (gordonwoodhull) [reply src/tools/impl/verapdf.ts:87]: I've changed this to look in the following order:

1. If `QUARTO_VERAPDF` is set, it's assumed to be a verapdf command prefix, like `docker run` or however people need to run verapdf
2. Run the JAR from the Quarto installation of veraPDF directly using `java` (no batch file). This turns out to be safe because the scripts already look for the veraPDF JAR in a forward-compatible way, so we just do the same thing.
3. Fall back to `verapdf` on the `PATH`, if any

Tested on Mac, Windows - let's see how it does in CI!

File: src/tools/impl/verapdf.ts
Comment (gordonwoodhull) [reply src/tools/impl/verapdf.ts:1]: Thanks @cderv!

File: src/tools/impl/verapdf.ts (hunk 3 above)
Comment (cderv) [reply src/tools/impl/verapdf.ts:87]: > Still want to get verapdf testing in windows ci, but I think it can wait.

Let's open an issue, and we can discuss it. I am happy to take this over when we decide what we want the CI todo (only test verapdf install command, and that our integration is working, or use it also to have test for accessiblity regression)

File: tests/docs/smoke-all/pdf-standard/pdfversion-numeric.qmd [hunk 5]
```diff
@@ -0,0 +1,24 @@
+---
+title: "PDF version (numeric YAML)"
+lang: en
+pdf-standard: 2.0
```
cderv, line 4: Regarding CI failure mentioned in latest  commit on main https://github.com/quarto-dev/quarto-cli/commit/4e7e8c831c183372d4056c7db52b5ffd2bf450be

I think this should be string
```
pdf-standard: "2.0"
```

because YAML will parse this as number otherwise I guess

PR #13872, opened 2026-01-11
feat(equations): add alt-text and div syntax for equations

### LINKED ISSUE #13870: Alt-text for math equations in Typst
PDF UA-1 accessibility standard requires math equations (cross-references) to have alternative text.
Typst (> 0.14) supports this with `alt` of [`math.equation()`](https://typst.app/docs/reference/math/equation/) but currently cannot be set.

- https://github.com/quarto-dev/quarto-cli/blob/c7b25eaa36565336a181cdd824bdd5eb172c3014/src/resources/filters/crossref/equations.lua#L56-L63

Expected:

````qmd
---
format:
  typst:
    keep-typ: true
---

$$
E = mc^2
$$

$$
E = mc^2
$$ {#eq-einstein alt="Einstein's mass-energy equivalence formula"}
````

```typst
...

$ E = m c^2 $

#math.equation(block: true, alt: "Einstein's mass-energy equivalence formula", numbering: "(1)", [ $ E = m c^2 $ ])<eq-einstein>
```

mcanouil:
> [!IMPORTANT]
> For the Typst tests to pass on CI/CD, Typst >= 0.14 is required.

Implement alt-text functionality for math equations to comply with the PDF UA-1 accessibility standard.

This enhancement allows users to specify alternative text for equations using the `alt` attribute in Typst 0.14 and above.

This PR also adds support for the div-syntax.

```md

$$
E = mc^2
$$ {#eq-einstein}

::: {#eq-einstein}

$$
E = mc^2
$$

:::

$$
E = mc^2
$$ {#eq-einstein alt="Einsteins mass-energy equivalence equation"}

::: {#eq-einstein alt="Einsteins mass-energy equivalence equation"}

$$
E = mc^2
$$

:::
```

Additionally, refactor the code to reduce duplication (brought by div processing).  

Fixes #13870


mcanouil:
Because to trigger the use of `math.equation` a cross-reference is required, it means `alt` can be added only when a cross-reference.
It's possible to allow to set alt text independently of the cross-reference, if deemed appropriate.

cderv:
> For the Typst tests to pass on CI/CD, Typst >= 0.14 is required.

Does this mean it requires #13249? 

You could make this PR against that branch, that way we'll benefit the updated version. and we'll be sure to merge after.

mcanouil:
Done

mcanouil:
The PR is a bit noisy because it contains commits from main not in the new base branch.

src/resources/filters/crossref/equations.lua
```diff
@@ -21,67 +22,56 @@ function process_equations(blockEl)
...
-          targetInlines:insert(pandoc.RawInline("typst", " ])<" .. label .. ">"))
-        else
-          local eqNumber = eqQquad
-          local mathMethod = param("html-math-method", nil)
-          if type(mathMethod) == "table" and mathMethod["method"] then
-            mathMethod = mathMethod["method"]
-          end
-          if _quarto.format.isHtmlOutput() and (mathMethod == "mathjax" or mathMethod == "katex") then
-            eqNumber = eqTag
+      elseif el.t == "Str" and startsWithEqLabel(el.text) then
```
cderv: Why does this require a new function?  We already have `refLabel()`
https://github.com/quarto-dev/quarto-cli/blob/c7b25eaa36565336a181cdd824bdd5eb172c3014/src/resources/filters/crossref/refs.lua#L178-L184

so `refLabel("eq", el)` works right ? 

just curious to avoid adding duplicates, and also because it seems that with your change `refLabel()` is now no more used at all.

src/project/types/website/website-meta.ts
cderv, line 1: This comes from a commit we already merged on main. 
We'll sort out while merging to keep PR scoped and clean.

src/resources/filters/crossref/refs.lua
```diff
@@ -188,7 +188,28 @@ function extractRefLabel(type, text)
...
+  if not text then return nil, nil end
+
+  -- Wrap in a markdown header since Pandoc requires text before attributes
+  -- to parse them correctly without regular expressions.
+  local parsed = pandoc.read("## x " .. text, "markdown")
+  if parsed and parsed.blocks[1] and parsed.blocks[1].attr then
+    local attr = parsed.blocks[1].attr
+    return attr.identifier, attr.attributes
+  end
+  return nil, nil
```
cderv: This reminds me of the process we do for table caption attributes 
https://github.com/quarto-dev/quarto-cli/blob/f82f02d15ce3c66eeea462f3eb36f8100b1be92a/src/resources/filters/common/tables.lua#L19-L50

Especially regarding your comment
>   -- Wrap in a markdown header since Pandoc requires text before attributes
>   -- to parse them correctly without regular expressions.
 
We do add text and we use `pandoc.read("## {attr}` directly. Is that wrong then ? 

Anyway, I think this might be an opportunity to at least perform the same processing to avoid confusion in the future about why a difference, and indeed create a refactoring function as you did above.

src/resources/filters/crossref/refs.lua
```diff
@@ -188,7 +188,28 @@ function extractRefLabel(type, text)
 end
 
 function refLabelPattern(type)
-  return "{#(" .. type .. "%-[^ }]+)}"
+  -- Captures the identifier (type-name) while allowing optional attributes
+  return "{#(" .. type .. "%-[^ }]+)[^}]*}"
+end
```
cderv: It means that this will impact all usages of `refLabelPattern()` right ? 

So any usage of `extractRefLabel()` and `refLabel()` will "support" attributes. 

I am worried of the side effect... 🤔 

We'll see with tests, but we should probably test also a new case that is using attribute with any case that is using `extractRefLabel()` or `refLabel()` or `refLabelPattern()`. 

Like I think now `\caption{My table {#tbl-foo .class}}` will be parsed. Hopefully it just does not nothing and ignore `.class` at least for now.

src/resources/filters/crossref/equations.lua
```diff
@@ -117,3 +146,93 @@ end
...
+  return text and text:match("^{#eq%-")
+end
+
+-- Collect a complete attribute block from inline elements.
+--
+-- Pandoc tokenises `{#eq-label alt="description"}` into multiple elements:
+--   Str "{#eq-label", Space, Str "alt=", Quoted [...], Str "}"
+--
+-- This function collects and joins these elements into a single string
+-- that can be parsed by parseRefAttr().
```
cderv: Is this different than generic `inlinesToString()` which wraps `pandoc.utils.stringify` ? 
https://github.com/quarto-dev/quarto-cli/blob/f82f02d15ce3c66eeea462f3eb36f8100b1be92a/src/resources/filters/common/pandoc.lua#L76-L85

It appears to be the same idea of obtaining a string output from parsed inlines using Pandoc. 

However, it seems you return `consumed` for some reason... 🤔 
So maybe `pandoc.utils.stringify` is not enough, but if we could avoid maintaining our own, it seems best.

src/resources/filters/crossref/equations.lua
```diff
@@ -117,3 +146,93 @@ end
 function isDisplayMath(el)
   return el.t == "Math" and el.mathtype == "DisplayMath"
 end
+
+-- Check if text starts with an equation label pattern {#eq-
+function startsWithEqLabel(text)
+  return text and text:match("^{#eq%-")
+end
```
cderv: This replace `refLabel()` - if this is needed, we should remove `refLabel`, but the latter seems more generic. 🤔

src/resources/filters/crossref/equations.lua
```diff
@@ -21,67 +22,56 @@ function process_equations(blockEl)
...
-            eqNumber = eqTag
+      elseif el.t == "Str" and startsWithEqLabel(el.text) then
+        -- Collect attribute block: {#eq-label alt="..."} may span multiple elements
+        local attrText, consumed = collectAttrBlock(inlines, i)
+
+        if attrText then
+          -- Parse to extract label and optional attributes (e.g., alt for Typst)
+          local label, attributes = parseRefAttr(attrText)
+          if not label then
+            label = attrText:match("{#(eq%-[^ }]+)")
```
cderv: Is this fallback if pandoc does not parse attributes and extract identifier? 
I would have expect `parseRefAttr` to always succeed, even if only an id 🤔 

Also, it seems the same as `extractRefLabel()` isn't it ?

src/resources/filters/crossref/equations.lua
```diff
@@ -103,7 +93,46 @@ function process_equations(blockEl)
...
+
+    -- Pandoc 3.1.7 started outputting a shadow section with a label as a link target
+    -- which would result in two identical labels being emitted.
+    -- https://github.com/jgm/pandoc/issues/9045
+    -- https://github.com/lierdakil/pandoc-crossref/issues/402
+    result:insert(pandoc.RawInline("latex", "\\end{equation}"))
+
+  elseif _quarto.format.isTypstOutput() then
+    local is_block = eq.mathtype == "DisplayMath" and "true" or "false"
+    local alt_param = alt and (", alt: \"" .. alt .. "\"") or ""
```
cderv: Can there be a problem with quotes here in `alt` ? 

Like `alt="The \"famous\" equation"`? Maybe a bad example,e but I am wondering about quote support . 🤔

tests/docs/smoke-all/2026/01/06/13847/.gitignore
cderv, line 1: Same for this test folder. We'll remove from this PR before merging to get clean PR as it does not belong to this one.

cderv (commented):
Thanks a lot ! 

I reviewed with mostly some questions because this PR touches functions in Lua that are used for all crossref. 

So it is indeed adding support for Typst, but we should verify the side effect for other cases and formats where the function are used - possibly by adding more tests than equations as all crossref will be impacted. 

Also, if we do add support for alt in equation for typst, do we need to try add support for other format ? usually when we add a feature we try to cover all formats, so I am wondering. 

At least we'll need a companion Documentation PR that explains the limitations. 

- Either we document `alt=` in equation syntax only in Typst page: https://quarto.org/docs/output-formats/typst.html
- Either we document new alt support in generic markdown: https://quarto.org/docs/authoring/cross-references.html#equations but in that case it will need a callout for "Typst only" feature.

So maybe former is better. 

Again, thanks for the feature!

src/resources/filters/crossref/equations.lua
```diff
@@ -117,3 +146,93 @@ end
...
+    return collected, consumed
+  end
+
+  return nil, 0
+end
+
+-- Process equation divs with optional alt-text attribute.
+-- Supports syntax: ::: {#eq-label alt="description"} $$ ... $$ :::
+-- The alt attribute is only used for Typst output (accessibility).
+function process_equation_div(divEl)
```
cderv: Regarding Div support, I wonder why it has not been done before... It seems logic, but I wonder if I am missing something.

Anyhow, this callout warning will need to be edited if this is correctly supported now. 
https://quarto.org/docs/authoring/cross-references-divs.html

mcanouil:
> Also, if we do add support for alt in equation for typst, do we need to try add support for other format ? usually when we add a feature we try to cover all formats, so I am wondering.

Briefly, I don't believe LaTeX/OpenXML care about this and if there is even a way. I did not take a deep look at those, so I might have missed something.
For HTML, it's usually the math library that is responsible of accessibility and it's not really doable to do something unless the team officially choose one library to support over the others.
Typst on the other hand has the right tooling and it is actually making us do it. Using `math.equation()` without alternative text is not possible if one wants to be compliant with UA-1.

For the context, I was playing around with my template when I hit those issues, instead of making myself a Lua workaround, I started this PR as it's mandatory for UA-1.

---

Thanks for the review and comments about other functions/tools defined elsewhere in the codebase.
I'll check and revise.

src/resources/filters/crossref/equations.lua
mcanouil: I have the same question.
Note that none of the previous tests failed anywhere in the codebase beside the expected Typst specific ones.

src/resources/filters/crossref/refs.lua
mcanouil: The "x" is actually not needed, it's just to be "extra safe".

> Anyway, I think this might be an opportunity to at least perform the same processing to avoid confusion in the future about why a difference, and indeed create a refactoring function as you did above.

Are you suggesting to refactor both codepath to use the same helper function or similar?

src/resources/filters/crossref/equations.lua
cderv: I am worried this syntax brings confusion if for example you want to do that 

````

::: {#eq-div-alt}

$$
E = mc^2
$$

This is a caption for my equation

:::

See @eq-div-alt.
````

This won't work, and we'll need to explain, than yes you can use Div syntax BUT with no caption. 

So it seems nice to have, but maybe this will be worth adding only if can make equation behave exactly like float ? I wonder. 

Adding this syntax should be discussed I guess.

mcanouil: Two options, revert the processing (original code refactored to share logic) or disable only the div part.
The div syntax was mostly a by-product, so not too much thought into it.

src/resources/filters/crossref/equations.lua
mcanouil: Let's add a test!

cderv:
> > Also, if we do add support for alt in equation for typst, do we need to try add support for other format ? usually when we add a feature we try to cover all formats, so I am wondering.
> 
> Briefly, I don't believe LaTeX/OpenXML care about this and if there is even a way. I did not take a deep look at those, so I might have missed something. For HTML, it's usually the math library that is responsible of accessibility and it's not really doable to do something unless the team officially choose one library to support over the others. Typst on the other hand has the right tooling and it is actually making us do it. Using `math.equation()` without alternative text is not possible if one wants to be compliant with UA-1.


Ok it makes sense, so maybe this something to document only at Typst doc level as an exception, and be sure that we don't impact other format with this. 

For LaTeX, I am aware of the following package for accessibility (https://ctan.org/pkg/axessibility), but I have never tried it. Just overall with the new accessibility requirement, maybe there is something with new standard that we don't know yet about. 

For HTML, yes this is the library, but we do know which one is used `html-math-method`. So it is our job to pass the information the right way for those method if they support alt text. Do you see what I mean? 
That is why I am wondering what we can do (and maybe it is pandoc as we rely on Pandoc writer for the equation rendering I think - which we don't for Typst).

mcanouil:
> For HTML, yes this is the library, but we do know which one is used html-math-method. So it is our job to pass the information the right way for those method if they support alt text. Do you see what I mean?
That is why I am wondering what we can do (and maybe it is pandoc as we rely on Pandoc writer for the equation rendering I think - which we don't for Typst).

I meant, I'm not aware of them using alternative text, for example MathJax https://docs.mathjax.org/en/latest/basic/accessibility.html

mcanouil:
@cderv I'll refocus this PR and open a separate feature request for the div syntax.
This will simplify this PR and "address" your concerns.

src/resources/filters/crossref/equations.lua
mcanouil: `refLabelPattern()` adds `{}`.

https://github.com/quarto-dev/quarto-cli/blob/f82f02d15ce3c66eeea462f3eb36f8100b1be92a/src/resources/filters/crossref/refs.lua#L178-L192

mcanouil: It seems `refLabel()` was used only for equations as they were not expected to get any attributes.

src/resources/filters/crossref/equations.lua
mcanouil: This was a bit tricky and I found a way to ensure the strings are collected and parsed properly for the current cross-reference syntax of equations.

src/resources/filters/crossref/equations.lua
mcanouil: Probably, but I don't see a good reason to bet on that.

I replaced with `extractRefLabel()`

src/resources/filters/crossref/equations.lua
mcanouil: There are similarities, but this function could be eventually be called in the loop, not replace the function (doing so will require quite some refactoring for very little or no gain at all).

There is maybe a better algorithmic approach.

I asked deepseek/Claude (limited power) to look in the codebase for something to be reused here, but no success.

src/resources/filters/crossref/equations.lua
mcanouil: I reverted all changes to refs.lua, thus this `refLabel()` is no longer used anywhere.

src/resources/filters/crossref/refs.lua
mcanouil: Because of the single/double quotes problem, the collect/parse got a bit bigger.
I agree that refactoring to use a shared utility would be preferable in general, but even if they do share a core logic (from what I can see, they only share `pandoc.read("## " .. text)` in term of pure code), both functions were designed/implemented towards two very specific goals (table caption which is a string and "Display math block with a string").
I would open a "maintenance" issue about it.

src/resources/filters/crossref/refs.lua
mcanouil: No longer a problem in this PR.

mcanouil:
- https://github.com/quarto-dev/quarto-cli/pull/13889

PR #13889, opened 2026-01-14
feat(equations.lua): support alt attribute for cross-referenced equations

### LINKED ISSUE #13870: Alt-text for math equations in Typst

mcanouil:
Enhance accessibility by allowing the use of the `alt` attribute for cross-referenced equations for Typst format(s).

Fixes #13870

### DIFF
--- a/src/resources/filters/crossref/equations.lua
+++ b/src/resources/filters/crossref/equations.lua
@@ -1,5 +1,5 @@
 -- equations.lua
--- Copyright (C) 2020-2022 Posit Software, PBC
+-- Copyright (C) 2020-2026 Posit Software, PBC
 
 -- process all equations
 function equations()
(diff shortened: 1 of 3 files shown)


src/resources/filters/crossref/equations.lua
```diff
@@ -21,67 +21,57 @@ function process_equations(blockEl)
...
+      -- Check "starts with" not complete match: Pandoc splits {#eq-label alt="..."} across elements
+      elseif el.t == "Str" and el.text:match("^{#eq%-") then
+        -- Collect attribute block: {#eq-label alt="..."} may span multiple elements
+        local attrText, consumed = collectAttrBlock(inlines, i)
+
+        if attrText then
+          -- Parse to extract label and optional attributes (e.g., alt for Typst)
+          local label, attributes = parseRefAttr(attrText)
+          if not label then
+            label = extractRefLabel("eq", attrText)
```
cderv: `extractRefLabel` returns two values, and label will be second one so it should be 
```suggestion
          if not label then
            local _, extracted = extractRefLabel("eq", attrText)
            label = extracted
```

cderv (approved):
Great thanks ! 

This will require a companion doc PR for 1.9. 

Also I am wondering about 
````
$$
E = mc^2
$$ {#eq-display-alt alt = "Einsteins mass-energy equivalence equation"}
````
See `alt = ` with the spaces. I think we do "support" it for fenced div by trying to fix them. 

I wonder if with your processing that reconstruct the tokens we could handle this nicely ? Or not worth doing anything ? 

I just see some users that could use that and I wonder how this will behave. What do you think ?

mcanouil:
Let me check quickly how spaces are handled right now.

Spaces are not allowed ...

mcanouil:
Adding support for spaces around equal signs in attributes.

mcanouil:
@cderv with/without spaces works now.

cderv:
Great, I'll merge this today. Please don't forget a companion doc PR so we can have the doc for 1.9. Thanks !

mcanouil:
Here it is:

- https://github.com/quarto-dev/quarto-web/pull/1850

PR #13894, opened 2026-01-15
fix: resolve symlinks in safeRemoveDirSync to fix cleanup with symlinked paths

### LINKED ISSUE #13890: Error rendering documents with `embed-resources: true` in sym-linked directories
### I have:

### Bug description

When rendering a qmd document using `embed-resources: true` and a full absolute path that involves a sym-linked directory the render fails when attempting to clean up the `*_files/` folder because of confusion between the given and normalized paths.

After the render completes we see an error like the following:

```
quarto render /home/c/cr173/tmp/quarto-bug/test.qmd

processing file: test.qmd
1/2                  
2/2 [unnamed-chunk-1]
output file: test.knit.md

pandoc 
  to: html
  output-file: test.html
  standalone: true
  embed-resources: true
  section-divs: true
  html-math-method: mathjax
  wrap: none
  default-image-extension: png
  variables: {}
  
metadata
  document-css: false
  link-citations: true
  date-format: long
  lang: en
  
ERROR: Refusing to remove directory /home/vis/cr173/tmp/quarto-bug/test_files that isn't a subdirectory of /home/c/cr173/tmp/quarto-bug

Stack trace:
    at safeRemoveDirSync (file:///opt/quarto-1.8.26/bin/quarto.js:2540:11)
    at file:///opt/quarto-1.8.26/bin/quarto.js:114469:9
    at Array.forEach (<anonymous>)
    at renderCleanup (file:///opt/quarto-1.8.26/bin/quarto.js:114467:23)
    at file:///opt/quarto-1.8.26/bin/quarto.js:115409:44
    at withTiming (file:///opt/quarto-1.8.26/bin/quarto.js:18977:21)
    at Object.complete (file:///opt/quarto-1.8.26/bin/quarto.js:115409:9)
    at eventLoopTick (ext:core/01_core.js:178:7)
    at async Object.onPostProcess (file:///opt/quarto-1.8.26/bin/quarto.js:123992:28)
    at async renderFileInternal (file:///opt/quarto-1.8.26/bin/quarto.js:123976:3)
```

Note the difference between `/home/vis/cr173/tmp/quarto-bug/test_files` and `/home/c/cr173/tmp/quarto-bug`. On our server `/home/c/cr173` is a sym-link to `/home/vis/cr173`.

Myself and @mine-cetinkaya-rundel ran into the issue on our departmental server(s) running Posit Workbench (2026.01.0+392.pro5). It was originally observed when rendering from Positron but we've been able to replicate it with the quarto cli using versions  1.8.25, 1.8.26, and 1.9.16.

### Steps to reproduce

Create `test.qmd` in `/data/quarto-bug/`:

````qmd
---
format: 
  html:
    embed-resources: true
---

```{r}
plot(mtcars)
```
````

Create a symlink from `quarto-bug/` to `quarto-bug-sym/` in `data/`.

Run the following from `/data/quarto-bug/` or `/data/quarto-bug-sym`:
* `quarto render /data/quarto-bug-sym/test.qmd` - fails
* `quarto render /data/quarto-bug/test.qmd` - succeeds
* `quarto render test.qmd` - succeeds

### Actual behavior

Using a path with a sym-link fails with the above error

### Expected behavior

Sym-link path should be normalized to avoid the error

### Your environment

- IDE: Posit Workbench 2026.01.0+392.pro5
- OS: Fedora 43

### Quarto check output

```bash
Quarto 1.8.26
      Quarto cache location: /home/c/cr173/.cache/quarto
      Pandoc version 3.6.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.3.1: OK
      Typst version 0.13.0: OK
      Version: 1.8.26
      Path: /opt/quarto-1.8.26/bin

      TinyTeX: (not installed)
      Chromium: (not installed)

      Using: Installation From Path
      Path: /usr/bin
      Version: 2023

      Chrome:  (not detected)

      Version: 3.14.2
      Path: /usr/bin/python3
      Jupyter: 5.8.1
      Kernels: python3

(/) Checking Jupyter engine render....[IPKernelApp] WARNING | debugpy_stream undefined, debugging will not be enabled

      Version: 4.5.2
      Path: /usr/lib64/R
      LibPaths:
        - /home/vis/cr173/R/x86_64-redhat-linux-gnu-library/4.5
        - /usr/local/lib/R/library
        - /usr/lib64/R/library
        - /usr/share/R/library
      knitr: 1.51
      rmarkdown: 2.30

```

cderv:
When rendering a `.qmd` document with `embed-resources: true` using an absolute path through a symlinked directory, the render fails during cleanup with:

```
ERROR: Refusing to remove directory /home/vis/cr173/tmp/quarto-bug/test_files
that isn't a subdirectory of /home/c/cr173/tmp/quarto-bug
```

## Root Cause

The path mismatch occurs in this sequence:

1. User provides symlinked path: `/home/c/cr173/tmp/quarto-bug/test.qmd`
2. `project.dir` keeps the symlink path (Quarto's `normalizePath()` does NOT resolve symlinks)
3. knitr resolves symlinks: R's `rmarkdown:::abs_path(intermediates_dir)` calls R's `normalizePath()` which DOES resolve symlinks
4. Supporting path becomes real: `/home/vis/cr173/tmp/quarto-bug/test_files`
5. `safeRemoveDirSync()` compares these paths as strings, which fails

The knitr symlink resolution happens in `src/resources/rmd/execute.R`:

```r
supporting <- if (!is.null(intermediates_dir) && file_test("-d", intermediates_dir)) {
  rmarkdown:::abs_path(intermediates_dir)  # THIS RESOLVES SYMLINKS
} else {
  character()
}
```

The deeper issue is that `src/core/deno/monkey-patch.ts` replaces `Deno.realPathSync` with `normalizePath`, which intentionally does NOT resolve symlinks. This was done in commit `3bb0b45dc` (Aug 2022) to work around a Windows UNC path bug where `Deno.realPathSync` returned malformed paths like `UNC\x.org\...` instead of `\\x.org\...`.

That UNC bug was fixed in Deno v1.16 ([denoland/deno#12243](https://github.com/denoland/deno/issues/12243)). Quarto now uses Deno 2.3.1+, so the original workaround is no longer needed.

## Fix

Rather than removing the monkey-patch globally (which would require auditing all code paths), this fix:

1. Creates `src/deno_ral/original-real-path.ts` that saves the original `Deno.realPathSync` before monkey-patching
2. Imports it in `src/quarto.ts` BEFORE the monkey-patch import
3. Modifies `safeRemoveDirSync()` to resolve both paths using the original `realPathSync` before comparison

The separate module with no imports avoids circular dependencies:
```
monkey-patch.ts → core/path.ts → deno_ral/fs.ts → monkey-patch.ts (CIRCULAR!)
```

This is a targeted fix that only affects the safety boundary check, with fallback to original paths if resolution fails.

## Testing

- Unit test added to `tests/unit/ral/safe-remove-dir.test.ts` for symlink handling
- Smoke test added at `tests/smoke/render/render-symlink-embed-resources.test.ts`
- All 206 existing unit tests pass

## Follow-up Consideration

Since the UNC bug was fixed in Deno v1.16 (2021), and Quarto now uses Deno 2.3.1+, the monkey-patch may no longer be necessary at all. A follow-up issue could review whether the monkey-patch can be removed entirely, which would restore proper symlink resolution globally.

References:
- Original monkey-patch: commit 3bb0b45dc8fb61a18c772517e78c43bf5db84c5c, [quarto-vscode#67](https://github.com/quarto-dev/quarto-vscode/issues/67)
- Deno issue: [denoland/deno#12243](https://github.com/denoland/deno/issues/12243)

Fixes #13890


cderv:
I think I got the understanding right, but the resolution wrong looking at commit history...

I see f9cdaa9231863b2410258d7655129a636c10e2b9 with message 
````
remove Deno.realPathSync from our codebase (keep monkey-patched version because deno's stdlib calls into it)
````

So it seems we don't want to use `Deno.realPathSync` at all ! However, we still have the problem of computation engine that could resolve symlinks in the resources and pass information to quarto 🤔 

Like this happens with knitr. So I assumed we needed that. 

@cscheid Do you remember the history behind f9cdaa9231863b2410258d7655129a636c10e2b9 ?  I assumed maybe comestic to avoid having some `Deno.fun` eveywhere while working on `deno_ral/` layer ? 

Initially the monkey was done because of https://github.com/denoland/deno/issues/12243 and I confirmed locally that the deno version we used solves this. 

This PR change is scoped in the sense we do use `Deno.realPathSync` only for remove function. 

but maybe we don't need to monkey-path after all the deno std itself. 

Anyhow, I know you're on other focus, but it feels having another dev look would help me confirm my initial fix.

PR #13901, opened 2026-01-16
Fix remote fonts not recognised when using a brand extension

gordonwoodhull:
## Summary

Fixes #13685

- Fix issue where remote font URLs in `brand.yml` are incorrectly treated as file paths when the brand file is in a subfolder (like an extension)
- Add `isExternalPath` helper to detect URLs (paths starting with a protocol like `https:`) and skip path joining for external font paths
- Add `ensureCssRegexMatches` test assertion for verifying CSS content in supporting files

## Test plan

- [x] Added smoke test that verifies:
  - CSS contains correct font URL (`src:url("https://notofonts.github.io/...`)
  - CSS does NOT contain broken path (`_extensions/my-brand/https:`)
- [x] Test passes locally

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/core/sass/brand.ts
+++ b/src/core/sass/brand.ts
@@ -150,6 +150,8 @@ const googleFontImportString = (description: BrandFontGoogle) => {
   }:${styleString}wght@${weights}&display=${display}');`;
 };
 
+const isExternalPath = (path: string) => /^\w+:/.test(path);
+
 const fileFontImportString = (brand: Brand, description: BrandFontFile) => {
   const pathPrefix = relative(brand.projectDir, brand.brandDir);
   const parts = [];
@@ -162,9 +164,12 @@ const fileFontImportString = (brand: Brand, description: BrandFontFile) => {
       weight = file.weight;
       style = file.style;
     }
+    const fontUrl = isExternalPath(path)
+      ? path
+      : join(pathPrefix, path).replace(/\\/g, "/");
     parts.push(`@font-face {
     font-family: '${description.family}';
-    src: url('${join(pathPrefix, path).replace(/\\/g, "/")}');
+    src: url('${fontUrl}');
     font-weight: ${weight || "normal"};
     font-style: ${style || "normal"};
 }\n`);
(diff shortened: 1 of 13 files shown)


gordonwoodhull:
It hallucinated `ensureCssRegexMatches`, which I've often wanted, so I told it to ask deepwiki how to add the assertion. 

Implementation is messy and IIUC just looks for CSS files in output directory - probably should be refined before merging.

cderv:
> It hallucinated ensureCssRegexMatches, which I've often wanted, so I told it to ask deepwiki how to add the assertion.

This is interesting! I think we did use playwright for this kind of CSS verification until now. Mainly to be sure that the CSS applied in browser is the one expected. Or at least this is how I tried to use some CSS value check using playwright.

src/core/sass/brand.ts
```diff
@@ -150,6 +150,8 @@ const googleFontImportString = (description: BrandFontGoogle) => {
   }:${styleString}wght@${weights}&display=${display}');`;
 };
 
+const isExternalPath = (path: string) => /^\w+:/.test(path);
```
cderv: FWIW we have this 
https://github.com/quarto-dev/quarto-cli/blob/ba576a817ef97e2f4b197c774acc35061712f377/src/core/url.ts#L10-L12

but yours is more generic. Also, if you need this one, you should probably add it in one of the core files as we already have several times

https://github.com/quarto-dev/quarto-cli/blob/ba576a817ef97e2f4b197c774acc35061712f377/src/core/brand/brand.ts#L275-L278

https://github.com/quarto-dev/quarto-cli/blob/ba576a817ef97e2f4b197c774acc35061712f377/src/project/types/website/website-navigation.ts#L1543-L1545

https://github.com/quarto-dev/quarto-cli/blob/ba576a817ef97e2f4b197c774acc35061712f377/src/project/types/website/website-utils.ts#L123-L125

Maybe it is time to have this helper in one place.

gordonwoodhull:
Improved implementation of `ensureCssRegexMatches` looks only at CSS referenced by the html. It's a little hacky and I think I really want `ensureScssRegexMatches` but I don't think it does any harm to have another assertion.

PR #13902, opened 2026-01-16
fix: resolve relative font-paths in typst extensions (#13745)

gordonwoodhull:
## Summary

- Resolves relative font paths to absolute paths before passing them to `typst compile`
- Fixes issue where `font-paths` specified in extension configurations weren't being found when the input document was in a subdirectory
- Adds test case verifying extension font-paths work correctly from subdirectories

## Problem

When using `font-paths` in a custom Typst format extension, relative paths weren't working because:
1. The extension mechanism resolves font-paths relative to the input document
2. But `typst compile` runs from a different working directory (project root or temp dir)
3. The relative paths therefore don't point to the correct location

## Solution

In `src/command/render/output-typst.ts`, convert relative font paths to absolute paths by joining them with `inputDir` (the directory containing the input document) before passing to `typstCompile`.

Fixes #13745

## Test plan

- [ ] New smoke test in `tests/docs/smoke-all/typst/extension-font-paths/` verifies no "unknown font family" warning when using extension with relative font-paths from a subdirectory
- [ ] CI tests pass


src/command/render/output-typst.ts
```diff
@@ -63,9 +69,26 @@ export function typstPdfOutputRecipe(
...
+    // (containing _extensions or starting with ..) need resolution relative to
+    // inputDir, while other paths (like .quarto/font-cache) should remain
+    // relative to the working directory.
+    const fontPaths = asArray(format.metadata?.[kFontPaths]).map((p) => {
+      const fontPath = p as string;
+      if (isAbsolute(fontPath)) {
+        return fontPath;
+      }
+      // Extension-resolved paths need to be resolved relative to input file
+      if (fontPath.includes("_extensions") || fontPath.startsWith("..")) {
```
gordonwoodhull, line 83: Ugh, this is a messy fix - might be better to understand why we are using file-relative paths here and use project-relative instead.

gordonwoodhull:
closing in favor of #14068

PR #13904, opened 2026-01-16
Warn when cell-renderings has duplicate names

### LINKED ISSUE #13900: cell-renderings.lua silently drops duplicated render names
> As for why it validates with four plots - there are other uses for this such as showing different versions of a plot on mobile vs desktop, so it's more general than just light and dark. It should probably warn if you reuse light and dark - the extras are discarded.

_Originally posted by @gordonwoodhull in https://github.com/quarto-dev/quarto-cli/discussions/13893#discussioncomment-15518484_

gordonwoodhull:
## Summary
- Adds a warning when `renderings` attribute contains duplicate names (e.g., `[dark, light, dark, light]`)
- Previously, duplicate names were silently dropped with only the last cell output for each name being used
- Users now get a clear warning explaining which outputs will be discarded

Fixes #13900

## Test plan
- [ ] Verify warning is produced when using duplicate rendering names
- [ ] Run smoke-all test: `./run-fast-tests.sh docs/smoke-all/2026/01/16/13900.qmd`


gordonwoodhull:
manually confirmed

```
(W) duplicate rendering name 'dark' in renderings; only the last cell output with each name will be used
(W) duplicate rendering name 'light' in renderings; only the last cell output with each name will be used
```
Sometimes I get Claude to write a haiku about the session, this one's a keeper

>   dark, light, dark, light plots
>   last one wins, the rest are lost
>   now a warning speaks

PR #13915, opened 2026-01-19
fix(shortcodes): use semicolon delimiter to avoid smart typography issue

### LINKED ISSUE #13913: Strange number in documentation

Report an issue on quarto.org

### Description

In the [documentation about using Julia and VSCode](https://quarto.org/docs/computations/julia.html#vs-code), it says:

> The extension integrates directly with the [Julia Extension](https://www.julia-vscode.org/docs) to provide the following b58fc729-690b-4000-b19f-365a4093b2ff-7B7B3C206D657461206A7570797465722D6C616E6775616765203E7D7D–specific capabilities:

This doesn't make sense.

cderv:
Shortcodes followed by a trailing dash (e.g., `{{< version >}}-`) were being rendered as hex-encoded strings:

```
b58fc729-690b-4000-b19f-365a4093b2ff-7B7B3C2076657273696F6E203E7D7D–
```

## Root Cause

PR #13792 introduced a two-stage transformation that encodes shortcodes as `UUID-HEX-` to safely pass through Pandoc's markdown parser. When a user writes `{{< version >}}-`, the encoded form becomes `UUID-HEX--`. Pandoc's `+smart` extension converts `--` to an en-dash `–`, breaking the decode pattern match.

## Fix

Changed the delimiter from `-` to `;` in the UUID-HEX encoding format. Semicolons have no markdown meaning and are not affected by smart typography.

Fixes #13913

### DIFF
--- a/src/resources/pandoc/datadir/lpegshortcode.lua
+++ b/src/resources/pandoc/datadir/lpegshortcode.lua
@@ -321,13 +321,13 @@ local function string_to_hex(str)
 end
 
 local md_shortcode_2_uuid         = "b58fc729-690b-4000-b19f-365a4093b2ff"
-local md_shortcode_2_uuid_pattern = "b58fc729%-690b%-4000%-b19f%-365a4093b2ff%-"
+local md_shortcode_2_uuid_pattern = "b58fc729%-690b%-4000%-b19f%-365a4093b2ff;"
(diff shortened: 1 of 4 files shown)


cderv:
@cscheid Just asking you for review to confirm this small change of delimiters does not change your initial fix in #13792

PR #13932, opened 2026-01-21
feat(website): add llms.txt support for LLM-friendly content

cscheid:
Add support for generating llms.txt and .llms.md files for Quarto websites, providing LLM-friendly markdown versions of HTML pages.

Features:
- New `llms-txt: true` option in website config
- Generates .llms.md companion files alongside HTML output
- Creates llms.txt index file linking to all markdown pages
- Converts HTML to clean markdown using Pandoc with Lua filter
- Handles callouts (blockquotes with bold type markers)
- Converts images to markdown syntax
- Converts internal links from .html to .llms.md
- Respects draft settings (excludes drafts from output)
- Cleans listing pages (removes empty links, category badges)
- Matches sitemap behavior for incremental builds

New files:
- src/project/types/website/website-llms.ts
- src/resources/filters/llms/llms.lua

Test coverage:
- Basic file generation
- Content conversion (callouts, code, tables, links)
- Draft handling
- Listing page cleanup


cderv:
Overall, it looks good! I think this is a great way to do it.  

So I had a look at **pkgdown** implementation to compare:

- They use `llm-docs: false` as fields to opt-out. 

- It also generates a `llms.txt` by mixing the content of `README.md` for the description part, and then an index of all the pages. For pkgdown, usually `README.md` is the `index.html` content. So it is included as such. Not sure it makes sense to add index.html content. 🤔 

- All pages '<name>.html` are just `<name>.md` - not `llm.md`. I don't know if there is a convention. Website like Claude doc all use `<name>.md` (example with https://code.claude.com/docs/en/settings.md)

- Links are absolute paths in **pkgdown** - they contain the site url. Same for claude docs website: https://code.claude.com/docs/llms.txt

- After links there is a description based on source file header, the `: <description>` optional part. It would be good. I wonder if we should add a `title` and `description` field under `llm-docs` to customize that ? Or if we can find a way to use a default if provided (like `open-graph.description` ? ). Otherwise there is a `description-meta` available in HTML if set already 
https://github.com/quarto-dev/quarto-cli/blob/6c8a9b110fcde62f166b86077195bb43ce5d3034/src/resources/formats/html/pandoc/html.template#L16-L18
 
- They do some special processing of some content, probably based on what a pkgdown can have. We will probably add more as users use this feature if we missed some. Just one example below. 

<details><summary>For example Definition list are converted to bullet list. Probably because GFM does not support them</summary>
<p>

```
❯ pandoc -f html -t gfm
<dl>
<dt>Term 1</dt>
<dd>
<p>Definition 1</p>
</dd>
</dl>
<p>Term 2 with <em>inline markup</em></p>
^Z
Term 1
Definition 1

Term 2 with *inline markup*
```

unless activated but is this really GFM syntax ? 
```
❯ pandoc -f html -t gfm+definition_lists
<dl>
<dt>Term 1</dt>
<dd>
<p>Definition 1</p>
</dd>
</dl>
<p>Term 2 with <em>inline markup</em></p>
^Z
Term 1

:   Definition 1

Term 2 with *inline markup*
```
    
</p>
</details> 

An example of their output: https://pkgdown.r-lib.org/llms.txt


I am thinking among good ideas: 
- Using absolute url for links if `site-url` is set? A LLM would not have to reconstruct the url from base + relative link that way. 🤔 
- using same argument `llm-docs` ? but `llm-txt` seems more clear regarding the specification website and created file. 
- allow to customize `[title](url): description` or just use `<meta name="description" content="$description-meta$" />` content if available ?

Just some ideas - I am sure we'll have more feedback when this will be tested.

tests/verify.ts
```diff
@@ -584,6 +584,78 @@ export const ensureFileRegexMatches = (
...
+
+// Verify the .llms.md companion file exists for an HTML output.
+export const ensureLlmsMdExists = (htmlFile: string): Verify => {
+  const llmsFile = htmlFile.replace(/\.html$/, ".llms.md");
+  return {
+    name: `File ${llmsFile} exists`,
+    verify: (_output: ExecuteOutput[]) => {
+      verifyPath(llmsFile);
+      return Promise.resolve();
+    },
```
cderv, line 665: We have `fileExists()` if we want to refactor and avoid duplication
https://github.com/quarto-dev/quarto-cli/blob/6c8a9b110fcde62f166b86077195bb43ce5d3034/tests/verify.ts#L228-L236

tests/verify.ts
```diff
@@ -584,6 +584,78 @@ export const ensureFileRegexMatches = (
...
+
+// Verify the .llms.md companion file does NOT exist for an HTML output.
+export const ensureLlmsMdDoesNotExist = (htmlFile: string): Verify => {
+  const llmsFile = htmlFile.replace(/\.html$/, ".llms.md");
+  return {
+    name: `File ${llmsFile} does not exist`,
+    verify: (_output: ExecuteOutput[]) => {
+      verifyNoPath(llmsFile);
+      return Promise.resolve();
+    },
```
cderv, line 677: We have `pathDoNotExists()` if we want to reuse and avoid code duplication
https://github.com/quarto-dev/quarto-cli/blob/6c8a9b110fcde62f166b86077195bb43ce5d3034/tests/verify.ts#L238-L246

tests/verify.ts
```diff
@@ -584,6 +584,78 @@ export const ensureFileRegexMatches = (
...
+// Verify the llms.txt index file in a website output directory.
+// Takes the HTML file path and looks for llms.txt in the same directory.
+export const ensureLlmsTxtRegexMatches = (
+  htmlFile: string,
+  matchesUntyped: (string | RegExp)[],
+  noMatchesUntyped?: (string | RegExp)[],
+): Verify => {
+  const llmsTxtPath = join(dirname(htmlFile), "llms.txt");
+  return verifyFileRegexMatches(regexChecker, `Inspecting ${llmsTxtPath} for Regex matches`)(llmsTxtPath, matchesUntyped, noMatchesUntyped);
+};
```
cderv, line 690: This verify helper is to be used only for `index.qmd` or another .qmd test that will be at the root of the output dir right ? 

If we need to have verify function that works on `output-dir` as input, it is just a matter of adding the function as special handling in `smoke-all.test.ts` and you could have

````ts
export const ensureLlmsTxtRegexMatches = (
  outputDir: string,
  matchesUntyped: (string | RegExp)[],
  noMatchesUntyped?: (string | RegExp)[],
): Verify => {
  const llmsTxtPath = join(outputDir, "llms.txt");
  return verifyFileRegexMatches(regexChecker, `Inspecting ${llmsTxtPath} for Regex matches`)(llmsTxtPath, matchesUntyped, noMatchesUntyped);
};
````

But I guess this is just a matter of being sure to use `ensureLlmsTxtRegexMatches()` only in compatible source document. 

Just a thought while reviewing the new functions in `verify.ts`

File: tests/verify.ts [hunk 4]
```diff
@@ -584,6 +584,78 @@ export const ensureFileRegexMatches = (
...
+// Takes the HTML file path and looks for llms.txt in the same directory.
+export const ensureLlmsTxtExists = (htmlFile: string): Verify => {
+  const llmsTxtPath = join(dirname(htmlFile), "llms.txt");
+  return {
+    name: `File ${llmsTxtPath} exists`,
+    verify: (_output: ExecuteOutput[]) => {
+      verifyPath(llmsTxtPath);
+      return Promise.resolve();
+    },
+  };
```
cderv, line 702: Same - could be `fileExists()`

tests/verify.ts
```diff
@@ -584,6 +584,78 @@ export const ensureFileRegexMatches = (
...
+// Takes the HTML file path and looks for llms.txt in the same directory.
+export const ensureLlmsTxtDoesNotExist = (htmlFile: string): Verify => {
+  const llmsTxtPath = join(dirname(htmlFile), "llms.txt");
+  return {
+    name: `File ${llmsTxtPath} does not exist`,
+    verify: (_output: ExecuteOutput[]) => {
+      verifyNoPath(llmsTxtPath);
+      return Promise.resolve();
+    },
+  };
```
cderv, line 715: And same could be `pathDoNotExists`

cscheid:
I'll fix the refactorings but I don't want to keep iterating on features: I want to just ship this.

tests/verify.ts
cscheid: Sorry, but I think it's `pathDoNotExists` that needs to go; verifyNoPath has been used for a long time, see eg `noSupportingFiles`.

```
611098eb4f (Charles Teague      2021-06-04 14:44:01 -0400 1155)       verifyNoPath(outputFile.supportPath);
```

cderv:
> I'll fix the refactorings but I don't want to keep iterating on features: I want to just ship this.

Yes agreed. Sorry it wasn't clear. 

It was just improvement ideas we can do later. 

Except maybe the absolute url. All 'llms.txt' I have worked with recently are using absolute path, and even with that sometimes agent are guessing some url. 

So with only relative path, it needs to be clear to the agent what is the base url to use. I think using absolute path directly would help.

cderv:
@cscheid Do you want me to handle the conflict and merge ? I think this is ready and @ashleyhenry15 (👋 ) is waiting patiently 😄

PR #13935, opened 2026-01-22
Fix interactive tool selection in quarto install/update/uninstall

cderv:
When running `quarto install`, `quarto update`, or `quarto uninstall` without arguments, the interactive tool selector shows the list but the selection always returns `undefined`.

## Root Cause

Same issue as #9765 - Cliffy's `Select.prompt()` returns the selected value directly, but the code incorrectly accesses `.value` on the result.

## Fix

Use the `prompt()` array syntax with a named field, same pattern as #9765.

### DIFF
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -127,6 +127,7 @@ All changes included in 1.9:
 
 ## Other fixes and improvements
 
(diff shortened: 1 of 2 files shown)


cderv:
Building 'quarto.exe' for this branch: https://github.com/quarto-dev/quarto-cli/actions/runs/21256479552

Because it seems that cliffy in dev version does not work well - I can't select choices in the list with the arrow 🤷

cderv:
Tested locally that this works as expected now. Glad we found it at last !!

PR #13957, opened 2026-01-26
Fix stack trace shown for Typst compilation errors

gordonwoodhull:
## Summary

- When Typst compilation fails (e.g., PDF/UA compliance errors), show only the Typst error message without a confusing internal stack trace
- Uses `ErrorEx` with `printStack=false` for clean user-facing errors
- Also adds support for array of `printsMessage` checks in smoke tests (enabling tests for both presence AND absence of messages)

Fixes #13942

## Test plan

- [x] Tested in production build - error output now shows:
  ```
  ERROR: error: PDF/UA-1 error: missing alt text
      ...
  ERROR: Typst compilation failed
  ```
  Without the stack trace that was previously shown.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/tests/smoke/smoke-all.test.ts
+++ b/tests/smoke/smoke-all.test.ts
@@ -277,7 +277,11 @@ function resolveTestSpecs(
               verifyFns.push(verifyMap[key](outputFile.outputPath, ...value));
             }
           } else if (key === "printsMessage") {
-            verifyFns.push(verifyMap[key](value));
+            // Support both single object and array of printsMessage checks
+            const messages = Array.isArray(value) ? value : [value];
+            for (const msg of messages) {
+              verifyFns.push(verifyMap[key](msg));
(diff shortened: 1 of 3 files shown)


tests/smoke/smoke-all.test.ts
cderv, line 1: Just curious, as there is no test associated with this change. What is the syntax that is enabled by this change? 

Thanks

gordonwoodhull: This suppresses the stack trace when showing a Typst failure in a production (packaged) build. 

It can’t be tested because we don’t currently do any significant testing of production builds, and the technique only works in production.

(As developers we’re used to stack traces, but at least in theory we shouldn’t be showing them for expected/known failures in prod. I don’t think we’re very good at this.)

PR #13967, opened 2026-01-28
Fix preview crash on re-render (BadResource error)

cderv:
When re-rendering a file during `quarto preview` in a project, the preview crashes with `BadResource: Bad resource ID` from the Sass cache Deno.Kv operations.

## Root Cause

This is a regression from #13804 which changed how `ProjectContext` is managed during preview. That PR made the project context persist across re-renders (instead of being recreated each time), but `renderForPreview()` still called `renderResult.context.cleanup()` after each render.

The cleanup closes shared resources like the SassCache Deno.Kv handle. On subsequent re-renders, the code returns the cached `SassCache` instance whose KV handle was already closed, causing the `BadResource` error when calling `kv.get()`.

## Fix

1. Remove the `cleanup()` call from `renderForPreview()` - the project context is owned by `preview()` and should only be cleaned up when preview exits via `onCleanup()`

2. Invalidate the `fileInformationCache` entry for the rendered file before each render - since `cleanup()` previously cleared all caches (including `fileInformationCache`), removing the call means stale file content would be reused. The cache entry for the specific file being rendered is now deleted so changes are picked up.

Fixes #13955

### DIFF
--- a/src/project/project-shared.ts
+++ b/src/project/project-shared.ts
@@ -505,18 +505,17 @@ export const ensureFileInformationCache = (
   project: ProjectContext,
   file: string,
 ) => {
-  const normalizedFile = normalizePath(file);
   if (!project.fileInformationCache) {
-    project.fileInformationCache = new Map();
+    project.fileInformationCache = new FileInformationCacheMap();
   }
   assert(
     project.fileInformationCache instanceof Map,
     JSON.stringify(project.fileInformationCache),
   );
-  if (!project.fileInformationCache.has(normalizedFile)) {
-    project.fileInformationCache.set(normalizedFile, {} as FileInformation);
+  if (!project.fileInformationCache.has(file)) {
+    project.fileInformationCache.set(file, {} as FileInformation);
   }
-  return project.fileInformationCache.get(normalizedFile)!;
+  return project.fileInformationCache.get(file)!;
 };
 
 export async function projectResolveBrand(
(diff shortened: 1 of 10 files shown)


src/project/project-shared.ts
cderv, line 1: This normalization is related to 
- https://github.com/quarto-dev/quarto-cli/issues/12401

Windows / Linux and Project vs Single file leads to different paths. I tried to make that works because we need to delete by key, so it is important to get the right path, but obviously this modify current behavior, and leads to failed test.

cderv:
😞 I tried to preview quarto-web with this PR, and I still encounter some issue with sass Kv, we may have an other issue.

cderv:
Ok so I don't manage to reproduce, maybe it was a bad state. `quarto preview` on quarto-web works ok to navigate; 

However, as expected, change .scss file for example does trigger re-render, but the change are not shown. Though I am really not even sure this working with current 1.8 as discussed. 

However, testing `quarto render` on a single doc, no project fails to udpate. 

And I am quite convinced this is because of this normalize path key. So not good to merge yet

- Quarto Project => quarto preview works
- Single file => quarto preview does not show change in output file.

cderv:
🤦 It is late for me and I spent much time of this, now this should really be non-normalized path for lookup. 

I'll test tomorrow manually the behavior

cderv:
So the normalizing is done correctly this time. All tests passes. I am currently checking other preview features and we're good!

cderv:
## Manual Testing Summary

Tested using Claude Code with Chrome DevTools MCP to verify live browser updates.

### Test Matrix

| Scenario | Code Path | Result |
|----------|-----------|--------|
| **Project context** (with `_quarto.yml`) | `renderProject` | ✅ |
| **Single file** (no `_quarto.yml`) | `singleFileProjectContext` | ✅ |
| **RevealJS format** | Different output handling | ✅ |

### Detailed Results

**Input file types tested:**
- `.qmd` files ✅
- `.md` files ✅  
- `.ipynb` files ✅

**Resource files tested:**
- `.scss` custom styles ✅
- `.lua` filters ✅
- Image files ⚠️ (reload only, no re-render - expected)

**Stress tests:**
- Rapid edits (5+ quick saves) ✅
- Stop and restart preview ✅

### Key Finding

**No BadResource errors** across 15+ render cycles in all scenarios.

### Pre-existing Issues Found (NOT regressions)

These were verified against baseline 1.9.17 - same behavior:
- Partial files (`{{< include _partial.qmd >}}`) not watched
- `_quarto.yml` changes not detected

cderv:
Yeah all tests passes with latest change too. If this is good TS change, then we can merge and build the next prerelease !

cscheid (approved):
This looks great 🚀

cderv:
Thank you. I'll clean the branch and merge

PR #13978, opened 2026-01-30
fix(typst): prevent breaking inside definition items

mcanouil:
Ensure terms and their descriptions remain together by preventing breaks within definition items.

The rule needs to be separate from the content rule, because here we play on the implicit block generated by Typst.

As this is a visual change, there are no tests, but I believe non breaking term/definition is a sensible default.

<table>
<tr><th>Code</th><th>Before</th><th>After</th></tr>
<tr><td>

````qmd
---
title: "Quarto Playground"
format: typst
keep-typ: true
---

## Introduction

{{< lipsum 4 >}}

Multi-format output
: Generate HTML, PDF, Word, ePub, and more from one source.
````

</td><td>

Two-page document showing Quarto Playground title page with Lorem Ipsum text with at the bottom Multi-format output followed on the facing page with definition in centered box

</td>

</td><td>

Two-page document showing Quarto Playground title page with Lorem Ipsum text and facing page with Multi-format output definition in centered box

</td></tr>
</table>

### DIFF
--- a/tests/docs/smoke-all/typst/definition-item-no-break.qmd
+++ b/tests/docs/smoke-all/typst/definition-item-no-break.qmd
@@ -1,0 +1,24 @@
+---
+title: "Definition Item No Break Test"
+format: typst
+_quarto:
+  tests:
+    typst:
+      ensurePdfTextPositions:
+        - # Term and description must be on the same page (fails if on different pages)
+          - subject: "Generate HTML"
+            relation: below
+            object: "Multi-format"
+        - # Description should be indented (NOT left-aligned with term)
+          - subject: "Generate HTML"
+            relation: leftAligned
+            object: "Multi-format"
+      noErrors: default
+---
+
+## Introduction
+
+{{< lipsum 4 >}}
+
+Multi-format
+: Generate HTML, PDF, Word, ePub, and more from one source.
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -56,6 +56,7 @@ All changes included in 1.9:
 - ([#13870](https://github.com/quarto-dev/quarto-cli/issues/13870)): Add support for `alt` attribute on cross-referenced equations for improved accessibility. (author: @mcanouil)
 - ([#13950](https://github.com/quarto-dev/quarto-cli/pull/13950)): Replace ctheorems with theorion package for theorem environments. Add `theorem-appearance` option to control styling: `simple` (default, classic LaTeX style), `fancy` (colored boxes with brand colors), `clouds` (rounded backgrounds), or `rainbow` (colored start border and colored title).
 - ([#13954](https://github.com/quarto-dev/quarto-cli/issues/13954)): Add support for Typst book projects via format extensions. Quarto now bundles the `orange-book` extension which provides a textbook-style format with chapter numbering, cross-references, and professional styling. Book projects with `format: typst` automatically use this extension.
+- ([#13978])(https://github.com/quarto-dev/quarto-cli/pull/13978)): Keep term and description together in definition lists to avoid breaking across pages. (author: @mcanouil)
 
 ### `pdf`
 
--- a/src/resources/formats/typst/pandoc/quarto/definitions.typ
+++ b/src/resources/formats/typst/pandoc/quarto/definitions.typ
@@ -28,6 +28,9 @@
   it
 }
 
+// Prevent breaking inside definition items, i.e., keep term and description together.
+#show terms.item: set block(breakable: false)
+
 // Some quarto-specific definitions.
 
 #show raw.where(block: true): set block(

### CODE REVIEW

**Comment by gordonwoodhull** [comment]
I added a test using ensurePdfTextPositions. Without the fix, it reports

    Cannot compare positions: "Generate HTML" is on page 2, "Multi-format" is on page 1

File: tests/docs/smoke-all/typst/definition-item-no-break.qmd
Comment (mcanouil) [inline tests/docs/smoke-all/typst/definition-item-no-break.qmd:1]: Wow, is this `ensurePdfTextPositions` new?

Comment (gordonwoodhull) [reply tests/docs/smoke-all/typst/definition-item-no-break.qmd:1]: Yes, I added these to validate article layout in Typst PDF output. 

It's currently just right-of, above, etc, and per-side alignment, so I tested that they are **not** left-aligned, but the predicates only work if the items are on the same page, so it worked out handy for this.

Come to think of it, I think I could have tested that the left edge is to the right of the other left edge as a positive test. 🤔 

I'm still getting used to it and figuring out the complete API, please try it out and see if it's helpful. I'll add `dev-docs` soon.

In theory it should also work with LaTeX tagged PDFs, but I haven't tested that yet.

**Review by gordonwoodhull — APPROVED** [review]

**Review by cderv — APPROVED** [review]
Thank you! 

I see this is added in `definitions.typ` and concerns Definition block which is a Pandoc blocks initially. 

Would that be worth also making a PR in Pandoc for this fix to be upstream too ?

**Comment by mcanouil** [comment]
The terms rule is already different from Pandoc, should it be part of the PR?
- https://github.com/jgm/pandoc/blob/893a00bcc25a3adad5772f1eff1bd5171d241187/data/templates/default.typst#L3-L10

Edit: in fact the rule is incompatible with how Pandoc writes terms definition.
It would be

```typst
#show terms: it => {
  it
    .children
    .map(child => block(breakable: false)[
      #text(weight: "bold")[#child.term]
      #block(inset: (left: 1.5em, top: -0.4em))[#child.description]
    ])
    .join()
}
```

Question is: should we update upstream and use the above rule instead of the two?

cderv:
I'll let @gordonwoodhull reply. I see it was adapted on our side for PDF accessibility so we need to keep ours, but maybe that would be worth reporting upstream this PDF accessibility improvement then 🤔  Commit: 9525c1bedb3cf80c6ebe40049fe48f5cd537a258

Keeping our template improvement with upstream is best when we can. And when they don't accept it, then we'll need to patch at each update (and remember why we are different).

mcanouil:
The rule I'm showing is a mixed between our fix and Pandoc current implementation.
It should (I did not verify though) pass the accessibility check as it's not using `#strong()`.

cderv:
To be clear, 9525c1b was a switch to what you see in Pandoc's template to a new way that pass PDF accessibility. Are you offering a third way? 

I was thinking of sharing this resulting PR code to Pandoc upstream to report 
- PDF accessibility improvement
- non-breaking definition list

mcanouil:
> Are you offering a third way?

Yes, one that fits with the two rules. The initial suggestion was a simpler rule as it relied on implicit block, but that's also why it then needed a second rule to change `breakable`.

cderv:
Ok so doing 
```
#show terms: it => {
  it
    .children
    .map(child => block(breakable: false)[
      #text(weight: "bold")[#child.term]
      #block(inset: (left: 1.5em, top: -0.4em))[#child.description]
    ])
    .join()
}
```

would avoid the `#show terms.item: set block(breakable: false)` ? 

@gordonwoodhull what do you think ?

mcanouil:
Yes, it would

````qmd
---
title: "Quarto Playground"
format:
  typst:
    keep-typ: true
---

```{=typst}
#show terms: it => {
  it
    .children
    .map(child => block(breakable: false)[
      #text(weight: "bold")[#child.term]
      #block(inset: (left: 1.5em, top: -0.4em))[#child.description]
    ])
    .join()
}
```

## Introduction

{{< lipsum 4 >}}

Multi-format output
: Generate HTML, PDF, Word, ePub, and more from one source.
````

**Comment by gordonwoodhull** [comment]
I'm in favor of a unified rule, and contributing upstream. However, you'll want to test with

```
format:
  typst:
    pdf-standard: ua-1
```

as the above proposal fails Typst's validation:

```
[typst]: Compiling definition-item-no-break.typ to definition-item-no-break.pdf...

ERROR: error: PDF/UA-1 error: invalid list (L) structure
    ┌─ tests/docs/smoke-all/typst/definition-item-no-break.typ:382:2
    │
382 │ / Multi-format: #block[
    │   ^^^^^^^^^^^^
    │
    = hint: list (L) may not contain paragraph (P)
    = hint: this is probably caused by a show rule

error: PDF/UA-1 error: invalid list (L) structure
    ┌─ tests/docs/smoke-all/typst/definition-item-no-break.typ:382:0
    │  
382 │ ╭ / Multi-format: #block[
383 │ │ Generate HTML, PDF, Word, ePub, and more from one source.
384 │ │ ]
    │ ╰─^
    │  
    = hint: list (L) may not contain marked content directly
    = hint: this is probably caused by a show rule
```

It does pass the `ensurePdfTextPositions` test, and looks visually OK!

Interestingly, the previous code (from this PR) does pass UA-1.

**Comment by mcanouil** [comment]
New variant that actually does not messed up the list structure as Pandoc code does.
Here we skip directly the parent to directly target the children (items), this way we can change the item content without touching the whole list structure.

```typst
#show terms.item: it => block(breakable: false)[
  #text(weight: "bold")[#it.term]
  #block(inset: (left: 1.5em, top: -0.4em))[#it.description]
]
```

This passed UA-1 Typst validation.

gordonwoodhull:
Much cleaner! We should definitely upstream this.

mcanouil:
Should I make the PR?

mcanouil:
Done

- https://github.com/jgm/pandoc/pull/11436

PR #14003, opened 2026-02-06
Add text fragments to quarto-search.js items

vezwork:
## Description

Adds text fragments to items in search. Example video:

https://github.com/user-attachments/assets/835dec4f-9b58-47c4-aa92-37732d39e195

I have attempted to write this code quite defensively so that it will not break things. Its not quite the ideal place for this code since this piece of code can be called multiple time for a single item, but it gets the job done.

Text fragments are [well supported](https://caniuse.com/?search=text+fragments) and apparently browsers will fall back to id linking if the text fragment is not supported or found (worked in my testing in chrome and safari and firefox on Desktop MacOS Tahoe, e.g. https://en.wikipedia.org/wiki/Dog#Domestication:~:text=blablabla).

### DIFF
--- a/src/resources/projects/website/search/quarto-search.js
+++ b/src/resources/projects/website/search/quarto-search.js
@@ -356,6 +356,22 @@ window.document.addEventListener("DOMContentLoaded", function (_event) {
             },
 
             item({ item, createElement }) {
+              // process items to include text fragments as they are rendered
+              if (item.text && item.href && !item.href.includes(':~:text=')) {
+                // e.g. `item.text` for a search "def fiz": "bla bla bla<mark class='search-match'>def fiz</mark> bla bla"
+                const fullMatches = item.text.matchAll(/<mark class='search-match'>(.*?)<\/mark>/g)
+                // extract capture group with the search match
+                // result e.g. ["def fiz"]
+                const searchMatches = [...fullMatches].map(match => match[1])
+                if (searchMatches[0]) {
+                  if (item.href.includes('#')) {
+                    item.href += ':~:text=' + encodeURIComponent(searchMatches[0])
+                  } else {
+                    item.href += '#:~:text=' + encodeURIComponent(searchMatches[0])
+                  }
+                }
+              }
+              
               return renderItem(
                 item,
                 createElement,


cscheid:
The PR looks good. Just to be clear, this exists in part to improve the situation we have wrt visibility of search results in the page, right?

cderv (approved):

vezwork:
> Just to be clear, this exists in part to improve the situation we have wrt visibility of search results in the page, right?

Yes, this will be helpful for all search results, but will be especially helpful once we have tabsets that have behaviour that they switch tabs to reveal specific content in response to browser ctrl+f or text fragment links. I am looking into using the `<details>` element to create a tabset that supports this. It should work based on my testing across desktop browsers.

vezwork:
Should we merge this? Whats the next release it will be in? Will it be in a "daily" sort of build before the main release so that we can use it in a build of the quarto.org docs?

cderv:
> Will it be in a "daily" sort of build before the main release so that we can use it in a build of the quarto.org docs?

Yes I'll build the new prerelease, and we can see effect in quarto.org

cderv:
1.9.20 has the change

mcanouil:
@vezwork

Probably worth taking a look at this discussion as it's about text fragments and query parameters in URL.
- https://github.com/orgs/quarto-dev/discussions/13123

Especially, the following part which is involving in removing the search text highlighting

- https://github.com/quarto-dev/quarto-cli/blob/1b9b5e43e34203866e3d2c67da8bf4b663796693/src/resources/projects/website/search/quarto-search.js#L56-L64

cderv:
Follow-up note from investigating #14047: text fragments work well on Algolia-powered sites where `?q=` is not in search result URLs. On sites using built-in fuse.js search, the `highlight()` function's DOM mutation (wrapping matches in `<mark>` tags) destroys the browser's text fragment highlight before the user sees it. This is because `?q=` is only added by `fuseSearch()` (~line 1299), triggering the `highlight()`/`replaceState()` code path. As @mcanouil noted in #13123, `replaceState()` also strips `:~:text=`, but testing shows the DOM mutation alone is sufficient to destroy the text fragment even with `replaceState()` blocked.

Both mechanisms serve different purposes (`<mark>` highlights all matches, `:~:text=` highlights one and scrolls to it — see #13123 discussion), so making them coexist on fuse.js sites would be valuable. Related: #14047.

PR #14012, opened 2026-02-09
inclusive parens, not dots

tdhock:
This is a small correction. 
On https://vitrinelinguistique.oqlf.gouv.qc.ca/25343/banque-de-depannage-linguistique/la-redaction-et-la-communication/feminisation-et-redaction-epicene/redaction-epicene/doublets/doublets-abreges/quest-ce-quun-doublet-abrege The Québec Office of the French Language (OQLF) recommends using parentheses or square brackets, not dots, for "doublets abrégés" (here, writing the word "author" in an abbreviated form, which is inclusive).

That page specifically mentions that dots are not recommended because they sometimes cause problems, "Les autres signes parfois employés pour créer des doublets abrégés (trait d’union, majuscule, point et virgule, par exemple) ne sont pas retenus par l’Office en raison des problèmes que leur emploi pose."


cderv:
Hi, thanks for your feedback. 

Do we know what are the problem mentioned? 

I recall this addition for `_language-fr.yml` was done by French users from France. The recommandation we have in France comes from some administration in a report: https://www.egalite-femmes-hommes.gouv.fr/guide-pour-une-communication-publique-sans-stereotype-de-sexe 
In this report, they recommend usage of the dot:
> "le point médian est le signe le plus adapté puisqu’il n’entraine aucune des connotations attachées aux autres signes employés en français. " 

So I believe we are in a situation of variant for a same language. 

The quarto language system supports subtags i.e `lang: fr` and `lang: fr-CA`. 

When the latter is used, `_language-fr-CA.yml` will be look up first, then `_language-fr.yml` and both will be merged. So we could have a `_language-fr-CA.yml` with those specific variations recommended by OQLF while keeping the main fr language file for `fr-FR`. 

Would that help you? 

Or is there technical reason to avoid the dot as delimiter here?

tdhock:
Thanks Christophe.
It does not mention the problem.
that works.

cderv:
> that works.

Do you mean it works locally if you add the file ? 

I was thinking we could add a `_language-fr-CA.yml` in quarto-cli to have this change built in when ou do `lang: fr-CA`

would that help ?

PR #14014, opened 2026-02-10
add lang fr-CA

tdhock:
Hi @cderv 
As you suggested in #14012 
here is a new file for fr-CA.
Does that look ok to you?
Merci !


cderv:
Thanks for the new PR. Yes this is good ! 

Though I was not precise enough: there should be inheritance and merge in the sense that 
- `_language-fr-CA.yml` will be read
- `_language-fr.yml` will be read

Both will be merged, meaning the former can only include what needs to be override in the later. 

That would simplify maintenance. 

To be tested for confirmation.

tdhock:
Thanks for your quick feedback.
I removed everything except the two author lines.
I don’t know how to write a test though.

cderv (approved):
Thanks @tdhock! The implementation looks correct — the minimal-override pattern (only the keys that differ from base fr.yml) is exactly the right approach, consistent with how other regional variants like de-CH.yml work.

I verified that translationsForLang() properly resolves fr-CA by loading _language-fr.yml first, then merging _language-fr-CA.yml on top. Only the overridden keys differ; all other French translations are inherited.

I've added a smoke-all test and one additional override:

- listing-page-field-author (line 114 in _language-fr.yml) also uses the middle-dot form "Auteur·rice". I've added it to the fr-CA file so all three middle-dot author labels are overridden.

The test confirms:
- Auteur(-trice) appears (fr-CA override)
- Date de publication appears (inherited from base _language-fr.yml)
- Auteur·rice does NOT appear (correctly overridden)

PR #14019, opened 2026-02-11
Improve veraPDF Java version error message

cderv:
When installing veraPDF with an unsupported Java version, the error message was generic: "Java is not installed or version is not supported. veraPDF requires Java 8, 11, 17, or 21."

This made it unclear whether Java was missing or just the wrong version.

## Root Cause

The prereq check's error message was a static string, so it couldn't report the detected Java version.

## Fix

Made `InstallPreReq.message` support async functions in addition to static strings. The veraPDF prereq now:
- Detects the installed Java version
- Reports "Java is not installed" if Java is missing
- Reports "Java X is installed but not supported" with the actual version number
- Dynamically builds the supported versions list from `kSupportedJavaVersions`

## Results 

Now I got 

````
❯ quarto install verapdf
Installing verapdf
Java 25 is installed but not supported. veraPDF requires Java 8, 11, 17, 21.
````

However, there is still the question of why 25 is not supported, because manually I can install verapdf

### DIFF
--- a/src/tools/impl/verapdf.ts
+++ b/src/tools/impl/verapdf.ts
@@ -34,14 +34,15 @@ const kVersionFileName = "version";
 export const verapdfInstallable: InstallableTool = {
   name: "VeraPDF",
   prereqs: [{
-    check: async () => {
+    check: async (context) => {
       const javaVersion = await getJavaVersion();
+      context.props.javaVersion = javaVersion;
       return javaVersion !== undefined &&
         kSupportedJavaVersions.includes(javaVersion);
     },
     os: ["darwin", "linux", "windows"],
-    message: async () => {
-      const javaVersion = await getJavaVersion();
+    message: (context) => {
+      const javaVersion = context.props.javaVersion as number | undefined;
       const supportedVersions = kSupportedJavaVersions.join(", ");
       if (javaVersion === undefined) {
         return `Java is not installed. veraPDF requires Java ${supportedVersions}.`;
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
This is good; I’ll test and improve before merging tomorrow.

As discussed, using a non-LTS version should probably be a warning, not an error. Etc.

cderv:
> As discussed, using a non-LTS version should probably be a warning, not an error. Etc.

Yes that would solve the main issue that I had Java but couldn't install. 

And by the way, I managed to make a script wrapper to run vera through docker. So no need to install JAVA, which is helpful ! 

Not sure where I should document that.

gordonwoodhull:
> And by the way, I managed to make a script wrapper to run vera through docker. So no need to install JAVA, which is helpful !
> 
> Not sure where I should document that.

Nice! It's kind of a catch-all but I'd start a page under [Advanced Docs](https://prerelease.quarto.org/docs/advanced/) on veraPDF and put the configuration there (with the `Dockerfile` as a copyable code block, unless it merits its own repo or gist). 

IMO it should be a `verapdf.qmd` page and not a directory... I reckon we won't have too much more to say about veraPDF. If we need more advanced docs in this area it will probably be specific hints and tips around passing PDF accessibility checks, not really about veraPDF itself.

gordonwoodhull:
I've changed non-LTS versions to a warning and also handled the case where Java 25 is LTS but it was released after the current veraPDF, so it needs a different warning.

cderv (commented):
Great thank you !

PR #14033, opened 2026-02-12
Add schema for editor_options in front matter (#14032)

cscheid:
Adds editor_options with chunk_output_type (inline | console) to the YAML schema, enabling autocompletion and validation in Positron and other editors.


cscheid:
Those .mjs regex changes are unexpected, but I think they're due to a change in esbuild behavior. The relevant code is in `src/command/dev-call/build-artifacts/cmd.ts`,

```ts
  const vsCodeSrc = await esbuild(
    resourcePath("../core/lib/yaml-intelligence"),
    "vs-code.ts",
    "esm",
    false,
  );

  const treeSitter = Deno.readTextFileSync(
    resourcePath("editor/tools/yaml/tree-sitter.js"),
  );
  ensureAllowableIDESyntax(treeSitter, "tree-sitter.js");

  Deno.writeTextFileSync(
    resourcePath("editor/tools/vs-code.mjs"),
    [treeSitter, vsCodeSrc].join(""),
  );
```

PR #14049, opened 2026-02-17
Add test coverage for search highlight persistence (#14047, #9802)

### LINKED ISSUE #14047: Search highlights (<mark>) cleared within milliseconds, invisible to users
When navigating to a page via search (with `?q=` parameter), the `highlight()` function adds `<mark>` elements around matching text. These marks are then cleared almost immediately (~24ms on quarto.org) by `quarto-hrChanged` events, making them invisible to users.

## Steps to Reproduce

1. Go to any Quarto website page with `?q=someterm` (e.g. `html-basics.html?q=format`)
2. Observe that no highlighted text is visible

Instrumented testing on quarto.org shows marks appear and disappear within a single frame:

```
404ms: 17 <mark> elements added by highlight()
428ms: 0 <mark> elements — all cleared by resetHighlighting()
```

## Root Cause

The clearing chain:

1. `quarto-search.js` registers a one-shot listener for `quarto-hrChanged` that calls `clearHighlight()`:
   https://github.com/quarto-dev/quarto-cli/blob/aaaba63c2/src/resources/projects/website/search/quarto-search.js#L67-L76

2. `quarto-nav.js` dispatches `quarto-hrChanged` from `updateDocumentOffset()`:
   https://github.com/quarto-dev/quarto-cli/blob/aaaba63c2/src/resources/projects/website/navigation/quarto-nav.js#L196-L197

3. `updateDocumentOffset()` is triggered during initial layout settling by a `ResizeObserver` on the header element and a `setTimeout(250)`:
   https://github.com/quarto-dev/quarto-cli/blob/aaaba63c2/src/resources/projects/website/navigation/quarto-nav.js#L254-L269

The comment says "Clear search highlighting when the user scrolls sufficiently", but the event fires from layout recalculation, not user interaction.

## Context

PR #14003 added text fragments (`:~:text=`) to search result URLs, which provides browser-native scroll-to-text highlighting that persists independently. This effectively works around the broken `<mark>` behavior for visible text, but text fragments cannot highlight content inside hidden elements (e.g. inactive Bootstrap tab panes).

## Possible Fix

Delay the `quarto-hrChanged`/`quarto-sectionChanged` listener registration to skip initial layout settling events, or gate the clearing on actual user interaction (scroll, click).

cderv:
PR #13442 by @jtbayly removed scroll-based search highlight clearing, fixing
both #9802 (highlights cleared on scroll) and #14047 (highlights cleared by
layout events before user sees them).

This PR adds Playwright test coverage and a changelog entry for the expected behavior:

- Search highlights persist through scroll and layout events at any time
- Search highlights clear when the user changes the search query
- No highlights appear without `?q=` parameter

Closes #14047

### DIFF
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -119,7 +119,7 @@ All changes included in 1.9:
 - ([#13932](https://github.com/quarto-dev/quarto-cli/pull/13932)): Add `llms-txt: true` option to generate LLM-friendly content for websites. Creates `.llms.md` markdown files alongside HTML pages and a root `llms.txt` index file following the [llms.txt](https://llmstxt.org/) specification.
 - ([#13951](https://github.com/quarto-dev/quarto-cli/issues/13951)): Fix `image-lazy-loading` not applying `loading="lazy"` attribute to auto-detected listing images.
 - ([#14003](https://github.com/quarto-dev/quarto-cli/pull/14003)): Add text fragments to search result links so browsers scroll to and highlight the matched text on the target page.
-- ([#14047](https://github.com/quarto-dev/quarto-cli/issues/14047)): Fix search highlights cleared before user can see them when landing on a page with `?q=` search parameter.
+- ([#9802](https://github.com/quarto-dev/quarto-cli/issues/9802), [#14047](https://github.com/quarto-dev/quarto-cli/issues/14047)): Fix search term highlighting disappearing on page scroll or layout events when navigating from search results. (author: @jtbayly, [#13442](https://github.com/quarto-dev/quarto-cli/pull/13442))
 
 ### `book`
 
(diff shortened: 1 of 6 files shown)


cderv:
FWIW I have find this problem and this fix by looking at some more ideas to improve search user experience as we discussed. I assumed something was not correct, and found that. This is the required change to follow up on my other ideas. I'll share at next meeting what this enable.

cderv:
Tested the interaction with text fragments from #14003 (context: #13123).

This PR does not worsen the text fragment situation. On fuse.js sites, the `highlight()` DOM mutation already destroys text fragment highlights before our `setTimeout` delay has any effect — the two mechanisms are in conflict regardless of this change.

This fix correctly preserves the JS `<mark>` highlights, which remain the primary highlight mechanism for fuse.js search and the only one that highlights all matches on a page (vs text fragments which highlight a single match). Text fragments continue to work on Algolia-powered sites where `highlight()` doesn't run.

cderv:
This needs to be looked at in addition to other one 
- https://github.com/quarto-dev/quarto-cli/pull/13442

vezwork:
Other PR for this: https://github.com/quarto-dev/quarto-cli/pull/13442

PR #14053, opened 2026-02-18
Activate tabs containing search matches

### LINKED ISSUE #14047: Search highlights (<mark>) cleared within milliseconds, invisible to users

cderv:
> [!IMPORTANT]
> This PR is to be merged AFTER #14049 which as a prerequisite fix 

> [!TIP]
> Demo website: https://cderv.github.io/quarto-search-highlight-demo/ 

When navigating to a page with `?q=term` in the URL, search matches inside inactive Bootstrap tabs are invisible because the matching tab is not activated. This builds on #14049 (mark-clearing fix) to also activate tabs that contain highlighted matches.

## How it works

After `highlight()` wraps matches in `<mark>` elements during DOMContentLoaded, a `pageshow` listener activates any tab panes containing marks. The pageshow timing is key — `tabsets.js` (loaded as an ES module) registers its own pageshow handler during module execution, which runs before DOMContentLoaded. By registering our handler during DOMContentLoaded, listener ordering guarantees we run *after* tabsets.js restores tab state from localStorage.

This means search activation wins over stored tab preferences, which is the intended UX — if you searched for something in the Python tab, you want to see the Python tab regardless of your stored language preference.

## Design decisions

- `bootstrap.Tab.show()` used instead of `.click()` to avoid triggering group sync in tabsets.js. Search activation is intentionally local to the tabset containing the match.
- Nested tabsets supported: walks ancestor `.tab-pane` elements, groups by `.tab-content`, activates outermost first.
- If the active tab already contains a match, no switch occurs (avoids unnecessary tab changes).
- `{once: true}` and `!event.persisted` guard on the pageshow listener for clean teardown.

## Test plan

- [x] Playwright tests (5 scenarios, all pass Chromium/Firefox/WebKit):
  1. Inactive tab activation
  2. Active-tab-has-match stays
  3. Outside-tabs no change
  4. Nested tabsets both activate
  5. Grouped + localStorage override
- [x] TDD verified: 3/5 tests fail against stock v1.9.20, all pass with this code
- [x] Chrome MCP timing experiments confirmed pageshow deferral approach

Fixes quarto-dev/quarto-cli#14047 (partial — tab activation aspect)

### DIFF
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -119,7 +119,7 @@ All changes included in 1.9:
 - ([#13932](https://github.com/quarto-dev/quarto-cli/pull/13932)): Add `llms-txt: true` option to generate LLM-friendly content for websites. Creates `.llms.md` markdown files alongside HTML pages and a root `llms.txt` index file following the [llms.txt](https://llmstxt.org/) specification.
 - ([#13951](https://github.com/quarto-dev/quarto-cli/issues/13951)): Fix `image-lazy-loading` not applying `loading="lazy"` attribute to auto-detected listing images.
 - ([#14003](https://github.com/quarto-dev/quarto-cli/pull/14003)): Add text fragments to search result links so browsers scroll to and highlight the matched text on the target page.
-- ([#14047](https://github.com/quarto-dev/quarto-cli/issues/14047)): Fix search highlights cleared before user can see them when landing on a page with `?q=` search parameter.
+- ([#9802](https://github.com/quarto-dev/quarto-cli/issues/9802), [#14047](https://github.com/quarto-dev/quarto-cli/issues/14047)): Fix search term highlighting disappearing on page scroll or layout events when navigating from search results. (author: @jtbayly, [#13442](https://github.com/quarto-dev/quarto-cli/pull/13442))
 
 ### `book`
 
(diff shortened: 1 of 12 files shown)


cderv:
Rebased on the updated `fix/issue-14047` branch (which is now rebased on main with #13442 merged). The tab activation code is unchanged — no conflicts with the scroll-clearing removal.

All 24 Playwright tests pass (3 highlight + 5 tabset × 3 browsers).

PR #14060, opened 2026-02-18
[sunset] Remove Quarto Pub from 'quarto publish'

cscheid:
Posit Connect Cloud is going to be the sole Posit publishing destination going forward.


cscheid:
We need a corresponding PR on quarto-web too...

src/publish/provider.ts
```diff
@@ -41,5 +39,13 @@ export function findProvider(name?: string) {
         `See https://docs.posit.co/cloud/whats_new/#october-2024 for details.`,
     );
   }
+  if (name === "quarto-pub") {
+    warning(
+      `The Quarto Pub publishing destination is no longer supported. ` +
+        `Consider publishing to Posit Connect Cloud instead ` +
+        `using \`quarto publish posit-connect-cloud\`. ` +
+        `See https://docs.posit.co/cloud/whats_new/#october-2024 for details.`,
```
cwickham: Not this link I think...

cscheid: Thanks!

PR #14068, opened 2026-02-19
fix: resolve relative font-paths for typst compilation

gordonwoodhull:
Fixes #13745
Replaces #13902

## Summary

- Resolves relative `font-paths` to absolute paths against the document directory before passing them to the Typst CLI
- Fixes the bug where extensions providing `font-paths` in their `_extension.yml` fail with "unknown font family" when the document is in a project subdirectory

The root cause: relative font paths (from extensions, document front matter, or project config) were passed unchanged as `--font-path` arguments to the Typst CLI. Typst resolves these relative to its process CWD, which has no defined relationship to the document directory. This caused failures whenever documents weren't at the project root.

The fix is a two-line change in `output-typst.ts` that resolves relative paths against `inputDir` (the document directory) before passing them to `typstCompile`. Absolute paths (from `brand.yaml` font resolution or already-resolved `/`-prefixed project paths) pass through unchanged.

## Test plan

- [x] New smoke test: extension contributing `font-paths` with document in subdirectory (was failing, now passes)
- [x] New smoke test: document `font-paths` in front matter with document in subdirectory (was failing, now passes)
- [x] New smoke test: brand.yaml file fonts with document in subdirectory (was already passing, confirms no regression)
- [x] Existing test: `relative-font-path` project-absolute `/artifacts/fonts` (confirms no regression)
- [x] Typechecks clean


gordonwoodhull:
It's still slightly inelegant that we resolve brand.yaml paths to file system absolute paths instead of resolving them to project absolute paths and then translating them. But Typst itself uses file system absolute paths for `font-paths` and I suppose there are users depending on using them too, so I think this is good enough.

PR #14080, opened 2026-02-20
Add q param to algolia search, fix search term highlighting across html nodes

vezwork:
## Examples
I highlighted and copied some text on various pages that spans decorated and undecorated text, then pasted it into algolia search, and these are the successful matches:

image
---
image
---
image
---

## Description
This PR 
1. replaces text fragments (from PR https://github.com/quarto-dev/quarto-cli/pull/14003) in algolia search urls with a `q={search term}` param, enabling multiple match highlighting
1. replaces the previous method of highlighting search terms in the page which only worked when the search term was fully contained within a single node. The previous method worked great for matches in paragraphs that did not have any decoration, but did not work for matches that contained both decorated and undecorated text, and did not work well for code cells with syntax highlighting.

I referenced @cderv's PR https://github.com/quarto-dev/quarto-cli/pull/14053 while finishing this PR. Building off of that, I found a method for opening containing tabsets that requires significantly less code. See `openAllTabsetsContainingEl`

### DIFF
--- a/src/resources/projects/website/search/quarto-search.js
+++ b/src/resources/projects/website/search/quarto-search.js
@@ -1260,16 +1260,10 @@ function openAllTabsetsContainingEl(el) {
 
 function scrollToFirstVisibleMatch(mainEl) {
   for (const mark of mainEl.querySelectorAll("mark")) {
-    let hidden = false;
-    let el = mark.parentElement;
-    while (el && el !== mainEl) {
-      if (el.classList.contains("tab-pane") && !el.classList.contains("active")) {
-        hidden = true;
-        break;
-      }
-      el = el.parentElement;
-    }
-    if (!hidden) {
+    const isMarkVisible = matchAncestors(mark, '.tab-pane').every(markTabPane =>
+      markTabPane.classList.contains("active")
+    )
+    if (isMarkVisible) {
       mark.scrollIntoView({ behavior: "smooth", block: "center" });
       return;
     }
--- a/src/resources/projects/website/search/quarto-search.js
+++ b/src/resources/projects/website/search/quarto-search.js
@@ -47,6 +47,19 @@ window.document.addEventListener("DOMContentLoaded", function (_event) {
     // perform any highlighting
     highlight(query, mainEl);
 
+    // Activate tabs on pageshow — after tabsets.js restores localStorage state.
+    // tabsets.js registers its pageshow handler during module execution (before
+    // DOMContentLoaded). By registering ours during DOMContentLoaded, listener
+    // ordering guarantees we run after tabsets.js — so search activation wins.
+    window.addEventListener("pageshow", function (event) {
+      if (!event.persisted) {
+        for (const mark of mainEl.querySelectorAll("mark")) {
+          openAllTabsetsContainingEl(mark);
+        }
+        requestAnimationFrame(() => scrollToFirstVisibleMatch(mainEl));
+      }
+    }, { once: true });
+
     // fix up the URL to remove the q query param
     const replacementUrl = new URL(window.location);
     replacementUrl.searchParams.delete(kQueryArg);
@@ -1170,7 +1183,7 @@ function searchMatches(inSearch, el) {
   /** @type {{i:number; els:Map<HTMLElement,{lo:number,hi:number}>}[]} */
   let curMatchContext = initMatch()
 
-  for (leaf of leafNodes) {
+  for (const leaf of leafNodes) {
     const leafStr = leaf.textContent.toLowerCase()
     // for each character in this leaf's text:
     for (let leafi = 0; leafi < leafStr.length; leafi++) {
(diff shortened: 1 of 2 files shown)


src/resources/projects/website/search/quarto-search.js
```diff
@@ -45,22 +45,7 @@ window.document.addEventListener("DOMContentLoaded", function (_event) {
...
-    // handler during module execution, before DOMContentLoaded. By registering
-    // ours during DOMContentLoaded, listener ordering guarantees we run after
-    // tabsets.js restores tab state from localStorage — so search activation
-    // wins over stored tab preference.
-    window.addEventListener("pageshow", function (event) {
-      if (!event.persisted) {
-        activateTabsWithMatches(mainEl);
-        // Let the browser settle layout after Bootstrap tab transitions
-        // before calculating scroll position.
-        requestAnimationFrame(() => scrollToFirstMatch(mainEl));
```
vezwork, line 61: this was moved inside of `highlight`

src/resources/projects/website/search/quarto-search.js
```diff
@@ -359,22 +343,12 @@ window.document.addEventListener("DOMContentLoaded", function (_event) {
...
 
             item({ item, createElement }) {
-              // process items to include text fragments as they are rendered
-              if (item.text && item.href && !item.href.includes(':~:text=')) {
-                // e.g. `item.text` for a search "def fiz": "bla bla bla<mark class='search-match'>def fiz</mark> bla bla"
-                const fullMatches = item.text.matchAll(/<mark class='search-match'>(.*?)<\/mark>/g)
-                // extract capture group with the search match
-                // result e.g. ["def fiz"]
-                const searchMatches = [...fullMatches].map(match => match[1])
-                if (searchMatches[0]) {
```
vezwork, line 369: We no longer use the matching text in the search, like we did for text fragments. Instead, for the `q` param, we use the user's full query, exactly as they typed it, whether algolia found a match or not. Algolia doesn't always find matches that the highlighting can, so this should help the user find anything that matches what they typed.

src/resources/projects/website/search/quarto-search.js
```diff
@@ -1123,151 +1097,189 @@ function clearHighlight(searchterm, el) {
...
 }
 
-function escapeRegExp(string) {
-  return string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"); // $& means the whole matched string
-}
-
-// After search highlighting, activate any tabs whose panes contain <mark> matches.
-// This ensures that search results inside inactive Bootstrap tabs become visible.
-// Handles nested tabsets by walking up ancestor panes and activating outermost first.
-function activateTabsWithMatches(mainEl) {
```
vezwork, line 1133: Replaced with `openAllTabsetsContainingEl`

cscheid:
Code LGTM but I believe we have a stale playwright test now.

cderv:
Thanks for the work on cross-node highlighting — the `searchMatches()` approach that handles text spanning decorated elements is a real improvement over the per-text-node regex matching.

So I am looking into this more, but the failing test is not stale, it just shows that this PR brings a regression compare to what I have tried to do in previous improvement. Here is my analysis for now

These tests were added in #14053 and cover specific edge cases in how search interacts with Bootstrap tabsets. They pass on current main (https://github.com/quarto-dev/quarto-cli/actions/runs/22241123723, after #14053 merged).

### What the tests cover

The test fixture has multiple tabsets with search terms placed strategically to exercise different scenarios. The two failing tests are:

1. "Search keeps active tab when it already has a match"

When a search term (gamma-both-tabs) exists in both the R tab (active by default) and the Python tab, the expected behavior is to leave the R tab active — it already shows a match, so switching tabs would be disruptive. The previous `activateTabsWithMatches()` grouped marks by their parent .tab-content container and checked whether the active pane already had a match before deciding to switch. openAllTabsetsContainingEl() doesn't have this check — it activates every .tab-pane ancestor of every match node, so the last one processed wins and the mark in the other tab becomes hidden.

2. "Search activation overrides localStorage tab preference" (line 79)

When the user has a stored tab preference in `localStorage` (e.g., "R" for grouped language tabs) but the search term (python-only-content) only matches in the Python tab, search should override the stored preference to show the match. This depends on event timing: `tabsets.js` registers a pageshow` handler that reads `quarto-persistent-tabsets-data` from `localStorage` and calls `toggleAll()` to restore previously-selected tabs. The previous code in #14053 handled this by registering tab activation as a `pageshow` listener during `DOMContentLoaded` — listener ordering guarantees it runs after `tabsets.js` restores stored state, so search activation wins. This PR runs tab activation synchronously during highlight() in `DOMContentLoaded`, so `tabsets.js`'s `pageshow` handler fires afterward and overwrites the search-driven tab state with the `localStorage` preference.

### What passes

The other 4 tests pass, which means the core mechanics work: activating an inactive tab for a match, nested tabset activation, not touching tabs when the match is outside tabs, and scrolling to the first match.

### Suggestion

The cross-node `searchMatches()` / `markMatches()` approach is the right direction — the old per-text-node regex couldn't handle matches spanning decorated elements. The tab activation logic just needs to be brought forward. Specifically:

1. Before activating a tab, check whether the currently active pane in that tabset already contains a mark. If so, skip it.
2. Move tab activation to a `pageshow` listener (registered during `DOMContentLoaded`) so it runs after tabsets.js restores localStorage state.

I haven't verified locally whether there are other behavioral differences beyond what the tests catch (e.g., how the setTimeout scroll interacts with tab transitions vs the previous requestAnimationFrame approach), but these two are the confirmed regressions from CI. 

Happy to discuss it. The demo website is there to show current behavior: https://cderv.github.io/quarto-search-highlight-demo/

cderv:
@vezwork I created https://github.com/quarto-dev/quarto-cli/pull/14089 to show a possible fix that I think keep your simplified logic, while making the tab handling worked as I initially tried. THe tests should be passing there. 

TBD.

vezwork:
Hey @cderv thanks for the fix. The tests are passing now. I'm ready to merge when you are.

cderv (approved):
I have updated changelog for this (including remove text fragment one) and also the demo website: https://cderv.github.io/quarto-search-highlight-demo/

So all good to merge !

PR #14093, opened 2026-02-23
Add group for options that need Typst specific content in docs

cwickham:
1. Adds typst to the document groups in `groups.yml` so that `document-typst.yml` options (logo, margin-geometry, theorem-appearance) are included when generating format reference pages.                                          
    
      Without this, the typst-specific options defined in  `document-typst.yml` were never read by `tools/reference.ts` in quarto-web, so they didn't appear on the typst reference page.
  
2. Moves the `typst` `margin` definition from `document-reveal-layout.yml` to `document-layout.yml`, resulting in two separate declarations of `margin`.  This pattern already exists for several other options in the schema so I'm hoping it doesn't cause problems.  

    @gordonwoodhull could you double-check the Typst schema for `margin` Claude proposed here?

Fix for #13412


cderv:
Awesome thank you !! 

Why Partial fix only ? 

I think this + the reference update covers it right ?

cwickham:
@cderv we still have a Slide Layout section on the typst page:

Screenshot 2026-02-23 at 9 45 23 AM

I'll add something to this PR for that shortly.

cwickham:
@cderv this should now close #13412. It's a bigger change, so please review the edited description. I've also tagged @gordonwoodhull for a review, particularly on the `margin` schema for `typst`.

gordonwoodhull (approved):
The margin schema looks good. 

I remember looking into grouping for this problem, and thinking it was not exactly the right tool, but I can't find notes/transcript of this. If it works we should use it!

cwickham:
After talking to @cscheid, this doesn't really work. Having `margin` in two places worked nicely for the docs, but caused trouble with the YAML validation. One schema would win, and documents using the other schema at the top level would then fail to validate.

This is now a smaller change:

* Add `typst-document.yml` to `groups.yml`. This ensures the options `margin-geometry` and `theorem-appearance` appear on the Reference pages.
* Move `margin` from `document-reveal-layout.yml` to `document-layout.yml`. This moves `margin` on the Reference pages for both Typst and Revealjs out of the "Slide Layout" section and into the "Layout" section.

cwickham:
This now closes the doc issue but revealed a bigger issue I did my best to document in: https://github.com/quarto-dev/quarto-cli/issues/14122

Issue #13781, opened 2025-12-11
Bug in axe accessibility checks for revealjs formats


paytonej:
### Description

Quarto >1.8 has integrated support for `axe-core` accessibility checking html. The docs say that the axe YAML metadata configuration can be added to HTML formats (`html`, `dashboard`, and `revealjs`).

However, adding axe to the yaml for the `revealjs` format results in an error in compiling the sass themes. The error is:
```{bash}
ERROR: Theme file compilation failed:

Error: Undefined variable.
   ╷
67 │   border: 1px solid $body-color;
```

Line 67 defining the border is in `quarto.js` This suggests that theme file sass compilation is occurring *after* quarto.js inclusion, though it needs to happen *before.*

Accessibility enhancements should be high-priority due to [new rules taking effect in April 2026 under Title II of the Americans with Disabilities Act](https://agb.org/news/agb-alerts/agb-policy-alert-ada-digital-accessibility-rule-requires-full-compliance-by-april-2026/) and the requirements of the [European Accessibility Act (EAA), which went into effect in June 2025](https://commission.europa.eu/strategy-and-policy/policies/justice-and-fundamental-rights/disability/union-equality-strategy-rights-persons-disabilities-2021-2030/european-accessibility-act_en). Many users are now liable for the documents they render meeting these requirements.

Related discussion items:

* [Ensure Quarto Meets WCAG 2.1 AA Standards #12644](https://github.com/orgs/quarto-dev/discussions/12644)
* [Increase options/guidance for compliance with Web Content Accessibility Guidelines (WCAG) in revealjs HTML output #8621](https://github.com/orgs/quarto-dev/discussions/8621)
* [https://github.com/orgs/quarto-dev/discussions/11847](https://github.com/orgs/quarto-dev/discussions/11847)

**Minimal (not) working example. Change format to "html" to create something that renders appropriately.**

````{md}
---
title: "Axe Causes Sass Compilation Error"
date: today
format: 
  revealjs:
    axe: true
---

## Axe test.

This violates contrast rules: [insufficient contrast.]{style="color: #eee"}.

This test works with the format as `html` but not with the format as `revealjs`.

## For revealjs, axe causes theme file compilations to fail.

The resulting error is:

```{bash}
ERROR: Theme file compilation failed:

Error: Undefined variable.
   ╷
67 │   border: 1px solid $body-color;
```

Line 67 defining the border is in `quarto.js`

[...]

PR #14095, opened 2026-02-23
Add `a11y` to schema groups

cwickham:
Adds `a11y` to `groups.yml` so options from `document-a11y.yml` appear on format reference pages on `quarto-web`.

### DIFF
--- a/src/resources/schema/groups.yml
+++ b/src/resources/schema/groups.yml
@@ -92,6 +92,8 @@ document:
     title: "Language"
   comments:
     title: "Comments"
+  a11y:
+    title: "Accessibility"
   includes:
     title: "Includes"
   metadata:


cwickham:
@cscheid This will mean `axe` will appear on the Reference pages from `html`, `revealjs`, and `dashboard`. Do we need to revisit the `$html-files` in the schema?

cscheid:
> @cscheid This will mean axe will appear on the Reference pages from html, revealjs, and dashboard. Do we need to revisit the `$html-files` in the schema?

Sorry, I don't understand your question!

cwickham:
The `axe` option is listed in the schema as:

```yaml
tags:
  formats: [$html-files]
```

But we have an open issue that it doesn't work in `revealjs`: https://github.com/quarto-dev/quarto-cli/issues/13781

Will we fix the bug for 1.9, or edit the schema to exclude `revealjs`?

cderv:
> Will we fix the bug for 1.9, or edit the schema to exclude revealjs?

Yes this will be fixed for 1.9, I am currently working on this

PR #14096, opened 2026-02-23
skip processing non-simple tables (#13776)

cscheid:
Closes #13776. It's more of a workaround than a real fix, but I don't want to change the `{from,to}_simple_table` round trip at this point in our release.


cwickham:
@cscheid If I understand correctly, this means I should swap out [this section](https://prerelease.quarto.org/docs/authoring/tables.html#column-widths-1) about using `tbl-colwidths` in the List Column docs, for list-tables native `widths` attribute. Correct?

PR #14110, opened 2026-02-24
fix: propagate fig-alt to Typst image alt text for PDF/UA-1

gordonwoodhull:
## Summary

- Fix `fig-alt` not being propagated to Typst's `image(alt: "...")` parameter for computed figures (R/Python code chunks)
- Refactor `figures.lua` to make `fig-alt` propagation format-agnostic instead of having separate HTML/LaTeX/Typst branches
- Add test case verifying `fig-alt` on R code chunks produces `image(alt: ...)` in Typst output

## Details

The `fig-alt` propagation in `quarto-pre/figures.lua` only handled HTML and LaTeX outputs. The Typst post-filter in `quarto-post/typst.lua` correctly looks for `image.attributes["alt"]` and generates `image(alt: "...")`, but the attribute was never set for Typst output.

The fix refactors the three format-specific branches into a single block:
- HTML: sets `alt` on the float itself (unchanged behavior)
- All other formats (LaTeX, Typst, etc.): walks into `float.content` to set `alt` on Image nodes

This means any future format will also automatically get `fig-alt` propagation.

Closes #14108

## Test plan

- [x] New test `typst-fig-alt-propagation.qmd` passes (verified `fig-alt` from R chunk → `image(alt: ...)` in .typ)
- [x] Existing `typst-image-alt-text.qmd` passes (caption-as-alt, explicit alt)
- [x] Existing `equations-alt.qmd` passes (html, pdf, typst)
- [x] Existing `ua-image-alt-text.qmd` passes (pdf, typst)

### DIFF
--- a/src/resources/filters/quarto-post/typst.lua
+++ b/src/resources/filters/quarto-post/typst.lua
@@ -227,7 +227,11 @@ function render_typst_fixups()
 
       -- Workaround for Pandoc not passing alt text to Typst image() calls
       -- See: https://github.com/jgm/pandoc/pull/11394
-      local alt_text = image.attributes["alt"]
+      -- Check fig-alt first (Quarto's custom alt text override), then alt, then caption
+      local alt_text = image.attributes[kFigAlt] or image.attributes["alt"]
+      if alt_text then
+        image.attributes[kFigAlt] = nil
+      end
       if (alt_text == nil or alt_text == "") and #image.caption > 0 then
         alt_text = pandoc.utils.stringify(image.caption)
       end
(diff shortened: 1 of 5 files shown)


gordonwoodhull:
Added a second fix: static (non-computed) images with `fig-alt` were also not getting alt text in Typst. Different code path — the Typst `Image` handler in `quarto-post/typst.lua` only checked the `alt` attribute, not Quarto's `fig-alt` attribute. The HTML post-processor already had this check; Typst was missing it.

New test in `typst-fig-alt-static-images.qmd` covers:
- Captioned image with `fig-alt` but no cross-ref label
- Uncaptioned image with `fig-alt` and no label

cderv:
FWIW I just removed backport label and moved to 1.9 for good tracking as your merged in main already.

PR #14112, opened 2026-02-25
Fix dark highlighting themes missing base text color

### LINKED ISSUE #14099: Dark syntax highlighting themes don't override light theme's base .sourceCode text color
## Description

When using light/dark theme toggling, certain dark syntax highlighting themes (zenburn, espresso) don't generate a base `.sourceCode > span` text color rule. The light theme's near-black base color persists in dark mode, making code tokens without a specific highlighting class invisible.

## Root Cause

The bug is in [`generateThemeCssClasses()`](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/command/render/pandoc-html.ts#L516-L587), not in the theme files. The function only emits the `.sourceCode > span` base color rule when it finds a `"Normal"` key in `"text-styles"`. It never checks the top-level `"text-color"` as a fallback.

[**kate**](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/kate.theme) (light theme) has `"Normal"` in `"text-styles"`:
```json
"text-styles": {
  "Normal": {"text-color": "#1f1c1b", ...}
}
```
→ Generates `.sourceCode > span { color: #1f1c1b; }`

[**zenburn**](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/zenburn.theme) (dark theme) is a valid Pandoc/KDE theme file — it defines `"text-color": "#cccccc"` at the **top level**, which is how these theme files can specify the default text color. But it has **no "Normal" entry in `"text-styles"`**:
```json
{
  "text-color": "#cccccc",
  "background-color": "#303030",
  "text-styles": {
    "Alert": {...},
    "Keyword": {...},
    ...
    // no "Normal" key
  }
}
```
→ No base `.sourceCode > span` rule is generated.

Since Quarto layers dark CSS on top of light (`// note: dark is layered on light, we don't disable primary!`), the light theme's `.sourceCode > span { color: #1f1c1b; }` persists unchallenged in dark mode.

Note: the triple-stylesheet issue from #13450 is unrelated — Puppeteer testing confirms the `quarto-color-scheme-extra` sheets are properly disabled by JS on page load and don't cause this problem.

## Affected Themes

7 themes are missing "Normal" in `text-styles`: [espresso](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/espresso.theme), [haddock](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/haddock.theme), [monochrome](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/monochrome.theme), [none](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/none.theme), [pygments](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/pygments.theme), [tango](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/tango.theme), [zenburn](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/resources/pandoc/highlight-styles/zenburn.theme).

Of the dark themes, **espresso** and **zenburn** are both affected — they define a top-level `text-color` but no "Normal" style. Any user-created custom theme with the same valid structure would also break silently.

## Reproduction

```yaml
# _quarto.yml
format:
  html:
    theme:
      light: cosmo
      dark: darkly
    highlight-style:
      light: kate
      dark: zenburn
```

Any code block with tokens that don't get a specific span class (e.g., plain identifiers, punctuation) will be unreadable in dark mode:

## Proposed Fix

In [`generateThemeCssClasses()`](https://github.com/quarto-dev/quarto-cli/blob/d1c3de0e03e80dde952b2f84a93aa469fa0c02d1/src/command/render/pandoc-html.ts#L516-L587), after processing `text-styles`, fall back to the theme's top-level `"text-color"` when no "Normal" style was found:

```typescript
// After the Object.keys(textStyles).forEach loop:
if (!tokenCssByAbbr[""] && themeJson["text-color"]) {
  const cssValues = [`  color: ${themeJson["text-color"]};`, "  font-style: inherit;"];
  toCSS("", "Normal", cssValues);
  ["pre > code.sourceCode > span", "code.sourceCode > span",
   "div.sourceCode,\ndiv.sourceCode pre.sourceCode"]
    .forEach((selector) => {
      otherLines.push(`\n${selector} {`);
      otherLines.push(...cssValues);
      otherLines.push("}\n");
    });
}
```

Verified by creating a patched zenburn theme with "Normal" added (simulating the fix output) — dark highlighting CSS then emits `.sourceCode > span { color: #cccccc; }` and all code tokens are readable without any custom SCSS workarounds.

## Workaround

```scss
body.quarto-dark {
  pre.sourceCode,
  .sourceCode > span,
  code.sourceCode > span {
    color: #dcdccc;
  }
}
```

## Environment

- Quarto 1.8.27
- macOS, Chrome/Safari

jdonaldson:
## Summary

Fixes #14099

Dark syntax highlighting themes like **zenburn** and **espresso** don't set a base text color when paired with a light theme. Unhighlighted code tokens inherit the light theme's near-black color, making them invisible against the dark background.

### Root cause

Two issues:

1. **`$code-block-color` never emitted for paired themes** (`format-html-scss.ts`): The `text-color` extraction from the theme JSON was gated by `!themeDescriptor.isAdaptive`. When users specify separate light/dark themes (`highlight-style: {light: kate, dark: zenburn}`), `isAdaptiveTheme()` returns `true` because the config object has both `light` and `dark` keys — even though neither kate nor zenburn is actually an adaptive theme. So `$code-block-color` was never set as an SCSS variable.

2. **No span-level color rule** (`_bootstrap-rules.scss`): Even with `$code-block-color` set, it was only applied to container elements (`div.sourceCode`, `div.sourceCode pre.sourceCode`). The light theme's syntax highlighting CSS generates `code.sourceCode > span { color: #1f1c1b; }` which has higher specificity than container inheritance.

### Fix

1. **`format-html-scss.ts`**: Move the `text-color` → `$code-block-color` extraction outside the `!isAdaptive` guard. Adaptive themes handle their own background-color, but text-color still needs to propagate for Bootstrap rules to use.

2. **`_bootstrap-rules.scss`**: Add span-level selectors (`pre > code.sourceCode > span`, `code.sourceCode > span`) alongside the existing container-level `$code-block-color` rules. This follows the same pattern as the existing background-color handling.

### Affected themes

7 built-in themes missing `"Normal"` in `text-styles`: espresso, haddock, monochrome, none, pygments, tango, zenburn. Also covers user-created custom themes with the same structure.

### Verified

Built quarto from source and tested:
- `{light: kate, dark: zenburn}` → dark CSS now includes `code.sourceCode > span { color: #ccc; }`
- `{light: kate, dark: espresso}` → dark CSS now includes `code.sourceCode > span { color: #bdae9d; }`
- Light mode unchanged (kate's rules still present, no regressions)

cc @mcanouil — this follows the `_bootstrap-rules.scss` pattern you suggested, plus fixes an upstream gating issue where `$code-block-color` was never emitted for paired (non-adaptive) themes.

### DIFF
--- a/src/format/html/format-html-scss.ts
+++ b/src/format/html/format-html-scss.ts
@@ -314,7 +314,13 @@ export function resolveTextHighlightingLayer(
           true,
         );
       }
+    }
 
+    // Inject a text color regardless of adaptive status. Adaptive themes
+    // handle their own bg, but when separate light/dark themes are paired
+    // (e.g. kate + zenburn), the dark theme still needs its text-color
+    // emitted as $code-block-color so Bootstrap rules can apply it.
+    if (themeDescriptor) {
       const textColor = themeDescriptor.json["text-color"] as string;
       if (textColor) {
         layer.defaults = layer.defaults + "\n" + outputVariable(
--- a/src/resources/formats/html/bootstrap/_bootstrap-rules.scss
+++ b/src/resources/formats/html/bootstrap/_bootstrap-rules.scss
@@ -863,6 +863,14 @@ div.sourceCode {
   div.sourceCode pre.sourceCode {
     color: $code-block-color;
   }
+  // Themes that omit "Normal" from text-styles (e.g. zenburn, espresso)
+  // don't generate a span-level color rule in the syntax highlighting CSS.
+  // When layered with a light theme that does, the light theme's near-black
+  // span color persists in dark mode. This ensures a base span color is set.
+  pre > code.sourceCode > span,
+  code.sourceCode > span {
+    color: $code-block-color;
+  }
 }
 
 pre.sourceCode {


mcanouil:
I don't think that's the right approach here.

You are fixing only a symptom not the root cause `isAdaptiveTheme`.

See the other issues and the last one I opened on the matter 

- https://github.com/quarto-dev/quarto-cli/issues/14135

jdonaldson:
Withdrawing this PR as well — the CLA requirements are more than I'm comfortable signing for a bug fix contribution. The fix and root cause analysis are documented in #14099 if someone on the team wants to pick it up. Thanks for the great project, wishing you all well.

PR #14120, opened 2026-02-26
fix: code-tools show/hide all code buttons, and allow them to act on cells with `output: asis`

benlubas:
## Description

Originally this fix was to allow code cells with `output: asis` to open/close with the codetools menu buttons. But I rebased it onto main and found out that code-tools show/hide all code doesn't work at all. So I've fixed that too.

I'm also like 90% sure that the rest of the function was dead code, so I've removed it. If it at one point did something, it no longer works. Let me know if it was supposed to do something I can add it back.

This is one of the fixes mentioned in #13946 (I would still appreciate feedback on that discussion).

## Checklist

I have (if applicable):

- [ ] added new tests


cscheid:
(Hi, we're pretty swamped with sorting out the 1.9 release. We'll come back to this in a few. Thanks for the patience!)

PR #14128, opened 2026-02-27
refactor(definitions.typ): use single rule for definition item

mcanouil:
Consolidate the item display rule.

Follow up on:

- https://github.com/quarto-dev/quarto-cli/pull/13978

Upstream:

- https://github.com/jgm/pandoc/pull/11436


cderv (approved):
Test are passing, and it is already merged upstream. Seems ok to merge for 1.9. 

@gordonwoodhull what do you think ?

gordonwoodhull:
Absolutely. Took me a minute to understand, because I thought we had already done this. 

But this is something we discussed after merging #13978, but never implemented here.

PR #14136, opened 2026-03-01
Fix duplicate schema definitions causing YAML validation issues

### LINKED ISSUE #14122: Duplicate schema definitions cause YAML validation issues
While working on https://github.com/quarto-dev/quarto-cli/pull/14093 I found a few places where we have more than one schema for an option. In practice, this means we can end up with validation failures when someone specifies the "losing" schema at the top-level of the document options.

I found four options with this problem. In two cases, the more permissive schema wins or the schemas are the same, so these aren't currently causing any issues. In two cases, there are situations where I can generate YAML validation errors with options that should be OK. 

### 1. `logo` — **Conflict**

Defined in: `document-options.yml` (beamer), `document-dashboard.yml`, `document-reveal-content.yml`, `document-typst.yml`.

Typst's `logo-light-dark-specifier-path-optional` schema wins at top level. Unlike dashboard/revealjs's `logo-light-dark-specifier`, it doesn't accept `false`.

**Fails:**
```yaml
---
title: Test
logo: false
format: dashboard
---
```
> ERROR: Field "logo" has value false, which must instead be a string

### 2. `subject` — **Conflict**

Defined in: `document-metadata.yml` (pdf/office/odt), `document-epub.yml`

Metadata's `string` schema wins at top level. Epub's `anyOf: [string, object]` is not used.

**Fails:**
```yaml
---
title: Test
subject:
  text: "Science"
  authority: "BISAC"
  term: "SCI000000"
format: epub
---
```
> ERROR: Field "subject" has value ... The value must instead be a string.

### 3. `copyright` — **No conflict**

Defined in: `document-attributes.yml` (jats), `document-metadata.yml` (html-doc/jats)

The more permissive `anyOf: [object, string]` schema wins. No validation failures.

### 4. `footer` — **No conflict**

Defined in: `document-includes.yml` (man), `document-reveal-content.yml` (revealjs)

Both are `string`. No validation failures.

jdonaldson:
## Summary

Fixes #14122.

When the same option name (e.g., `logo`, `subject`) is defined in multiple schema YAML files, `objectRefSchemaFromGlob()` silently overwrites — last writer wins. This causes YAML validation errors when users specify valid values that match a "losing" schema variant.

- **`logo`**: Add `enum: [false]` to `logo-light-dark-specifier-path-optional` so `logo: false` validates for all formats (dashboard, revealjs, typst), matching the existing `logo-light-dark-specifier` behavior
- **`subject`**: Consolidate into `document-metadata.yml` as `anyOf: [string, object]` covering both simple (pdf/office) and structured (epub) forms; remove duplicate entry from `document-epub.yml`
- Regenerated all build artifacts (`json-schemas.json`, `schema-types.ts`, editor tooling)
- Added unit tests for both cases

## Test plan

- [x] `schema-schema.test.ts` — 3/3 pass (schema YAML files well-formed)
- [x] `schema-files.test.ts` — 4/4 pass (2 existing + 2 new test cases)
  - `logo-false.yml` — validates `false` accepted by `logo-light-dark-specifier-path-optional`
  - `subject-structured.yml` — validates structured object form passes schema validation
- [ ] Manual verification: `logo: false` in dashboard format renders without validation error
- [ ] Manual verification: structured `subject` in epub format renders without validation error


jdonaldson:
Just adding a small comment that there's a human behind this PR.  I used Claude here to make some small changes, which were reviewed before submitting.  Tests are also added.

gordonwoodhull:
This pointed out a minor (possibly harmless) bug in `resolveLogo` so I've added a fix.

gordonwoodhull (approved):
I've reviewed the `logo` part (my bug) - looks good, adding a tiny fix. 

The `subject` resolution also makes sense to me.

Claude did not think the structured epub subject test was a rigorous enough test, so we've also added a smoke-all test. I'll let it comment on that.

gordonwoodhull (commented):
**End-to-end epub test for structured `subject`** — The `subject-structured.yml` schema test inlines a copy of the schema rather than referencing it (unlike the `logo-false.yml` test which uses `ref:`). This means it validates the *shape* of the anyOf but not the actual schema Quarto uses — if someone later changes `document-metadata.yml`, this test would still pass against its own detached copy. Added `subject-structured-epub.qmd` as a smoke-all test that renders an actual epub with structured subject metadata and verifies the OPF output contains the expected `dc:subject`, `authority`, and `term` elements. This exercises the full pipeline: schema validation → metadata resolution → Pandoc rendering.

cscheid:
Hello, and thanks for the PR! We will need you to fill out the original template for the PR, specifically because we need you to attest you've signed the CLA for this project.

jdonaldson:
Thanks for the review and the additional smoke test, @gordonwoodhull — appreciated.

@cscheid — I took a look at the individual CLA and it's more than I'm comfortable with for a bug fix. Full copyright assignment, blanket patent non-assertion across all Posit software, and open-ended obligations (clause 5) are a significant ask for an external contributor. A DCO or license-grant CLA would be more proportionate, but I understand if that's not on the table.

I'm going to withdraw this PR. The fix is straightforward and well-documented in #14122, so hopefully it's easy for someone on the team to pick up. Wishing you all well — Quarto is a great project.

PR #14142, opened 2026-03-02
Fix caption used as fallback alt text in PDF/UA

### LINKED ISSUE #14107: Caption should not be used as fallback alt text for images in PDF/Typst
## Summary

When an image has a caption but no explicit `fig-alt`, Quarto silently copies the caption into the image's alt text attribute for LaTeX/PDF and Typst output. This was introduced as part of PDF/UA compliance work but is an accessibility anti-pattern.

## The problem

Captions and alt text serve fundamentally different purposes:

- **Caption**: Visible text describing the figure's significance in context (e.g., "Figure 3: Comparison of methods A and B showing the 2x improvement")
- **Alt text**: Non-visible text describing **what the image looks like** for screen readers (e.g., "A bar chart with two bars labeled Method A and Method B, where B is approximately twice the height of A")

Using the caption as alt text violates WCAG guidance, which explicitly notes that when a caption is already present, the alt text should describe the image content rather than duplicate the caption. The current behavior silences PDF/UA validators without actually improving accessibility.

## Affected code

The fallback logic exists in three places:

1. **`src/resources/filters/layout/latex.lua:335-337`** (and `:387-389`, `:672-674`) — three separate code paths that copy the stringified caption into `image.attributes["alt"]` before clearing the caption for LaTeX rendering
2. **`src/resources/filters/quarto-post/typst.lua:228-232`** — same pattern for Typst output

All introduced in commits `a867c3c24` (Jan 9, 2026, "Add pdf-standard option with verapdf validation") and `ba75b374f` (Jan 10, 2026, "Add Typst image alt text for PDF/UA accessibility").

## Note: explicit `fig-alt` works correctly

The explicit `fig-alt` propagation in `src/resources/filters/quarto-pre/figures.lua:31-53` is fine — it correctly propagates user-specified alt text. The problem is only in the **fallback** when no `fig-alt` is provided.

## Suggested fix

Remove the caption-as-alt fallback and instead either:

- Emit no alt text, letting PDF/UA validators flag it so authors add proper `fig-alt`
- Emit a build warning when an image has a caption but no `fig-alt`, prompting the author to add one

gordonwoodhull:
## Summary

Removes the anti-pattern where figure captions are copied into image alt text for PDF/UA compliance. Captions describe a figure's significance in context; alt text describes what the image looks like. Using one as the other is an accessibility anti-pattern that merely silences validators without helping screen reader users.

This was introduced in the Jan 2026 PDF/UA work (commits `a867c3c24` and `ba75b374f`) to satisfy PDF/UA validators, which require every `<Figure>` structure element to have an `/Alt` string.

Fixes #14107

## What changed

**LaTeX** (`latex.lua`): Removed 3 caption-as-alt blocks that copied `image.caption` into `image.attributes["alt"]` before clearing the caption for separate rendering.

**Typst** (`pandoc3_figure.lua`, `floatreftarget.lua`, `typst.lua`): Used a marker attribute (`_quarto_no_caption_alt`) to prevent the caption-as-alt fallback for figure images, while preserving it for inline images where `image.caption` IS the standard markdown alt text (`alt text` in running text).

The distinction matters because of how Pandoc 3 represents alt text in its AST.

## Pandoc 3 AST analysis

In Pandoc 3, `{alt="text"}` on an image does **not** set `image.attributes["alt"]` — it **replaces** `image.caption` with the alt value. The `![visible caption]` text moves to `figure.caption.long`, while `image.caption` becomes the alt text. This means `image.caption` serves double duty:

- **Inline images**: `image.caption` IS the alt text (the `![...]` content is alt by the HTML spec)
- **Figure images**: `image.caption` is a copy of the visible caption (unless overridden by `{alt="..."}`)

To distinguish these cases, we compare `image.caption` to the figure's visible caption (`figure.caption.long` or `float.caption_long`). If they match, no explicit alt was provided — we set `_quarto_no_caption_alt` to suppress the fallback. If they differ, `{alt="..."}` was used and `image.caption` IS the explicit alt text.

This is the same heuristic Pandoc's own Markdown writer uses (confirmed via [Pandoc source](https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Writers/Markdown.hs) — it compares image alt to figure caption to decide whether to emit `{alt="..."}`). Pandoc itself acknowledges the ambiguity: the AST cannot distinguish "caption that happens to match alt" from "no explicit alt provided." The recommended workaround is Quarto's `fig-alt` attribute, which flows through a separate, unambiguous path.

## How `fig-alt` vs `alt` work

- **`{alt="text"}`**: Pandoc replaces `image.caption`, no separate attribute. Works but ambiguous when caption and alt are intentionally identical.
- **`{fig-alt="text"}`**: Quarto stores as `image.attributes["fig-alt"]`, propagated explicitly to `\includegraphics[alt=...]` (LaTeX) or `image(alt: "...")` (Typst). Always unambiguous. This is the recommended approach.

## Test plan

- `caption-not-alt-ua.qmd` — verifies caption is NOT copied to `\includegraphics[alt=]` and that Quarto warns about missing alt text
- `typst-image-alt-text.qmd` — TC1 moved to must-not-match; TC8 added for `fig-alt`
- `ua-image-alt-text.qmd` — now uses explicit `fig-alt` 
- All 39 pdf-standard tests pass

### DIFF
--- a/src/command/render/latexmk/parse-error.ts
+++ b/src/command/render/latexmk/parse-error.ts
@@ -125,8 +125,10 @@ export function findPdfAccessibilityWarnings(
 
   // Match: Package tagpdf Warning: Alternative text for graphic is missing.
   //        (tagpdf)                Using 'filename' instead.
+  // Note: tagpdf wraps long filenames across multiple (tagpdf) continuation
+  // lines, so we allow optional line breaks with (tagpdf) prefixes.
   const altTextRegex =
-    /Package tagpdf Warning: Alternative text for graphic is missing\.\s*\n\(tagpdf\)\s*Using ['`]([^'`]+)['`] instead\./g;
+    /Package tagpdf Warning: Alternative text for graphic is missing\.\s*\n\(tagpdf\)\s*Using ['`]([^'`]+)['`]\s*(?:\n\(tagpdf\)\s*)?instead\./g;
   let match;
   while ((match = altTextRegex.exec(logText)) !== null) {
     result.missingAltText.push(match[1]);
--- a/src/command/render/latexmk/pdf.ts
+++ b/src/command/render/latexmk/pdf.ts
@@ -204,7 +204,7 @@ async function initialCompileLatex(
       if (accessibilityWarnings.missingAltText.length > 0) {
         const fileList = accessibilityWarnings.missingAltText.join(", ");
         warning(
-          `PDF accessibility: Missing alt text for image(s): ${fileList}. Add alt text using ![alt text](image.png) syntax for PDF/UA compliance.\n`,
+          `PDF accessibility: Missing alt text for image(s): ${fileList}. Add alt text using fig-alt in YAML or {fig-alt="description"} on the image for PDF/UA compliance.\n`,
         );
       }
       if (accessibilityWarnings.missingLanguage) {
--- a/src/resources/filters/customnodes/floatreftarget.lua
+++ b/src/resources/filters/customnodes/floatreftarget.lua
@@ -978,6 +978,22 @@ end, function(float)
   local kind = "quarto-float-" .. ref
   local supplement = titleString(ref, info.name)
 
+  -- For figures: mark images so typst.lua won't use caption-as-alt fallback
+  -- when caption IS the visible figure caption (not an explicit alt override).
+  -- In Pandoc 3, {alt="text"} replaces image.caption with the alt value,
+  -- so image.caption != float.caption means an explicit alt was provided.
+  if ref == "fig" then
+    local float_caption_text = pandoc.utils.stringify(float.caption_long or {})
+    float.content = _quarto.ast.walk(float.content, {
+      Image = function(img)
+        if pandoc.utils.stringify(img.caption) == float_caption_text then
+          img.attributes["_quarto_no_caption_alt"] = "true"
+        end
+        return img
+      end
+    })
+  end
+
   -- Inject show rule to left-align listing figures (only once per document)
   -- This overrides any template centering for listing-kind figures
   -- https://github.com/quarto-dev/quarto-cli/issues/9724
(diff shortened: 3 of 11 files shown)


gordonwoodhull:
## Second commit: Fix tagpdf warning parsing and document UA-2 structural issue

The second commit (`32d1172ab`) addresses two things discovered while testing:

### 1. tagpdf warning regex fix (`parse-error.ts`)

The regex that parses tagpdf's "Alternative text for graphic is missing" warning assumed the filename and `instead.` appeared on a single `(tagpdf)` continuation line. When filenames are long, tagpdf wraps across multiple lines:

```
Package tagpdf Warning: Alternative text for graphic is missing.
(tagpdf)                Using 'caption-not-alt-ua_files/mediabag/penrose.pdf'
(tagpdf)                instead.
```

Fixed the regex to allow an optional `(tagpdf)` line break before `instead.`. Also updated the warning message to recommend `fig-alt` instead of the old `alt text` advice (which was itself the caption-as-alt anti-pattern).

### 2. UA-2 structural issue with unlabeled captioned figures (`ua2-unlabeled-figure-caption.qmd`)

Discovered that unlabeled captioned figures (`caption` without `{#fig-label}`) produce invalid UA-2 structure. These go through `pandoc3_figure.lua` → `latexImageFigure()` → bare `\begin{figure}[H]`, and tagpdf places `<Caption>` as a sibling of `<Figure>` directly under `<Document>`:

```
/Document
  /Caption    ← UA-2 violation: must be inside a grouping element
  /Figure     ← orphaned from its caption
```

Labeled figures (`{#fig-label}`) go through FloatRefTarget, which wraps in a `\Div` that provides the grouping context:

```
/Document
  /Div
    /Caption  ← properly nested ✓
    /Figure
```

This is a pre-existing LaTeX/tagpdf limitation, not caused by our changes. Added `ua2-unlabeled-figure-caption.qmd` to document it with the expected verapdf validation warning, and updated `caption-not-alt-ua.qmd` to use a labeled figure to avoid conflating the two issues.

PR #14150, opened 2026-03-03
Fix `quarto preview subdir/file.qmd` crashing with doubled path

cderv:
`quarto preview subdir/page.qmd` in a website project crashes with `readfile subdir\subdir\page.qmd` — a doubled path. `quarto render` works fine.

## Root Cause

Preview makes two render calls: `renderFormats` (receives the original relative path) then `renderProject` (normalizes to absolute). The first call creates a cached `ExecutionTarget` with `source: "subdir/page.qmd"`. The second gets a cache hit with the stale relative source.

`projectPath()` was designed for output filenames — it resolves relative paths via `join(dirname(target.source), path)`.
https://github.com/quarto-dev/quarto-cli/blob/0f97e0540fbb1c28dd9d0215095e7bcca937d531/src/command/render/render.ts#L389-L405

But then, `context.target.source` itself is passed through this function.
https://github.com/quarto-dev/quarto-cli/blob/0f97e0540fbb1c28dd9d0215095e7bcca937d531/src/command/render/render.ts#L414

When `target.source` was relative, `dirname + join` doubled the subdirectory:

```
join(dirname("subdir/page.qmd"), "subdir/page.qmd") → "subdir/subdir/page.qmd"
```

`projectPath(finalOutput!)` is used correctly though — output filenames like `page.html` do need the `dirname + join` resolution.
https://github.com/quarto-dev/quarto-cli/blob/0f97e0540fbb1c28dd9d0215095e7bcca937d531/src/command/render/render.ts#L424

## Fix

Split `projectPath` into two functions with distinct responsibilities:

- `projectRelativeInput(sourcePath)` — for the source file path. Uses `normalizePath` directly, which handles both relative and absolute paths, because our `normalizePath` function does make path absolute if not.
- `projectOutputPath(path)` — for output filenames (like `page.html`). Preserves the existing `dirname + join` logic since output names are relative to the source directory.

## Bisect

Last good: v1.9.17. First bad: v1.9.18 (77633b39a).

## Related

Follow up on preview fixes that introduced this

- https://github.com/quarto-dev/quarto-cli/pull/13967 (for issue https://github.com/quarto-dev/quarto-cli/issues/13955)

which was a regression of broader quarto preview behavior change

- https://github.com/quarto-dev/quarto-cli/pull/13804

Also, in a way related to this observation we've done since new single project context is out

- https://github.com/quarto-dev/quarto-cli/pull/12401

### DIFF
--- a/src/command/render/render.ts
+++ b/src/command/render/render.ts
@@ -385,8 +385,24 @@ export async function renderPandoc(
           ));
       }
 
-      // if there is a project context then return paths relative to the project
-      const projectPath = (path: string) => {
+      // Compute the project-relative path for the input source file.
+      // Uses normalizePath to handle both relative and absolute source paths.
+      const projectRelativeInput = (sourcePath: string) => {
+        if (context.project) {
+          return relative(
+            normalizePath(context.project.dir),
+            normalizePath(sourcePath),
+          );
+        }
+        return sourcePath;
+      };
+
+      // Resolve an output file path to a project-relative path.
+      // Output paths (like "page.html") are relative to the source file's
+      // directory, so we join with dirname(target.source) before computing
+      // the project-relative result. Absolute output paths pass through
+      // normalizePath directly.
+      const projectOutputPath = (path: string) => {
         if (context.project) {
           if (isAbsolute(path)) {
             return relative(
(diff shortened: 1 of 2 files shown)


cderv:
@cscheid tests are passing, and I think this is a change that make sense and solve the problem. 

I can now run preview correct. 

So if good to you, please merge and build a new prerelease.

Issue #14061, opened 2026-02-18
Meta shortcode doesn't preserve line breaks in values


cwickham:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

Line breaks in a value that is used with the `meta` shortcode are not preserved

### Steps to reproduce

````qmd
---
key: |
  Something with a line break 

  I want to preserve.
---

``` yaml
{{< meta key >}}
```


````

### Actual behavior

Output consistent with:

````markdown
``` yaml
Something with a line break I want to preserve.
```
````

### Expected behavior

Output consistent with:

````markdown
``` yaml
Something with a line break 

I want to preserve.
```
````

### Your environment

- OS Mac Tahoe 26.3

### Quarto check output

```bash
Quarto 1.9.21
[✓] Checking environment information...
      Quarto cache location: /Users/charlottewickham/Library/Caches/quarto
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.4.5: OK
      Typst version 0.14.2: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.9.21
      Path: /Applications/quarto/bin

[✓] Checking tools....................OK
      TinyTeX: v2025.12
      Chromium: (not installed)
      Chrome Headless Shell: (not installed)
      VeraPDF: (not installed)

[✓] Checking LaTeX....................OK
      Using: TinyTex
      Path: /Users/charlottewickham/Library/TinyTeX/bin/universal-darwin
      Version: 2025

[✓] Checking Chrome Headless....................OK
      Using: Chrome found on system
      Path: /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
      Source: MacOS known location

[✓] Checking basic markdown render....OK

[✓] Checking R installation...........OK
      Version: 4.5.1
      Path: /Library/Frameworks/R.framework/Resources
      LibPaths:
        - /Users/charlottewickham/Library/R/arm64/4.5/library
        - /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library
      knitr: 1.50
      rmarkdown: 2.29

[✓] Checking Knitr engine render......OK

[...]

PR #14160, opened 2026-03-05
Fix meta shortcode truncating multi-word values in raw HTML contexts

### LINKED ISSUE #14159: Meta shortcode truncates multi-word values in raw HTML contexts
`{{< meta key >}}` drops everything after the first space when the value appears in raw HTML contexts (include-after-body text blocks, raw HTML blocks). Inline paragraph usage works correctly. This is a regression introduced by #14073.

## Repro

`_quarto.yml`:
```yaml
project:
  type: website
my-value: Release Candidate
```

`index.qmd`:
````markdown
---
format: html
include-after-body:
  text: |
    <script>window["test-include"] = '{{< meta my-value >}}';</script>
---

Inline: {{< meta my-value >}}

```{=html}
<script>window["test-rawblock"] = '{{< meta my-value >}}';</script>
```
````

Output:
```
Inline: Release Candidate           ← correct
window["test-rawblock"] = 'Release' ← truncated
window["test-include"] = 'Release'  ← truncated
```

## Root Cause

#14073 (commit f6ecb1169) added a `text` context branch to `handleMeta()` at `src/resources/filters/quarto-pre/shortcodes-handlers.lua:270`:

```lua
if type(optionValue) == "table" and #optionValue > 0 and optionValue[1].t == "Str" then
  return optionValue[1].text
```

For `"Release Candidate"` = `[Str("Release"), Space, Str("Candidate")]`, only `optionValue[1].text` ("Release") is returned. The condition `#optionValue > 0` matches all multi-word strings, but was intended only for single-node pandoc-native values.

## Version

Works in v1.9.23, broken since v1.9.24 (first release containing f6ecb1169).

## Real-world Impact

quarto.org download buttons show "1.9.29 Release" instead of "1.9.29 Release Candidate" because `prerelease-mode` is set via `include-after-body` text block.

cderv:
When `{{< meta key >}}` appears in raw HTML contexts (include-after-body text blocks, raw HTML blocks), multi-word values are truncated to the first word. `"Release Candidate"` becomes `"Release"`.

## Root Cause

Shortcodes in raw text (RawBlock, CodeBlock, include text) are resolved by an lpeg parser that calls the handler with `context = "text"` and then stringifies the result (`customnodes/shortcodes.lua:247-248`). The `text` branch added by #14073 checked `#optionValue > 0` to handle single pandoc-native Str nodes, but this matched any non-empty Inlines list. For `[Str("Release"), Space, Str("Candidate")]`, only `optionValue[1].text` was returned.

## Fix

Narrow the guard to `#optionValue == 1` (exact single-Str match) and add an Inlines branch using `pandoc.utils.stringify()` which concatenates all inline elements. The Blocks path (`pandoc.write(markdown)`) is unchanged, preserving #14061 line-break behavior.

Fixes #14159

### DIFF
--- a/.claude/rules/changelog.md
+++ b/.claude/rules/changelog.md
@@ -65,7 +65,9 @@ For external contributors (not core team):
 
 ## Regression Fixes
 
-**What qualifies:** Bugs introduced in recent versions (same major.minor)
+**What qualifies:** Bugs introduced in a previous major.minor version (e.g., a 1.8 feature that broke in 1.9)
+
+**Skip changelog entirely** when the regression was introduced by another change in the same version cycle. For example, if a 1.9 bug fix introduced a new bug also fixed in 1.9, the fix does not need a changelog entry — the original entry covers the feature area.
 
 **Placement:** Always at TOP of the file, before Dependencies
 
--- a/.claude/rules/testing/smoke-all-tests.md
+++ b/.claude/rules/testing/smoke-all-tests.md
@@ -63,6 +63,21 @@ _quarto:
         - ['\\begin\{figure\}']
 ```
 
+### Regex Match Array Format
+
+`ensureFileRegexMatches` (and variants like `ensureTypstFileRegexMatches`, `ensureLatexFileRegexMatches`) takes two arrays:
+
+```yaml
+ensureFileRegexMatches:
+  -                                   # First array: patterns that MUST match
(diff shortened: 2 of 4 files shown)


cscheid:
The service we use to get placeholder images seems to be failing. Another thing to replace with our own...

PR #14165, opened 2026-03-06
Replace fig-pos [H] with [h] when PDF tagging is active

### LINKED ISSUE #14164: Code chunk figures with fig-pos='H' produce broken PDF tag structure with pdf-standard: ua-2
## Summary

When an R code chunk produces a figure with `fig-cap` and a cross-reference label, the resulting PDF has incorrect tag structure: `/Caption` and `/Figure` appear as direct children of `/Document` instead of being nested inside `/figures` → `/float`. This causes a PDF/UA-2 validation failure: `<Document> shall not contain <Caption>`.

Static markdown images with cross-references produce the correct structure. Only code chunk figures are affected.

## Reproducer

````qmd
---
title: "Code chunk figure tag structure"
lang: en
format:
  pdf:
    pdf-standard: ua-2
---

```{r}
#| label: fig-mpg-by-weight
#| fig-cap: "MPG vs weight by cylinder count"
library(ggplot2)
ggplot(mtcars, aes(wt, mpg, color = factor(cyl))) +
  geom_point() +
  geom_smooth(method = "lm")
```

See @fig-mpg-by-weight.
````

**Expected:** verapdf validation passes (as it does for static images with cross-refs).
**Actual:** verapdf reports `<Document> shall not contain <Caption>`.

## Root cause

Both execution engines automatically inject `fig-pos='H'` for code chunk figures when code is echoed:

- **Knitr**: `src/resources/rmd/hooks.R:661-667` — sets `fig-pos='H'` when `echo=TRUE`, no user `fig-pos`, no layout options, and LaTeX output
- **Jupyter**: `src/core/jupyter/jupyter.ts:1625-1631` — same logic

The `[H]` specifier (from LaTeX's `float` package) redefines the figure environment in a way that's incompatible with lualatex's PDF tagging system. Standard placement specifiers use LaTeX's native float mechanism, which the tagging system handles correctly.

### Empirical testing of placement specifiers

Tested each placement specifier in isolation with `pdf-standard: ua-2` and verapdf validation:

| Specifier | verapdf result |
|-----------|---------------|
| (none) | **PASSED** |
| `[htbp]` | **PASSED** |
| `[h]` | **FAILED** — `<Document> shall not contain <Caption>` |
| `[H]` | **FAILED** — `<Document> shall not contain <Caption>` |

## Proposed fix

When PDF tagging is active (`pdf-standard` includes a standard that requires tagging, like `ua-2`), replace `fig-pos='H'` with `fig-pos='htbp'`. This preserves the intent (keeping figures near code) while using LaTeX's native float mechanism that the tagging system handles correctly.

### Files to modify

1. **`src/format/pdf/format-pdf.ts`** — expose tagging state as metadata so Lua filters can read it
2. **`src/resources/filters/quarto-pre/figures.lua`** — replace `fig-pos='H'` with `'htbp'` when tagging is active
3. **`src/resources/filters/quarto-post/latex.lua`** — same change in `handle_panel_layout` which also forces `[H]`

## Workaround

Setting `echo: false` on the code chunk avoids the `[H]` injection, since the placement logic only triggers when code is echoed.

gordonwoodhull:
## Summary

- Fix broken PDF tag structure for figures when `pdf-standard: ua-2` is set
- Replace `[H]` float placement with `[h]` when PDF tagging is active, across all filter pipeline locations
- Also fixes the previously-known non-cross-referenceable figure caption issue

## Details

The `[H]` specifier (from LaTeX's `float` package) redefines the figure environment in a way that breaks lualatex's PDF tagging system — `/Caption` and `/Figure` end up as direct children of `/Document` instead of being properly nested. This causes verapdf to report `<Document> shall not contain <Caption>`.

Multiple code paths inject `[H]`: knitr/Jupyter execution engines (for echoed code chunks), `pandoc3_figure.lua` (for non-cross-referenceable figures), and `quarto-post/latex.lua` (for panel layouts). Cross-referenceable static images were unaffected only because they happen not to get `[H]` — the LaTeX output is structurally identical in all cases.

Empirical testing of placement specifiers with `pdf-standard: ua-2`:

| Specifier | verapdf result |
|-----------|---------------|
| (none) | PASSED |
| `[h]` | PASSED |
| `[htbp]` | PASSED |
| `[H]` | **FAILED** — `<Document> shall not contain <Caption>` |

Only `[H]` (from the `float` package) breaks tagging. All standard LaTeX specifiers work correctly. `[h]` is the closest semantic equivalent to `[H]` ("place here") without the `float` package's incompatible reimplementation.

Upstream: [latex3/tagging-project#890](https://github.com/latex3/tagging-project/issues/890)

## Changes

- **\`src/format/pdf/format-pdf.ts\`** — expose \`pdf-tagging\` metadata to Lua filters
- **\`src/resources/filters/quarto-pre/figures.lua\`** — replace \`fig-pos='H'\` with \`'h'\` when tagging is active
- **\`src/resources/filters/quarto-post/latex.lua\`** — same for panel layouts and star environments
- **\`src/resources/filters/layout/pandoc3_figure.lua\`** — same for non-cross-referenceable figures
- **\`tests/docs/smoke-all/pdf-standard/ua-code-chunk-figure.qmd\`** — new test for R code chunk figure with ua-2
- **\`tests/docs/smoke-all/pdf-standard/ua2-unlabeled-figure-caption.qmd\`** — updated from known-failure to passing

## Test plan

- [x] New test \`ua-code-chunk-figure.qmd\` passes (R code chunk + ua-2 + verapdf)
- [x] Updated \`ua2-unlabeled-figure-caption.qmd\` now passes
- [x] Full \`pdf-standard/\` test suite: 40 passed, 0 failed
- [x] TypeScript typecheck passes

Fixes #14164

### DIFF
--- a/src/format/pdf/format-pdf.ts
+++ b/src/format/pdf/format-pdf.ts
@@ -352,6 +352,12 @@ function createPdfFormat(
             extras.metadata = extras.metadata || {};
             extras.metadata[kPdfStandardApplied] = standards;
           }
+          // Expose tagging state to Lua filters so they can adjust
+          // figure placement (e.g., avoid [H] which breaks tag structure)
+          if (needsTagging) {
+            extras.metadata = extras.metadata || {};
+            extras.metadata["pdf-tagging"] = true;
+          }
         }
 
         return extras;
--- a/src/resources/filters/layout/pandoc3_figure.lua
+++ b/src/resources/filters/layout/pandoc3_figure.lua
@@ -124,9 +124,13 @@ function render_pandoc3_figure()
       -- if this ends up in a layout without fig-pos = H, it'll fail
       -- 'H' forces it to not float
       if figure.identifier == "" then
+        -- Use [htbp] instead of [H] when PDF tagging is active.
+        -- [H] (float package) breaks lualatex's tag structure.
+        -- See https://github.com/quarto-dev/quarto-cli/issues/14164
+        local forced_pos = option("pdf-tagging", false) and "h" or "H"
         figure = _quarto.ast.walk(figure, {
           Image = function(image)
-            image.attributes['fig-pos'] = 'H'
+            image.attributes['fig-pos'] = forced_pos
             return image
           end
         })
--- a/src/resources/filters/quarto-post/latex.lua
+++ b/src/resources/filters/quarto-post/latex.lua
@@ -260,7 +260,8 @@ function render_latex()
       if f ~= nil then
         noteHasColumns()
         el.content = strip(el.content, f)
-        tprepend(el.content, {pandoc.RawBlock("latex", "\\begin{figure*}[H]")})
+        local star_pos = option("pdf-tagging", false) and "h" or "H"
+        tprepend(el.content, {pandoc.RawBlock("latex", "\\begin{figure*}[" .. star_pos .. "]")})
         tappend(el.content, {pandoc.RawBlock("latex", "\\end{figure*}")})
         return el, false
       end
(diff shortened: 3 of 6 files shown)


gordonwoodhull:
This is apparently not the way - there are hooks that can be set in order to make the `float` package work.

EDIT: Further research shows that no, there are no such hooks for `float`. But I don’t think making our floats into non-floats will be an acceptable solution.

PR #14170, opened 2026-03-06
feat(typst): add code annotation and filename support

### LINKED ISSUE #9768: [Typst] Add Code Filename support
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/9627

<sup>Originally posted by **MathisC22** May  9, 2024</sup>
### Description

File names of code cell blocks are not shown when using Typst as backend, whereas they do show when you use  LaTeX or HTML to render the file. That is, the following cell only shows the contents:
```` 
```{.python filename="matplotlib.py"}
plt.plot([1,23,2,4])
plt.show()
```
```` 
Would be really neat if it would also work with Typst!</div>

### PR DESCRIPTION
**Pull request by mcanouil**
Introduce a unified wrapper for code blocks and skylighting transformed code blocks in Typst.

Add filename processing (Lua/Typst) and code-annotation processing (Lua/Typst): `filename` renders as a tab above the block, and `# <n>` annotations render as circled numbers at the end of each line, linked both ways with the annotation list below the block.

closes #9768

## Behaviours

| Behaviour | Before (1.10) | After (this PR) |
|---|---|---|
| `filename` on a code block | Ignored, plain block. | Filename tab above the block. |
| Code annotations | Stripped, replaced by a "Line N" definition list. | Circled numbers on the annotated lines, linked both ways with the annotation list. |
| One annotation spanning several lines | "Lines 1-2" definition list. | A circle on each line, one shared link target. |
| Annotations in a `lst-cap` cross-referenceable listing | Stripped, plain ordered list. | Circles inside the listing, annotation items after the float. |
| Identifiers containing `.`/`:` (e.g. `{#lst-a.b}`) | n/a | Supported. |
| `startFrom`, with or without `number-lines` | n/a | Circle follows the code line. |
| Fenced-echo display blocks (```` ```{{r}} ````) | n/a | Circle lands on the code line, not the fence. |
| Annotation with no trailing ordered list | n/a | Unlinked circle, render still succeeds. |
| `syntax-highlighting: idiomatic` / `none` | n/a | Native path via a `raw.line` show rule. |
| `syntax-highlighting: {light: ..., dark: ...}` | n/a | Resolves to the light theme, annotations keep highlighting. |
| Languages unknown to Skylighting, bare blocks, code containing backtick runs, filenames needing escaping | n/a | Handled. |

Any remaining mismatch (for example a marker that cannot be merged) degrades to an unlinked circled number instead of a Typst compile error.

## Example

Side-by-side comparison of the same two-page document rendered with Quarto 1.10 (before) and with this PR (after). Before: code blocks are plain, no filename is shown, and annotations appear as "Line 1"-style definition lists below each block. After: code blocks show a "hello.R" filename tab, circled annotation numbers at the end of the annotated lines, and matching circled numbers in the annotation lists, including inside a captioned listing, with startFrom, in a fenced echo block, and for an annotation without a following list.

Side-by-side comparison of the idiomatic-highlighting and light/dark-theme variant documents. Before: plain code with "Line 1" definition lists. After: a filename tab and circled annotation numbers on the idiomatic variant, and preserved GitHub-theme syntax highlighting with circled annotation numbers on the light/dark variant.

````qmd
---
title: "Typst Code Annotations & Filename"
format:
  typst:
    keep-typ: true
code-annotations: true
---

## Filename

```{.r filename="hello.R"}
print("Hello, world!")
```

## Filename and annotations together

```{.r filename="hello.R"}
print("Hello, world!") # <1>
```

1. Greet the world.

## Annotations

```{.python}
def greet(name):        # <1>
    msg = f"Hi {name}"  # <2>
    return msg          # <3>
```

1. Define the function.
2. Build the message.
3. Return it.

## One annotation spanning several lines

```{.python}
x = 1  # <1>
y = 2  # <1>
z = x + y  # <2>
```

1. Set the inputs.
2. Compute the sum.

## Cross-referenceable listing

```{#lst-greet .python lst-cap="A greeting function"}
print("Hello")  # <1>
```

1. Annotated inside a captioned, referenceable listing.

See @lst-greet.

## Identifier containing a dot

```{#lst-a.b .python}
x = 1  # <1>
```

1. Dotted identifiers are valid Pandoc identifiers.

## startFrom

```{.python startFrom="5"}
x = 1  # <1>
```

1. Works whether or not line numbers are displayed.

## startFrom with displayed line numbers

```{.python .number-lines startFrom="5"}
x = 1  # <1>
```

1. The circle follows the code even when numbering starts at 5.

## Fenced echo display block

```{{r}}
x <- 1 # <1>
```

1. The circle lands on the code line, not the fence.

## Annotation without a following list

```{.python}
x = 1  # <1>
```

No ordered list follows: the circled number still renders, without a link,
instead of failing the render.
````

The same document also renders with `syntax-highlighting: idiomatic` or `none` (native Typst path) and with a light/dark theme object.

All behaviours above are covered by smoke-all tests under `tests/docs/smoke-all/typst/code-annotations/` and `tests/docs/smoke-all/typst/code-filename/`.

### CODE REVIEW

**Comment by mcanouil** [comment]
It appears the Lua limit was already at 200, so anything added at "top-level" was leading the bundle `main.lua` to exceed the locals limit.
I refactored as a module which is load inside the annotation function, this way no locals are added by require or anything.

Note that futur changes in the Lua processing chain is very likely to hit again as there are currently no room for any more locals.

- Anyhow, the PR now works with CI/CD passing, but a final check with @gordonwoodhull's PR (https://github.com/quarto-dev/quarto-cli/pull/14262) is probably needed to ensure wrapping/nesting works as expected.

Currently working with `idiomatic`

````qmd
---
format:
  typst:
    keep-typ: true
    syntax-highlighting: idiomatic
---

::: {#exm-one}

## `devtools`

A useful package for R developers.

```{.r filename="R"}
install.packages("devtools")
```

```{.r filename="R"}
install.packages("devtools") # <1>
```

1. Install the `devtools` package.

```r
install.packages("devtools")  # <1>
```

1. Install the `devtools` package.

:::
````

**Comment by mcanouil** [comment]
~~TO-DO~~: tests in layout structure such as

- [x] `layout-*`, callouts, cross-references.
- [x] list

**Comment by cderv** [comment]
Bad merging conflict resolution it seems... oups. I'll fix this

**Comment by cderv** [comment]
OK this should fix CI

@mcanouil one thing I notice, the change will add a 'stroke' line across all code blocks

Main 
image

This branch 
image


Is this expected style change for default style (when filename not used) ?  

Thanks

File: src/resources/formats/typst/pandoc/quarto/definitions.typ [hunk 1]
```diff
@@ -24,12 +24,95 @@
...
+      #h(0.4em)
+      #content
+    ]
+  }
+}
+
+// Style native raw code blocks with default inset, radius, and stroke
+#show raw.where(block: true): it => block(
+  fill: luma(230), width: 100%, inset: 8pt, radius: 2pt,
+  stroke: 0.5pt + luma(200), it,
```
mcanouil:
```suggestion
```

This is what is setting stroke for default/native codeblocks

mcanouil:
I need to check, because it's a workaround around skylighting destructing code blocks.

src/format/typst/format-typst.ts
```diff
@@ -187,40 +187,110 @@ export function typstFormat(): Format {
...
+// works with both current and future upstream bgcolor init patterns.
 function skylightingPostProcessor(brandBgColor?: string) {
   // Match the entire #let Skylighting(...) = { ... } function.
   // The signature is stable and generated by Skylighting's Typst backend.
   const skylightingFnRe =
     /(#let Skylighting\(fill: none, number: false, start: 1, sourcelines\) = \{[\s\S]*?\n\})/;
 
+  // Annotation markers emitted by the Lua filter as Typst comments
+  const annotationMarkerRe =
+    /\/\/ quarto-code-annotations: ([\w-]*) (\([^)]*\))\n(\s*(?:#block\[\s*)*(?:#quarto-code-filename\([^\n]*\)\[\s*)?)#Skylighting\(/g;
```
cderv: This could be edge case, but 🤖 had it verified 

The cell-id capture here is `[\w-]*`, but Pandoc identifiers can also contain `.` and `:` so a code block whose id has one of those (e.g. a listing id like `{#lst-a.b}`) never matches, the marker isn't merged into `#Skylighting(...)`, and the `…-annote-N-back` label is never emitted — while the annotation list below still links to it. Typst then fails to compile:

````
```{#lst-a.b .python}
x = 1  # <1>
```

1. first
````

throws something like `error: label <lst-a.b-annote-1-back> does not exist in the document`

The same document with `syntax-highlighting: idiomatic` compiles fine, so it's specific to this Skylighting-merge path. 

Probably testing this edge case helps covers it.

File: src/resources/filters/modules/typst-code-annotations.lua [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+    end
+    return '(' .. table.concat(parts, ', ') .. ')'
+  end
+
+  -- Skylighting mode: emit a Typst comment that the TS post-processor
+  -- will merge into the Skylighting call site.
+  local function typstAnnotationMarker(annotations, cellId)
+    local dict = typstAnnotationsDict(annotations)
+    return pandoc.RawBlock("typst", "// quarto-code-annotations: " .. (cellId or "") .. " " .. dict)
+  end
```
cderv: Another edge case probably:

`typstAnnotationMarker` serializes the annotation line numbers as-collected, but `wrapTypstAnnotatedCode` just below adjusts them for `startFrom` (`lineNo - startFrom + 1`). The Skylighting function renders lines starting at 1, so with `startFrom` set the marker keys point at lines Skylighting never emits → the circle and its `…-back` label are never produced, and the annotation item links to a missing label:

````markdown
```{.python startFrom="5"}
x = 1  # <1>
```

1. first
````

throws `error: label <annotated-cell-1-annote-1-back> does not exist in the document`

File: src/format/typst/format-typst.ts [hunk 4]
```diff
@@ -187,40 +187,110 @@ export function typstFormat(): Format {
...
-      return;
-    }
+    if (match) {
+      let fn = match[1];
 
-    let fn = match[1];
+      // Fix block() call: add width, inset, radius, stroke
+      fn = fn.replace(
+        "block(fill: bgcolor, blocks)",
+        "block(fill: bgcolor, width: 100%, inset: 8pt, radius: 2pt, stroke: 0.5pt + luma(200), blocks)",
```
mcanouil: and here for the same stroke

cderv (commented):
Thanks for this, nice feature. We don't have a lot in v1.10, I was aiming for a improvement and bug fix only. So maybe this should go to v1.11 early as first feature and first pre-release 🤔 

Two edge cases found with 🤖 review help (both Skylighting-path only, native path is fine) that end in a hard Typst compile error rather than a graceful fallback. we should probably see if we can cover for them.

mcanouil:
The stroke was added to ensure the code block could be distinguished from prose regardless of the theme used. For example, `github-light`has a white or near white background which means you no longer see the block and only the code.

One could argue that's fine, especially since that's the behaviour in HTML for example.
edit: I'll push a commit to make it consistent with HTML.
screenshot showing code blocks formatting in HTML where code block background is white with no stroke

mcanouil:
> Thanks for this, nice feature. We don't have a lot in v1.10, I was aiming for a improvement and bug fix only. So maybe this should go to v1.11 early as first feature and first pre-release 🤔
> 
> Two edge cases found with 🤖 review help (both Skylighting-path only, native path is fine) that end in a hard Typst compile error rather than a graceful fallback. we should probably see if we can cover for them.

I can't check the edge cases now, so yes let's put this for 1.11.
I'll use Claude Code to target and find other cases like this to make the feature more robust.

cderv:
> The stroke was added to ensure the code block could be distinguished from prose regardless of the theme used. For example, github-lighthas a white or near white background which means you no longer see the block and only the code.

To be clear, this could be a good idea to add this stroke. Maybe it helps with contrast and accessibility too. 

However, if we do that, we probably want to do it in all formats - unless this is a problem with Typst only. I did not check.

mcanouil:
> However, if we do that, we probably want to do it in all formats - unless this is a problem with Typst only. I did not check.

The "missing stroke" is global, so if we want to address this, it should be in a follow-up PR and keep this PR's scope narrow to bringing the missing feature "as is".

The screenshot in https://github.com/quarto-dev/quarto-cli/pull/14170#issuecomment-5056897874 is from HTML format.

PR #14173, opened 2026-03-07
Add typst-gather 0.2.2 binary dependency

gordonwoodhull:
## Summary

- Adds [typst-gather](https://github.com/quarto-dev/typst-gather) v0.2.2 as a downloaded binary dependency, following the same pattern as Pandoc, Typst, esbuild, and other binary deps
- Pre-built binaries are fetched from GitHub releases for all 5 supported platform/arch combos (x86_64 + aarch64, Linux/macOS/Windows)
- This is a dependency-only prerequisite for #13983 (selective typst package staging)

## Files changed

- `configuration` — add `TYPST_GATHER=0.2.2`
- `package/src/common/dependencies/typst-gather.ts` — new dependency definition
- `package/src/common/dependencies/dependencies.ts` — register in dependency list

## Test plan

- [ ] Verify `./configure.sh` downloads the binary to `package/dist/bin/tools/<arch>/typst-gather`
- [ ] Verify `typst-gather --version` reports `0.2.2`
- [ ] Verify CI passes on all platforms

### DIFF
--- a/configuration
+++ b/configuration
@@ -17,6 +17,7 @@ export PANDOC=3.8.3
 export DARTSASS=1.87.0
 export ESBUILD=0.25.10
 export TYPST=0.14.2
+export TYPST_GATHER=0.2.2
 export VERAPDF=1.28.2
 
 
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Hi @cscheid - I'm moving `typst-gather` to its own repo. 

Here is the dependency change only, using the new `archiveOnly` flag, to be removed in the follow-up
- #14174

To be safe, let's merge this as soon as we start 1.10. Should be safe to backport if we want to.

cderv:
Thanks ! This will make worktrees configuration faster ! I was looking into this !

PR #14181, opened 2026-03-09
Support ARM64 Linux and new TinyTeX naming scheme

cderv:
`quarto install tinytex` fails on ARM64 Linux because `needsSourceInstall()` blocks non-x86_64 platforms and `tinyTexPkgName()` only generates old-style filenames. TinyTeX has since added ARM64 Linux prebuilt binaries and adopted a new platform-aware naming scheme for all platforms:

- rstudio/tinytex#483 — ARM64 Linux prebuilt binaries
- rstudio/tinytex#484, rstudio/tinytex#486 — New `TinyTeX-{os}[-{arch}]-{ver}.{ext}` naming with `.tar.xz` compression
- rstudio/tinytex-releases#37 — Legacy filenames will eventually be removed

The TinyTeX install shell scripts were updated (rstudio/tinytex#486) to handle this transition with version-gated logic: if version > v2026.03.02, use new names; otherwise use old names. However, `quarto install tinytex` always fetches the latest monthly release via the GitHub API (`/releases/latest`) — there is no pinned version to compare against. So instead, `tinyTexPkgName()` now generates candidate filenames for both new and old naming schemes (new preferred, old as fallback) and picks the first match from the release assets. This handles the transition seamlessly regardless of which naming convention a given release uses.

## Changes

- Remove the `needsSourceInstall()` prereq gate that blocked ARM64 Linux
- Update `tinyTexPkgName()` to generate candidate filenames for both naming schemes, with platform and architecture awareness (new naming: `TinyTeX-linux-arm64-v2026.04.tar.xz`, old: `TinyTeX-arm64-v2026.03.02.tar.gz`)
- Update `tinyTexUrl()` to iterate candidates and match first available asset
- Change `tar xfz` to `tar xf` in `unzip()` for automatic compression detection (`.tar.gz`, `.tgz`, `.tar.xz`)
- Add `.exe` self-extracting archive support in `unzip()` for the new Windows naming
- Add unit tests for package name generation covering all platform/arch combinations
- Add asset-existence tests that verify candidate names match actual TinyTeX release assets

Fixes #12124

### DIFF
--- a/news/changelog-1.9.md
+++ b/news/changelog-1.9.md
@@ -181,6 +181,7 @@ All changes included in 1.9:
 
 - ([#4426](https://github.com/quarto-dev/quarto-cli/issues/4426)): New `quarto install verapdf` command installs [veraPDF](https://verapdf.org/) for PDF/A and PDF/UA validation. When verapdf is available, PDFs created with the `pdf-standard` option are automatically validated for compliance. Also supports `quarto uninstall verapdf`, `quarto update verapdf`, and `quarto tools`.
 - ([#11877](https://github.com/quarto-dev/quarto-cli/issues/11877), [#10961](https://github.com/quarto-dev/quarto-cli/issues/10961), [#6821](https://github.com/quarto-dev/quarto-cli/issues/6821), [#13704](https://github.com/quarto-dev/quarto-cli/issues/13704)): New `quarto install chrome-headless-shell` command downloads [Chrome Headless Shell](https://developer.chrome.com/blog/chrome-headless-shell) from Google's Chrome for Testing API. This is the recommended headless browser for diagram rendering (Mermaid, Graphviz) to non-HTML formats. Smaller and lighter than full Chrome, with fewer system dependencies.
+- ([#12124](https://github.com/quarto-dev/quarto-cli/issues/12124)): Support `quarto install tinytex` on ARM64 Linux and adopt new TinyTeX naming scheme with `.tar.xz` compression.
 
 ### `preview`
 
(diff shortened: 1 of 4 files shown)


cderv:
I would like to merge this for 1.9 - ok with this @cscheid ? 

This PR adds support for both the old and new schemes. CI passes to it means TinyTeX can still be installed. And when next monthly release we install will have the new names, we'll support it without patch.

yihui (commented):
I wonder what are the differences between the installation method in `src/tools/impl/tinytex.ts` and the scripts `install-bin-windows.bat`/`install-bin-unix.sh` in https://github.com/rstudio/tinytex/tree/main/tools. If the main purpose of the typescript re-implementation was to allow a custom installation directory, this is already supported in the installation scripts via the `TINYTEX_DIR` env var.

Ideally I think we should just download the installation scripts and run them (with desired env vars set), otherwise we are making strong assumptions about the installers that are too specific (e.g., the URLs, compression formats, etc). They don't change often (no changes in the past 6 years), but there's no guarantee. That said, backward compatibility will definitely be provided, and older filenames won't disappear overnight.

BTW, TinyTeX has also started supporting binaries for musl/Alpine Linux (currently daily build only). I'm not sure if it's worth adding the support here.

cderv:
Thanks Yihui ! 

I'll consider for next version what it would mean to use the script for installation when running `quarto install tinytex`. 

Maybe this was done this way so no external scripts are downloaded just for doing the install.
We could bundle them too, but this would not solve the update problem is it happens again. 

Anyhow, with this PR we should be covered for next month change.

PR #14190, opened 2026-03-10
Fix intermittent smctl not found in signing workflow

cderv:
The `make-installer-win` job in `create-release.yml` intermittently fails at "Sync certificates" with `smctl` not recognized, despite the previous step reporting successful installation.

## Root Cause

The `Install SMCTL` step has no error handling. When the 90 MB MSI download from S3 silently fails (0 bytes received), `msiexec` runs on the empty file and exits with a non-zero code that PowerShell doesn't propagate. The script unconditionally prints "SMCTL installed and added on PATH" and the next step fails when `smctl.exe` doesn't exist on disk.

Evidence from failed runs ([#22838611306](https://github.com/quarto-dev/quarto-cli/actions/runs/22838611306), [#22635880163](https://github.com/quarto-dev/quarto-cli/actions/runs/22635880163)): curl shows 0 bytes downloaded while the successful run ([#22887281659](https://github.com/quarto-dev/quarto-cli/actions/runs/22887281659)) shows a normal 90.7 MB download.

## Fix

- Download retry loop (3 attempts) with file size validation
- Synchronous msiexec via `Start-Process -Wait` with exit code check
- Post-install verification that `smctl.exe` exists on disk
- Fail fast with clear error messages on any failure

## Test plan

- Trigger `create-release.yml` via `workflow_dispatch` with `publish-release=false` on this branch to exercise the signing path without publishing

### DIFF
--- a/.github/workflows/actions/sign-files/action.yml
+++ b/.github/workflows/actions/sign-files/action.yml
@@ -44,9 +44,43 @@ runs:
       run: |
         Write-Output "::group::Install smctl if needed"
         if (!(Get-Command smctl -ErrorAction SilentlyContinue)) {
-          curl -o smtools-windows-x64.msi "https://rstudio-buildtools.s3.amazonaws.com/posit-dev/smtools-windows-x64.msi"
-          msiexec /i smtools-windows-x64.msi /quiet /qn /log smtools-windows-x64.log
-          "C:/Program Files/DigiCert/DigiCert One Signing Manager Tools" | Out-File -FilePath $env:GITHUB_PATH -Append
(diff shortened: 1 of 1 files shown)


cscheid:
(christophe + claude is a scary combo!) amazing, thanks for the fix.

cderv:
I am beginning to have a nice workflow, and get what I want from claude to get targetted fix. 

By the way testing workflow passed: https://github.com/quarto-dev/quarto-cli/actions/runs/22905990760/job/66464755541

We now get 3 attempts in case of transient error with the s3 download, and then clear failure at installation instead of later at signing.

PR #14192, opened 2026-03-10
websites - add pagefind.app support

cscheid:
[pagefind](https://pagefind.app) is an alternative content search library to fusejs, and offers a bit more customizability. It should also be significantly faster and more resource-efficient than fusejs.

To be done:

- [ ] changelog entry (we don't have a 1.10 file)
- [ ] documentation


cscheid:
This would work if we didn't try and bundle the JS, but the bundler has a failure we can't easily work around, sadly.

PR #14210, opened 2026-03-13
Fix preview of single-file Julia documents

gordonwoodhull:
Engine extensions (like the bundled Julia engine) were only loaded in singleFileProjectContext when renderOptions was provided. The preview command creates the project context before render services exist, so engine extensions were never discovered for single files. This caused Jupyter to win the language claim for Julia code blocks.

Now engine extensions are always resolved, using a fresh extension context as fallback when renderOptions is absent — matching the pattern used by projectContext.

Fixes #14208

### DIFF
--- a/src/project/types/single-file/single-file.ts
+++ b/src/project/types/single-file/single-file.ts
@@ -35,6 +35,7 @@ import {
   mergeExtensionMetadata,
   resolveEngineExtensions,
 } from "../../project-context.ts";
+import { createExtensionContext } from "../../../extension/extension.ts";
 
 export async function singleFileProjectContext(
   source: string,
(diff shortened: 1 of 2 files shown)


cderv:
The fix looks correct to me — I arrived at the same approach independently.

For testing, since preview is hard to test end-to-end, a unit test on `singleFileProjectContext` works well. The idea is to call it **without** `renderOptions` (the preview code path) and assert that engine extensions are still resolved:

```typescript
// tests/unit/project/single-file-engine-resolution.test.ts
import { unitTest } from "../../test.ts";
import { assert } from "testing/asserts";
import { join } from "../../../src/deno_ral/path.ts";
import { singleFileProjectContext } from "../../../src/project/types/single-file/single-file.ts";
import { notebookContext } from "../../../src/render/notebook/notebook-context.ts";
import { initYamlIntelligenceResourcesFromFilesystem } from "../../../src/core/schema/utils.ts";

unitTest(
  "singleFileProjectContext resolves engine extensions without renderOptions",
  async () => {
    await initYamlIntelligenceResourcesFromFilesystem();

    const tmpDir = Deno.makeTempDirSync({ prefix: "quarto-test" });
    const file = join(tmpDir, "test.qmd");
    Deno.writeTextFileSync(
      file,
      "---\ntitle: test\nengine: julia\n---\n",
    );

    try {
      const nbContext = notebookContext();
      const project = await singleFileProjectContext(file, nbContext);

      assert(
        project.config !== undefined,
        "config should be initialized even without renderOptions",
      );
      assert(
        Array.isArray(project.config?.engines) &&
          project.config.engines.length > 0,
        "engine extensions should be resolved (bundled engines discovered)",
      );
    } finally {
      Deno.removeSync(tmpDir, { recursive: true });
    }
  },
);
```

This fails before the fix (`config` is `undefined`) and passes after. The key insight is that `initYamlIntelligenceResourcesFromFilesystem()` is needed because `createExtensionContext()` reads extension schemas.

gordonwoodhull:
Thanks @cderv, added.

PR #14216, opened 2026-03-16
Update font in cosmo.scss

lynn:
## Description

Source Sans Pro is now known as Source Sans 3. But the real reason I make this change is to include the italic: as it stands, a default-settings Quarto document has rather ugly-looking automatically-generated italic letters (slanted versions of the upright).

The bottom line in this screenshot shows the new, proper italic glyphs for _a_ and _g_:

image

## Checklist

I have (if applicable):

- [ ] added new tests


cscheid:
Thanks!

We want to be a bit careful here because we use SCSS in unexpected places (in translations to Typst specifically), and we need to determine that these changes are safe.

PR #14301, opened 2026-04-02
[backport] Fix preview browse URL for single-file documents

### LINKED ISSUE #14298: Quarto 1.9 preview server appends output filename to Browse URL, breaking proxied server access
## Bug description

Identified thru failing Workbench test `test_running_quarto_for_python_app_in_vscode` on Workbench version `2026.04.0-daily-460.pro4`.

Starting in Quarto 1.9, `quarto preview` appends the output filename (e.g., `/hello.html`) to the "Browse at" URL. This breaks proxied server access in VS Code and Positron sessions on Posit Workbench, because the Proxied Servers pane uses the base URL (without the filename) to link users to the app.

In Quarto 1.8, clicking the proxied server link takes the user to the correct page. In 1.9, the same link returns a 404 because the app is now served at the path with the filename appended.

https://github.com/user-attachments/assets/aa461b6f-b89f-41b4-956e-56033fa0d17f

## Reproduction

1. Start a VS Code session in Workbench `2026.04.0-daily-460.pro4` and run this in the session terminal to setup the local env:
```
mkdir test; cd test
python3 -m venv .venv
source .venv/bin/activate
python3 -m pip install matplotlib plotly_express numpy jupyter pyyaml
export PATH="/usr/lib/rstudio-server/bin/quarto/bin:$PATH"
```

2. Create `test/hello.qmd`:
```
---
title: "Quarto Basics"
format:
  html:
    code-fold: true
jupyter: python3
---

For a demonstration of a line plot on a polar axis, see @fig-polar.

```{python}
#| label: fig-polar
#| fig-cap: "A line plot on a polar axis"

import numpy as np

r = np.arange(0, 2, 0.01)
theta = 2 * np.pi * r
fig, ax = plt.subplots(
  subplot_kw = {'projection': 'polar'}
)
ax.plot(theta, r)
ax.grid(True)
plt.show()
\```# remove the \
```

3. run `quarto preview hello.qmd` on the terminal

4. Try to view the app by clicking the link in the Proxied Servers pane

## Impact

This affects VS Code and Positron sessions on Posit Workbench. Users can no longer view a Quarto app by clicking the link in the Proxied Servers pane. Fixing this on the Workbench side would be a significant lift.

## Workaround

Downgrade to Quarto 1.8.

cderv:
> [!IMPORTANT]
> Backport of #14300

`quarto preview hello.qmd` prints `Browse at http://localhost:PORT/hello.html` instead of `http://localhost:PORT/`, and `GET /` returns 404. This breaks Posit Workbench Proxied Servers pane.

PR #13804 made `project` always non-null via `singleFileProjectContext()`. Two code paths branch on `project` being truthy: the URL path computation (prints filename) and the handler selection (uses wrong default file). Fix guards both with `project.isSingleFile`.

Verified locally: v1.8.27 serves root URL correctly, v1.9.36 returns 404, fix restores v1.8 behavior. Also ran T17/T18 manual preview tests directly on this branch — both pass (`GET /` and `GET /plain.html` return 200). Added missing `plain.qmd` test fixture that was not part of the original backport.

Fixes #14298

### DIFF
--- a/.claude/commands/quarto-preview-test/SKILL.md
+++ b/.claude/commands/quarto-preview-test/SKILL.md
@@ -1,0 +1,145 @@
+---
+name: quarto-preview-test
+description: Use when testing preview functionality, verifying live reload, or validating preview fixes. Covers starting preview with port/logging, browser verification via /agent-browser, and checking logs/filesystem for artifacts.
+---
+
+# Quarto Preview Test
+
+Interactive testing of `quarto preview` with automated browser verification.
+
+## Tools
+
+| Tool | When to use |
+|------|-------------|
+| `/agent-browser` | **Preferred.** Token-efficient browser automation. Navigate, verify content, screenshot. |
+| Chrome DevTools MCP | Deep debugging: console messages, network requests, DOM inspection. |
+| `jq` / `grep` | Parse debug log output. |
+
+## Prerequisites
+
+- Quarto dev version built (`./configure.sh` or `./configure.cmd`)
+- Test environment configured (`tests/configure-test-env.sh` or `.ps1`)
+- `/agent-browser` CLI installed (preferred), OR Chrome + Chrome DevTools MCP connected
+
+## Starting Preview
+
+Preview needs the test venv for Jupyter tests. Activate it first (`tests/.venv`), matching how `run-tests.sh` / `run-tests.ps1` do it.
+
+```bash
+# Linux/macOS
+source tests/.venv/bin/activate
+./package/dist/bin/quarto preview <file-or-dir> --no-browser --port 4444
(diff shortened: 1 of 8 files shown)


cscheid:
Thanks! (We would really benefit from a broader effort to have the hosted teams bring us tests against their expected behavior..)

cderv:
Yes for sure. I have now added a skill to have help from claude testing preview. 

So manual regression tests we can run when we are touching preview code base. This will help. 

This expected behavior is now registered. 

It seems they have automatic tests though as this is how they discovered. We should probably see if we can run those tests again all our pre-release !

PR #14307, opened 2026-04-03
Fix quarto inspect standalone file project emission for RStudio

cderv:
When publishing a standalone Quarto document from RStudio's publishing wizard with Quarto >= 1.9.35, the wizard crashes with "missing value where TRUE/FALSE needed".

## Root Cause

Commit f6a0f8988e (v1.9.35, fix for #14208) moved `result.config = result.config || { project: {} }` outside the `if (renderOptions)` block in `singleFileProjectContext` so Julia engine extensions resolve correctly during preview. Side effect: `context.config` is now always truthy for standalone files, so `inspectDocumentConfig` emits a `project` field with `dir` set to the file's parent directory.

RStudio's C++ bridge (`SessionQuarto.cpp`) reads `project.dir` from `quarto inspect` output and assumes `_quarto.yml` exists there. It adds the config file to the deploy list, `file.info()` returns NA for the non-existent file, and the size comparison crashes.

## Fix

Gate `project` emission in `inspectDocumentConfig` on `!(isSingleFile && isRStudio())`. This preserves the "everything is a project" direction for non-RStudio consumers while maintaining compatibility with current RStudio releases.

## Alternative Considered

Unconditionally suppressing `project` for all single-file contexts (`!context.isSingleFile`) would also fix the issue without RStudio detection. However, single-file projects are intentional internal behavior and other consumers may rely on the project metadata. The RStudio-specific gate is more targeted.

The RStudio-side fix is in rstudio/rstudio#17336 (merged) but will not be backported to older releases. This Quarto-side workaround protects users on older RStudio versions that upgrade Quarto independently.

Also adds `_setIsRStudioForTest()` override in `platform.ts` to avoid `Deno.env.set()` race conditions in parallel tests (#14218).

## Test Plan

- [x] `quarto inspect standalone.qmd` with RStudio override: no `project` field in output
- [x] `quarto inspect standalone.qmd` without override: `project` field present with `dir`
- [x] Existing inspect tests pass (extensions, includes, code-cells, cleanup)

Ref: rstudio/rstudio#17333

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -5,6 +5,7 @@ All changes included in 1.10:
 - ([#14267](https://github.com/quarto-dev/quarto-cli/issues/14267)): Fix Windows paths with accented characters (e.g., `C:\Users\Sébastien\`) breaking dart-sass compilation.
 - ([#14281](https://github.com/quarto-dev/quarto-cli/issues/14281)): Fix transient `.quarto_ipynb` files accumulating during `quarto preview` with Jupyter engine.
 - ([#14298](https://github.com/quarto-dev/quarto-cli/issues/14298)): Fix `quarto preview` browse URL including output filename (e.g., `hello.html`) for single-file documents, breaking Posit Workbench proxied server access.
+- ([rstudio/rstudio#17333](https://github.com/rstudio/rstudio/issues/17333)): Fix `quarto inspect` on standalone files emitting project metadata that breaks RStudio's publishing wizard.
 
 ## Formats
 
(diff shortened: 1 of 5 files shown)


cderv:
@cscheid is that what you add in mind as a fix ? 

I wanted to try gate this for RStudio Version, but I don't think it can be accessed easily unfortunately. So I just have a scoped build to hide project from inspect in rstudio. 

But maybe you want to remove singleFile project info from inspect in all cases ? 

I hesitated doing this, so happy to get your thought. 

Also I have made the backport so current release 1.9 works in existing and old RStudio IDE

cscheid (approved):
That looks very much like what we need to do here, thanks!

PR #14343, opened 2026-04-09
catch exception when directory contents change mid-traversal (#14342)

### LINKED ISSUE #14342: expandGlobSync has a TOCTOU race
There's an unavoidable race in Deno's expandGlobSync; some other process might change the directory between resolving the globs and `stat`ing them. We need to defend against this. Most directly, this manifests in `resolvePathGlobs()`, `src/core/path.ts:165-187` (which is called by projectResourceFiles, which sometimes causes crashes on `quarto preview` inside Positron).

cscheid:
Closes #14342.

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -40,4 +40,4 @@ All changes included in 1.10:
 
 - ([#6651](https://github.com/quarto-dev/quarto-cli/issues/6651)): Fix dart-sass compilation failing in enterprise environments where `.bat` files are blocked by group policy.
 - ([#14255](https://github.com/quarto-dev/quarto-cli/issues/14255)): Fix shortcodes inside inline and display math expressions not being resolved.
-
+- ([#14342](https://github.com/quarto-dev/quarto-cli/issues/14342)): Work around TOCTOU race in Deno's `expandGlobSync` that can cause unexpected exceptions to be raised while traversing directories during project initialization.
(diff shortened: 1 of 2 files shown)


cscheid:
(@cderv, added you here just FYI. We can't easily create a regression test here, because it's a race condition.)

I want to merge this and build a new 1.10 release to see if this addresses the problem @mine-cetinkaya-rundel is experiencing.

cscheid:
This is really a workaround more than anything else. The try-catch should happen inside the main `expandGlobSync` loop, tightly around `statSync`. This is so that the glob walker doesn't abort. That would be getting rid of the Deno stdlib call, though (which might be a good idea nevertheless)

cderv:
I see the problem. Indeed not easy to reproduce, even manually without any IDE. 

FWIW, current "regression" test on preview I did are manual test with a description in README on what we want - see https://github.com/quarto-dev/quarto-cli/tree/main/tests/docs/manual/preview

So that we can use claude to help us verify a behaviour using https://github.com/quarto-dev/quarto-cli/blob/main/.claude/commands/quarto-preview-test/SKILL.md

Not that automated, but better than nothing. 

However, this race condition may even be hard to put in this context, i.e. creating a manual test scenario to check. 

> This is really a workaround more than anything else

Should we try consider a better solution for later if this solves the problem ?

cscheid:
> > This is really a workaround more than anything else

> Should we try consider a better solution for later if this solves the problem ?

Yes. I have a plan. We should rewrite that expandGlobSync function ourselves (it's not that complicated). But I don't want to do that before ensuring that this is a step in the right direction.

PR #14354, opened 2026-04-13
Fix revealjs title author ORCID spacing

jnkatz:
## Summary

Fixes revealjs fancy title slides so authors without ORCID IDs do not render a trailing space after the author name.

The title slide partial had a literal separator space before the optional ORCID block. When `by-author.orcid` was unset, that separator remained as trailing whitespace in the generated author-name block. This moves the separator inside the ORCID conditional so it is only emitted when the ORCID link is emitted.

## Validation

- Added a revealjs fixture without an ORCID ID.
- Added a smoke-test assertion that the generated author name is followed directly by `</div>` and does not contain trailing horizontal whitespace before that closing tag.
- Ran `git diff --check`.
- Ran `./configure.sh` successfully after allowing Deno network access outside the local sandbox.
- Ran `DENO_DIR=aarch64 QUARTO_TESTS_NO_CONFIG=true QUARTO_TESTS_FORCE_NO_VENV=true ./run-tests.sh smoke/render/render-reveal.test.ts`.

The new `title-fancy-author.qmd` regression test passed. Three existing tests in the same file failed in this local checkout because the fresh R test environment did not have `knitr`/`rmarkdown` installed; those failures were unrelated to this title-slide change.

- Also rendered the new fixture directly with the dev Quarto build and verified `Jonathan N. Katz` is followed by a newline and `</div>`, with no trailing horizontal whitespace.


cderv:
Thank you for the report ! 

This is indeed a space not correctly placed. I'll handle it from here - we prefer to have smoke-all test now , and we need changelog.

PR #14366, opened 2026-04-15
Fix .quarto_ipynb files not cleaned up after render

### LINKED ISSUE #14359: .quarto_ipynb files not cleaned up when preview is killed ungracefully
Reported by @mine-cetinkaya-rundel.

When using Positron and clicking the trash/bin icon on the terminal pane running `quarto preview`, the process is killed without clean shutdown. The transient `.quarto_ipynb` file created during render is not cleaned up. On the next preview, a new `.quarto_ipynb_1` is created, and this accumulates (`_2`, `_3`, ...) each time the user repeats the pattern.

This is distinct from #14281, which fixed accumulation *within* a single preview session (re-renders). This issue is about accumulation *across* sessions when the process exits ungracefully.

## Steps to reproduce

Create `test.qmd` with Jupyter engine content:

````qmd
---
format: html
---

```{python}
1 + 1
```
````

1. Open in Positron
2. Click the Preview button (opens terminal running `quarto preview test.qmd`)
3. Wait for render to complete
4. Click the bin/trash icon on the terminal pane (kills terminal)
5. Click Preview again
6. Repeat steps 4-5 a few times

## Actual behavior

Files accumulate:
```
test.quarto_ipynb
test.quarto_ipynb_1
test.quarto_ipynb_2
```

## Expected behavior

No `.quarto_ipynb` files left behind between preview sessions.

## Analysis

In `quarto render`, the `.quarto_ipynb` is cleaned up promptly after render completes — `cmd.ts` calls `context.cleanup()` which triggers `cleanupFileInformationCache()` → `invalidateForFile()` → `safeRemoveSync()`.

In `quarto preview`, cleanup is deferred: the file sits on disk between renders and is only deleted at process exit (via `onCleanup`) or before the next re-render (via `invalidateForFile`). If the process is killed ungracefully (SIGKILL, `taskkill`, terminal destruction), the exit cleanup never fires.

Historically, `cleanupNotebook()` in `jupyter.ts` used to delete the file immediately after execution when `keep-ipynb` was not set. This was changed in #12793 to integrate with the `fileInformationCache` system (fixing #12780), but the immediate deletion for the `keep-ipynb: false` case was dropped as a side effect. The file deletion was fully delegated to the cache cleanup, which only runs at context disposal time.

## Suggested fix

Restore immediate deletion in `cleanupNotebook()` for the `keep-ipynb: false` case, while preserving the cache integration for `keep-ipynb: true`:

- When `transient && keep-ipynb`: mark cache entry as non-transient (existing behavior from #12793)
- When `transient && !keep-ipynb`: delete the file from disk immediately after execute

This restores the original behavior from before #12793 and aligns preview with how `quarto render` handles cleanup. The `execute()` function already has a guard that recreates the file if missing (with a comment: "could have been removed by the cleanup step of another render"), so re-renders continue to work.

Related: #14281, #12780, #12793, #11597

cderv:
When rendering a `.qmd` file with Jupyter execution, the intermediate `.quarto_ipynb` file is never deleted. Each subsequent preview creates numbered variants (`_1`, `_2`, etc.) that accumulate on disk.

## Root Cause

PR #12793 added `keep-ipynb: true` support by integrating `cleanupNotebook()` with `fileInformationCache`. As a side effect, it dropped the immediate file deletion for the default `keep-ipynb: false` case. The function only flips a `transient` flag in the cache when `keep-ipynb: true` is set, but does nothing otherwise — the file just stays on disk.

## Fix

Add an `else if (data.transient)` branch in `cleanupNotebook()` that calls `safeRemoveSync(target.input)` when `keep-ipynb` is false, restoring the original deletion behavior while preserving the cache integration for `keep-ipynb: true`.

This is safe because `execute()` already has a guard (lines 441-447) that recreates the notebook if it's missing, and `safeRemoveSync` tolerates missing files.

Also adds `fileNotExists` verification to the smoke-all test framework and a test that verifies the `.quarto_ipynb` file is cleaned up after render.

Fixes #14359

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -52,3 +52,4 @@ All changes included in 1.10:
 - ([#6651](https://github.com/quarto-dev/quarto-cli/issues/6651)): Fix dart-sass compilation failing in enterprise environments where `.bat` files are blocked by group policy.
 - ([#14255](https://github.com/quarto-dev/quarto-cli/issues/14255)): Fix shortcodes inside inline and display math expressions not being resolved.
 - ([#14342](https://github.com/quarto-dev/quarto-cli/issues/14342)): Work around TOCTOU race in Deno's `expandGlobSync` that can cause unexpected exceptions to be raised while traversing directories during project initialization.
+- ([#14359](https://github.com/quarto-dev/quarto-cli/issues/14359)): Fix intermediate `.quarto_ipynb` file not being deleted after rendering a `.qmd` with Jupyter engine, causing numbered variants (`_1`, `_2`, ...) to accumulate on disk across renders.
--- a/src/command/render/output-typst.ts
+++ b/src/command/render/output-typst.ts
@@ -62,7 +62,7 @@ export interface NeededPackage {
 
 // Collect all package source directories (built-in + extensions)
 async function collectPackageSources(
-  input: string,
+  inputDir: string,
   projectDir: string,
 ): Promise<string[]> {
   const sources: string[] = [];
@@ -74,7 +74,7 @@ async function collectPackageSources(
   }
 
   // 2. Extension packages
-  const extensionDirs = inputExtensionDirs(input, projectDir);
+  const extensionDirs = inputExtensionDirs(inputDir, projectDir);
   const subtreePath = builtinSubtreeExtensions();
   for (const extDir of extensionDirs) {
     const extensions = extDir === subtreePath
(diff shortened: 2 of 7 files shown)


cderv:
CI failures were caused by a latent bug in `stageTypstPackages()` / `collectPackageSources()` in `output-typst.ts`. These functions received the execution target's input path (the transient `.quarto_ipynb` for Jupyter targets) but only need the input **directory** for extension discovery. When the `.quarto_ipynb` cleanup restored by this PR deleted the file before the typst `complete()` phase, `inputDirName()` crashed on `Deno.statSync()` of the missing file.

Fixed in 0cadc5e — pass `inputDir` (already computed by `dirAndStem`) instead of the file path. The typst/great-tables tests that were failing now pass locally.

PR #14374, opened 2026-04-17
Handle missing status field in execute_reply from XEUS-based kernels

ChrisJefferson:
XEUS-based Jupyter kernels (such as the Maple 2025 kernel) omit the 'status' field from execute_reply messages when execution produces an error.  This violates the Jupyter messaging protocol, which requires every execute_reply to include status.  The missing field causes a KeyError that crashes notebook execution.

This is a workaround for a kernel-side protocol violation — the right fix is in Maple 2025, but I think it's useful to fix (I assumed, incorrectly, it was quarto's fault at first, and the error message is very confusing).  But the workaround is low-risk: we catch a specific KeyError("'status'") around client.execute_cell() and record it as a cell error output rather than crashing.  Kernels that conform to the protocol are unaffected.

Tested against Maple 2025 (XEUS 2.3.1) with 18 Quarto documents containing executable Maple code blocks. I think this might effect other XEUS based kernels, not just Maple (although I haven't tested them).

- [x ] filed a [contributor agreement](https://github.com/quarto-dev/quarto-cli/blob/main/CONTRIBUTING.md).
- [x ] ensured the present test suite passes
- [ ] added new tests (this is hard to do as it requires Maple)


cderv:
Thanks for report and fix. 

> I think this might effect other XEUS based kernels, not just Maple (although I haven't tested them)

Do you have a list of other Xeus based kernel we could look at testing ? 

I would like to maybe add a test for this in our codebase, and we would need an open working kernel.

cderv:
Looked into XEUS itself to answer the question — this is kernel-specific, not XEUS-framework-wide.

## Why xeus-python and xeus-r are not affected

XEUS has helpers that always set `status` on the reply:

https://github.com/jupyter-xeus/xeus/blob/69d6d1397c68ba0ac1f6ab766dbeebb8a81e5b03/src/xhelper.cpp#L65-L75

Both xeus-python and xeus-r go through these helpers:

https://github.com/jupyter-xeus/xeus-python/blob/ab188126dad0b559315228f06f4ad3135dd31b0e/src/xinterpreter.cpp#L150-L161

https://github.com/jupyter-xeus/xeus-r/blob/d10c5fcc548b3d78ed0eecf07bc8ebb732573dcb/src/xinterpreter.cpp#L136-L146

So a conforming XEUS kernel produces `execute_reply` with `status` present — the `except KeyError` branch doesn't fire for xeus-python, xeus-r, or xeus-cling. The omission comes from kernels that bypass the helpers or predate them. Maple 2025 is the only one I know of currently.

## Testing

A smoke-all test with xeus-python or xeus-r won't exercise the guard. The cleanest path would be a mocked unit test against `cell_execute` in `notebook.py`, but we don't have Python unit test infrastructure in the repo yet — that's follow-up rather than blocking.

I'll merge once CI is green.

ChrisJefferson:
Thanks, I checked the source of xeus and saw it looked possible to miss this, I didn't notice the helper -- I should have gone and actually tried some other kernels.

PR #14464, opened 2026-04-30
fix(pdf/latex): wrap Note content in Blocks list for reference-location margin (#13588)

mcanouil:
`n.content` on a `Note` element must be a `Blocks` list, but the handler introduced for #7534 was assigning a bare `pandoc.Para(...)` (a single `Block`), causing `scopedwalk.lua` to crash with `attempt to get length of a Block value` whenever a footnote appeared alongside a figure with `fig-cap` in a PDF document using `reference-location: margin`.

Wrapping the `Para` in `pandoc.Blocks({...})` restores the correct type and preserves the `\endgraf` behaviour needed for multi-paragraph sidenotes.

Closes #13588

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -9,6 +9,10 @@ All changes included in 1.10:
 
 ## Formats
 
+### `pdf`
+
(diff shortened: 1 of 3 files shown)


cderv (approved):
Thank you!

Issue #13677, opened 2025-11-12
Path inconsistency in listings


cwickham:
Example repo: http://github.com/cwickham/listings

cc @cscheid 

## Relative links for listing `contents`

I want to be able to have a listing on a post, that gives other “recent
posts”. E.g. on `posts/post-with-code/index.qmd`.

This doesn’t work:

<div class="code-with-filename">

**posts/post-with-code/index.qmd**

``` yaml
listing:
  contents: /posts
```

</div>

    WARN: The listing in 'posts/post-with-code/index.qmd' using the following contents:
    - /posts
    doesn't match any files or folders.

Neither does this:

<div class="code-with-filename">

**posts/post-with-code/index.qmd**

``` yaml
listing:
  contents: ../
```

</div>

    WARN: The listing in 'posts/post-with-code/index.qmd' using the following contents:
    - ../
    doesn't match any files or folders.

This works, but why?:

<div class="code-with-filename">

**posts/post-with-code/index.qmd**

``` yaml
listing:
  contents: ../*/index.qmd
```

</div>

## Project relative links for listing `template`

It might be nice to specify a global listing template in
`_metadata.yml`, e.g.

<div class="code-with-filename">

**\_metadata.yml**

``` yaml
listing:
  template: blog.ejs
```

</div>

It needs to specified to work regardless of where the listing is used.
E.g. `blog.ejs` for `index.qmd`, `../../blog.ejs` for
`posts/post-with-code/index.qmd`. Using a project path doesn’t work:

<div class="code-with-filename">

**\_metadata.yml**

``` yaml
listing:
  template: /blog.ejs
```

</div>

PR #14513, opened 2026-05-14
Add `recursive` option to anchor listing `contents` globs to the listing directory

### LINKED ISSUE #14512: Listing `contents` glob silently matches files in sibling subdirectories (smart-glob `**/` prefix leaks across the project)
### I have:

- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest) (1.9.37)

### Bug description

A listing configured with `contents: "posts/*.qmd"` (or any non-`/`-anchored path/glob) silently matches files in **any directory named `posts/` anywhere in the project**, not just the `posts/` directory adjacent to the listing page.

This breaks the natural use case of a multi-section / bilingual website where two listings live at different depths and each owns its own `posts/` subdirectory.

Likely related (but a different symptom of the same path-handling area): #13677.

### Steps to reproduce

Minimal project:

```
project/
├── _quarto.yml          # project.type: website
├── index.qmd            # listing.contents: "posts/*.qmd"
├── posts/foo.qmd
└── ko/
    ├── index.qmd        # listing.contents: "posts/*.qmd"
    └── posts/foo.qmd
```

`_quarto.yml`:

```yaml
project:
  type: website
```

`index.qmd`:

```yaml
---
title: EN
listing:
  contents: "posts/*.qmd"
---
```

`ko/index.qmd`:

```yaml
---
title: KO
listing:
  contents: "posts/*.qmd"
---
```

Each `posts/foo.qmd` and `ko/posts/foo.qmd` just contains a title + date.

Render and inspect `_site/listings.json` (or open the rendered `index.html`). The English listing in `index.html` includes **both** `/posts/foo.html` and `/ko/posts/foo.html`.

`quarto render index.qmd --log-level debug` confirms:

```
[listing] Contents: posts/*.qmd
[listing] matches 2 files:
[listing] Reading file .../project/posts/foo.qmd
[listing] Reading file .../project/ko/posts/foo.qmd     ← unexpected
```

### Expected behavior

`contents: "posts/*.qmd"` from a listing in `index.qmd` should match only `dirname(index.qmd)/posts/*.qmd` — i.e., posts under the listing's own directory. Cross-section matching is surprising for the multi-section / i18n use case.

### Actual behavior

The glob `posts/*.qmd` is silently rewritten to `**/posts/*.qmd`, so it also matches `ko/posts/foo.qmd`.

### Root cause

In `src/core/path.ts`, `resolveGlobs.asFullGlob`:

```ts
if (!glob.startsWith("/")) {
  if (smartGlob && (!options || !options.explicitSubfolderSearch)) {
    return "**/" + glob;   // prepends `**/` to relative globs
  }
  ...
}
```

`useSmartGlobs()` returns `true` by default when no `options` is passed.

In `src/project/types/website/listing/website-listing-read.ts`, `filterListingFiles` calls `resolvePathGlobs(...)` (and `filterPaths(...)`) **without `GlobOptions`**, so `smartGlob` is `true` and every relative entry in `contents` is silently widened to `**/<glob>`. This is the design that breaks parallel-section listings.

`GlobOptions.explicitSubfolderSearch` already exists in `path.ts` precisely for this case (per its inline comment: *"set this to true to never prepend `**/` to globs to create nested directory searching"*) — it's just never wired through from the YAML.

### Workaround

Project-rooted leading-`/` paths work for a listing whose directory equals the project root:

```yaml
listing:
  contents: "/posts/*.qmd"
```

`asFullGlob` strips the leading `/` and skips the `**/` prefix. But this is asymmetric: from a non-root listing (e.g. `ko/index.qmd`) the `/`-prefix is resolved against `project.dir` in `expandGlob` and against `dirname(source)` in `resolvePathGlobs`, so the path doubles up and matches nothing. The remaining workarounds are:

- rename one of the colliding directories (e.g. `ko/articles/` instead of `ko/posts/`),
- enumerate files explicitly in `contents` (brittle, and also fails when basenames collide across sections — same `**/`-prefix bug).

### Suggested fix

Expose the existing internal `explicitSubfolderSearch` knob as a per-listing YAML option, preserving today's default (smart-glob `true`). E.g.:

```yaml
listing:
  contents: "posts/*.qmd"
  recursive: false     # anchored to dirname(listing.qmd); does not match sibling subdirs
```

Happy to send a PR (small: ~25 lines across `website-listing-shared.ts`, `website-listing-read.ts`, and the `website-listing` schema in `definitions.yml`). Naming open to maintainer preference (`recursive` / `anchored` / `match-subfolders` / …).

### Your environment

- Quarto CLI 1.9.37
- Linux (Ubuntu 22.04, kernel 5.15)
- Project type: `website`

MilkClouds:
## Summary

Adds a new `recursive: false` option to listing configuration that anchors `contents` globs to the listing's own directory, preventing them from matching files in sibling subdirectories elsewhere in the project.

Closes #14512. Related: #13677 (different symptom in the same `listing` path-handling area).

## Motivation

Today, Quarto's listing applies "smart glob" semantics that silently prepends `**/` to any non-`/`-anchored entry in `contents`. This is intentional and helpful for nested post folders, but it has a surprising side effect in multi-section / bilingual projects:

```
project/
├── index.qmd              # listing: contents: "posts/*.qmd"
├── posts/foo.qmd
└── ko/
    ├── index.qmd          # listing: contents: "posts/*.qmd"
    └── posts/foo.qmd
```

The English listing matches **both** `posts/foo.qmd` and `ko/posts/foo.qmd` because `posts/*.qmd` is silently rewritten to `**/posts/*.qmd`. Debug log:

```
[listing] Contents: posts/*.qmd
[listing] matches 2 files:
[listing] Reading file .../project/posts/foo.qmd
[listing] Reading file .../project/ko/posts/foo.qmd     ← unexpected
```

`GlobOptions.explicitSubfolderSearch` in `src/core/path.ts` already controls exactly this — per its inline comment, *"set this to true to never prepend `**/` to globs to create nested directory searching"*. It's just never wired through from the YAML.

## Behavior

Default behavior is preserved (smart globs continue to match recursively):

```yaml
listing:
  contents: "posts/*.qmd"   # default: matches **/posts/*.qmd across project
```

Opt out with the new flag:

```yaml
listing:
  contents: "posts/*.qmd"
  recursive: false          # anchored to dirname(listing.qmd)
```

## Changes

Four source files, +29/−0:

- `src/project/types/website/listing/website-listing-shared.ts` — add `kRecursive` constant.
- `src/project/types/website/listing/website-listing-read.ts` — in `filterListingFiles`, read `listing[kRecursive]` and pass `{ mode: "auto", explicitSubfolderSearch: true }` to `filterPaths` / `resolvePathGlobs` when `false`. (`recursive !== false`, so omitting the field or leaving the default keeps current behavior.)
- `src/resources/schema/definitions.yml` — add `recursive` field to the `website-listing` schema with user-facing docs.
- `news/changelog-1.10.md` — entry under Projects → Websites.

I did regenerate the schema artifacts locally with `quarto dev-call build-artifacts` to confirm the description text doesn't break code generation (an earlier draft used `` `**/` `` literally, which terminated the block comment in `schema-types.ts` and broke `validate-bundle`). The regenerated artifact diffs also included drift from earlier source changes that hadn't been re-emitted yet (e.g. the "Supports markdown formatting" descriptions from #14503), so per the maintainer convention (`20cb9d252` etc.) I'm **leaving artifact regeneration out of this PR** and letting it ride along with the next periodic `artifacts` commit.

## Test plan

- [x] Manual: minimal bilingual repro (`posts/foo.qmd` + `ko/posts/foo.qmd`) — listing with `recursive: false` excludes the sibling. Verified locally that `listings.json` matches the listing's directory only.
- [x] Default behavior unchanged for projects without the new field (omitted → `recursive !== false` is true → no `GlobOptions` passed → existing code path).
- [x] `quarto dev-call build-artifacts` produces a `deno fmt --check`-clean `schema-types.ts` with the new field, confirming the schema description does not break the type generator.
- [ ] Docs PR to `quarto-dev/quarto-web` to follow after this is merged.

Open to maintainer suggestions on the field name (`recursive` vs `anchored` vs `match-subfolders` vs …) — happy to rename.

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -32,6 +32,7 @@ All changes included in 1.10:
 ### Websites
 
 - ([#13565](https://github.com/quarto-dev/quarto-cli/issues/13565), [#14353](https://github.com/quarto-dev/quarto-cli/issues/14353)): Fix sidebar logo not appearing on secondary sidebars in multi-sidebar website layouts.
+- ([#14512](https://github.com/quarto-dev/quarto-cli/issues/14512)): Add `recursive: false` option to `listing` to anchor `contents` globs to the listing's own directory. Prevents a listing's `contents: posts/*.qmd` from also matching files in sibling subdirectories such as `ko/posts/*.qmd`, which is the common breakage in multi-section / bilingual websites.
 
 ## Commands
 
--- a/src/project/types/website/listing/website-listing-read.ts
+++ b/src/project/types/website/listing/website-listing-read.ts
@@ -20,6 +20,7 @@ import { existsSync } from "../../../../deno_ral/fs.ts";
 import { Format, Metadata } from "../../../../config/types.ts";
 import {
   filterPaths,
+  GlobOptions,
   pathWithForwardSlashes,
   resolvePathGlobs,
   safeExistsSync,
@@ -68,6 +69,7 @@ import {
   kListing,
   kMaxDescLength,
   kPageSize,
+  kRecursive,
   kSortAsc,
   kSortDesc,
   kSortUi,
@@ -720,6 +722,14 @@ async function readContents(
   );
 
   const filterListingFiles = (globOrPaths: string[]) => {
+    // Default: smart globs (current behavior, recursive across the project).
+    // Opt out with `recursive: false` on the listing to anchor `contents`
+    // globs to the listing's own directory.
+    const recursive = listing[kRecursive] !== false;
+    const globOptions: GlobOptions | undefined = recursive
+      ? undefined
+      : { mode: "auto", explicitSubfolderSearch: true };
+
     const hasDefaultGlob = globOrPaths.some((glob) => {
       return glob === kDefaultContentsGlobToken;
     });
@@ -758,6 +768,7 @@ async function readContents(
         dirname(source),
         inputsWithoutSource,
         finalGlobs,
+        globOptions,
       );
 
       // If a glob points to a file that exists, go ahead and just include
@@ -776,6 +787,7 @@ async function readContents(
         dirname(source),
         finalGlobs,
         ["_*", ".*", "**/_*", "**/.*", source],
+        globOptions,
       );
     }
   };
--- a/src/project/types/website/listing/website-listing-shared.ts
+++ b/src/project/types/website/listing/website-listing-shared.ts
@@ -93,6 +93,7 @@ export const kTableHover = "table-hover";
 // Items to include
 export const kInclude = "include";
 export const kExclude = "exclude";
+export const kRecursive = "recursive";
 
 // Fields
 export const kFieldTitle = "title";
--- a/src/resources/schema/definitions.yml
+++ b/src/resources/schema/definitions.yml
@@ -1530,6 +1530,21 @@
             - string
             - ref: website-listing-contents-object
         description: "The files or path globs of Quarto documents or YAML files that should be included in the listing."
+      recursive:
+        boolean:
+          default: true
+          description:
+            short: "Match `contents` globs recursively across the project."
+            long: |
+              When `true` (the default), non-anchored globs in `contents`
+              (e.g. `posts/*.qmd`) are expanded with an implicit globstar
+              prefix and may match nested directories anywhere under the
+              project root.
+
+              Set to `false` to anchor `contents` globs to the listing's own
+              directory. Useful for multi-section or bilingual sites where two
+              listings each own a sibling `posts/` subdirectory and should not
+              pull each other's files.
       sort:
         anyOf:
           - boolean


mcanouil:
I don't think this is the right fix.

Avoiding `**/` to be prepended also when `./` in addition to `/` would probably be better, i.e., allowing explicit relative glob.
More generally, with the other issue,  it appears to need a more in depth change to path handling which this PR does not address or help with.

I would suggest to discuss the proposed fix on the issue before asking Claude to implement its own suggestion.

MilkClouds:
Closing this.

On further investigation, `./posts/*.qmd` already produces the anchored behavior this PR's `recursive` flag was trying to add — the `./` → `/` normalization in `asFullGlob` routes through the same branch as `/`-prefixed paths and skips the implicit `**/` expansion. Verified with a minimal bilingual repro.

So the right answer for the case I was hitting was a path-syntax I missed, not a new flag. Apologies for the noise — I should have exhausted simpler workarounds and discussed on the issue before opening this. Thanks @mcanouil for the redirect.

PR #14525, opened 2026-05-18
fix(typst-book): localize orange-book running header when lang is set

### LINKED ISSUE #14524: Typst book running header not localized when `lang:` is set to a non-English locale
For a Typst book project with `lang:` set to a non-English locale, the running header on chapter pages keeps the English `Chapter N.` prefix instead of the locale's word (e.g. `Chapitre N.` for `lang: fr`).

Minimal reproduction:

```yaml
# _quarto.yml
project:
  type: book

book:
  title: "Livre Test"
  author: "Auteur Test"
  chapters:
    - index.qmd
    - chapter1.qmd

lang: fr

format:
  typst:
    keep-typ: true
```

```markdown
<!-- index.qmd -->
# Préface {.unnumbered}

Préface.
```

```markdown
<!-- chapter1.qmd -->
# Introduction

Du contenu, assez long pour faire déborder sur une deuxième page et faire apparaître l'en-tête de page.
```

After `quarto render`, the running header on chapter pages reads `Chapter 1. Introduction`. Expected: `Chapitre 1. Introduction`.

The orange-book extension's `book(...)` accepts a `supplement-chapter` parameter (default `"Chapter"`) that drives the running header via a `show heading` rule:

https://github.com/quarto-dev/quarto-cli/blob/a8d0dcfee2394f9caba3f1576b371c7e3bbeb47b/src/resources/extension-subtrees/orange-book/_extensions/orange-book/typst/packages/preview/orange-book/0.7.1/lib.typ#L311

https://github.com/quarto-dev/quarto-cli/blob/a8d0dcfee2394f9caba3f1576b371c7e3bbeb47b/src/resources/extension-subtrees/orange-book/_extensions/orange-book/typst/packages/preview/orange-book/0.7.1/lib.typ#L419

The Quarto-side `typst-show.typ` never passes `supplement-chapter` to `book.with(...)`, so the default English value applies:

https://github.com/quarto-dev/quarto-cli/blob/a8d0dcfee2394f9caba3f1576b371c7e3bbeb47b/src/resources/extension-subtrees/orange-book/_extensions/orange-book/typst-show.typ

The localized value exists in `_language-fr.yml` as `crossref-ch-prefix: "Chapitre"`:

https://github.com/quarto-dev/quarto-cli/blob/a8d0dcfee2394f9caba3f1576b371c7e3bbeb47b/src/resources/language/_language-fr.yml#L93

It is exposed as a Lua filter param via `languageFilterParams` but never written into Pandoc YAML metadata where the `typst-show.typ` template variable would resolve. The existing `list-of-figure-title` and `list-of-table-title` pipes in the same template are affected the same way and currently render as empty strings when the user enables `lof: true` / `lot: true`.

cderv:
When rendering a Typst book with `lang:` set to a non-English locale, the orange-book running header keeps the English `Chapter N.` prefix instead of the locale's word (e.g. `Chapitre N.` for `lang: fr`).

## Root cause

The orange-book `book(...)` exposes a `supplement-chapter` parameter (default `"Chapter"`) consumed by a `show heading.where(level: 1): set heading(supplement: ...)` rule that drives the running header. The Quarto-side `typst-show.typ` never passed this parameter to `book.with(...)`, so the English default applied regardless of `lang:`. The same gap silently broke the existing `list-of-figure-title` / `list-of-table-title` pipes — they rendered as empty strings.

The localized values live in `format.language["crossref-ch-prefix"]` (loaded from `_language-<lang>.yml`). Until #14530, no general channel exposed `format.language` to Pandoc templates without leaking keys into rendered output via `+yaml_metadata_block`.

## Fix

Consumes the `$quarto.language.*$` Pandoc template-variable namespace introduced in #14530. The orange-book `typst-show.typ` migrates all three pipes (`supplement-chapter`, `list-of-figure-title`, `list-of-table-title`) to the new namespaced form via [quarto-ext/orange-book#4](https://github.com/quarto-ext/orange-book/pull/4), pulled in via `quarto dev-call pull-git-subtree orange-book`. The extension version is bumped to `0.2.0`.

The interim Lua `surfaceParamToMeta` mechanism (which copied the three keys into Pandoc meta and triggered a YAML-leakage regression on `tests/docs/smoke-all/2024/01/22/8389.md.snapshot`) is removed.

User-override path via the `crossref:` block (`$crossref.lof-title$` / `$crossref.lot-title$`) is preserved as the first cascade branch in the orange-book template.

`list-of-figure-title` and `list-of-table-title` use nested `$if(...)$` guards so on pre-#14530 Quarto the argument is omitted entirely and `book()`'s built-in defaults apply. `supplement-chapter` gates on `$if(quarto.language.crossref-ch-prefix)$...$endif$` so older Quarto versions fall through to `book()`'s `"Chapter"` literal default.

## Related upstream observation

While auditing the migration, the orange-book `book()` implementation (`@preview/orange-book:0.7.1` lib.typ:606-608) was found to call `my-outline-sec(...)` unconditionally — so `lof: false` / `lot: false` suppresses only the title arg, not the outline section itself. Pre-existing behavior, out of scope here. Tracked at [quarto-ext/orange-book#5](https://github.com/quarto-ext/orange-book/issues/5).

## Documentation

`llm-docs/localization-architecture.md` (delivered with #14530) is updated in this branch: the in-flight forward-reference at the orange-book paragraph is replaced with the shipped channel-2d form, and frontmatter `main_commit` / `analyzed_date` are refreshed per `.claude/rules/llm-docs-maintenance.md`.

The subtree workflow guards introduced earlier in this branch (`.claude/rules/extension-subtrees.md`, `dev-docs/subtree-extensions.md` additions) remain.

> [!IMPORTANT]
> This branch contains subtree-pull commits and **cannot be rebased** — merge to `main` without rebasing. A merge commit from `upstream/main` is included on the branch to make the dependency on #14530 explicit.

Fixes #14524

## Checklist

I have (if applicable):

- [x] added new tests


gordonwoodhull:
This is the right fix! But currently it needs to be made on:

https://github.com/quarto-ext/orange-book

Then once it’s on main on that repo, run

```
quarto dev-call pull-git-subtree orange-book
```

to update the subtree. 

Modifying it directly will “work”, but only until the next time that repo is updated. (And won’t help anyone who might have forked that repo for their own changes.)

It’s debatable whether that repo needs to be a subtree extension instead of fully built-in. My rationale is that it is somewhat independent of Quarto; could get forked; we could switch to another fork or another extension. Glad to discuss.

https://github.com/quarto-dev/quarto-cli/blob/main/dev-docs/subtree-extensions.md

For that matter it’s also possible to push your changes back to the repo with `git subtree push` but I haven’t tried that.

cderv:
How did I forgot !! :facepalm: We discussed it. Sorry for this and thanks for the reminder.

gordonwoodhull:
You’ll also find that you can’t rebase the subtree & will get weird path errors if you try. So you’ll need to merge the branch without rebasing.

(The complexity has to go somewhere I guess. Subtrees are still less confusing than submodules IMO.)

cderv:
`git subtree push` was not adapted here. It was inspected al 20000+ commits to split before pushing the one change. 

So now I understand how it works, and it is a nice way to externalize some work, and have monorepo + external repo. 

Anyway, I did https://github.com/quarto-ext/orange-book/pull/3 and then rebase the branch to edit change before calling dev-call command. 

It should be good.

cderv:
I don't like the way those language var are exposed. So I think want to make a deeper feature fix by leveraging newest `--variable-json` for language keys. I'll look into this.

Issue #14466, opened 2026-05-01
Document localisation of cross-reference titles/prefixes via the language key


mcanouil:
### What would you like to do?

Give feedback or suggest an improvement

### Description

Follow-up to <https://github.com/orgs/quarto-dev/discussions/14465>.

The discussion confirmed that localisation of cross-reference titles and prefixes (for built-in floats, theorem environments, callouts, and custom kinds) is fully supported but not documented on <https://quarto.org>.
Users currently have to read `_language.yml` and infer the key naming convention from there.

### What is supported but undocumented

Overriding cross-reference titles and prefixes for built-in floats and theorem environments via the `language:` key, e.g.:

```yaml
---
lang: es
language:
  crossref-fig-title: "Figura"
  crossref-tbl-title: "Tabla"
  crossref-eq-prefix: "Ecuación"
---
```

The same mechanism works for callouts (`crossref-nte-prefix`, `crossref-tip-prefix`, `crossref-wrn-prefix`, `crossref-imp-prefix`, `crossref-cau-prefix`) and for any custom cross-reference kind, because the language filter passes through any key matching `crossref-*-(title|prefix)`.

For the 11 float/theorem types (`fig`, `tbl`, `lst`, `thm`, `lem`, `cor`, `prp`, `cnj`, `def`, `exm`, `exr`), `crossref-{type}-prefix` defaults to `crossref-{type}-title` when omitted.

The per-language alternates pattern already documented for `title-block-*` keys works for `crossref-*` keys too:

```yaml
lang: es
language:
  es:
    crossref-fig-title: "Figura"
  fr:
    crossref-fig-title: "Figure"
```

### Pages that need updating on quarto.org

- `docs/authoring/language.qmd` (Alternate Language section): the only example currently uses `title-block-*` keys.
  Add an example that uses `crossref-*-title` / `crossref-*-prefix`, and cross-link to the cross-references docs.
- `docs/authoring/cross-reference-options.qmd`: clarify the relationship between `crossref: { fig-title: ..., fig-prefix: ... }` (per-document overrides) and `language: { crossref-fig-title: ..., crossref-fig-prefix: ... }` (locale-aware overrides), and list the supported keys.
- `docs/authoring/cross-references.qmd` callout section / `docs/authoring/_cross-references-callouts.qmd`: add a localisation snippet showing how to translate callout prefixes such as `crossref-nte-prefix`.
[...]

PR #14530, opened 2026-05-20
Add `quarto.*` Pandoc template variable namespace

cderv:
Pandoc templates today cannot see Quarto's resolved localized strings
without per-key meta copies. Those copies leak into the rendered YAML
header for writers that emit `yaml_metadata_block` (markdown, native,
json) and require format-by-format gating to suppress.

`format.language` (the merged result of Quarto's bundled
`_language.yml` plus per-locale and per-document overrides) is now
surfaced to Pandoc templates via the defaults-file `variables:`
section, under a reserved internal namespace `quarto.*`:

```yaml
variables:
  quarto:
    language:
      crossref-ch-prefix: Chapter
      toc-title-document: Table of contents
      # ... every key from format.language
```

Templates access keys via dotted form:

```
$quarto.language.crossref-ch-prefix$
$quarto.language.toc-title-document$
```

The defaults-file `variables:` section is not serialized back into
rendered output by any writer, so the leakage problem goes away
without per-format gating.

## Why a new namespace

`quarto.*` is reserved as an internal, Quarto-owned namespace inside
Pandoc template variables. Users overriding a localized string keep
using the existing top-level `language:` YAML key (see #14466). The
schema does not surface `variables.quarto.*` as a user-facing option,
and on collision the user-set value wins, so an explicit YAML escape
hatch keeps working.

## Contribution point

A single file — `src/command/render/quarto-template-variables.ts` —
is the contribution point for new keys under `quarto.*`. Adding a new
internal value to expose to templates is a typed two-step:

1. Add a field to `QuartoTemplateVariables` (use the tightest
   existing type, e.g. `FormatLanguage`).
2. Add a contribution branch in `buildQuartoTemplateVariables` that
   reads from `PandocOptions` and assigns to that field.

The wiring into the defaults file (`generateDefaults` in
`src/command/render/defaults.ts`) is single-point and does not need
changes per new key.

## Opportunities this opens

- Future internal exposures fit the same channel without any further
  template-channel design work — `$quarto.format.X$`,
  `$quarto.project.X$`, `$quarto.version$`, etc.
- The orange-book Typst running-header bug (#14524, tracked as a
  separate PR) becomes a one-line template edit on top of this
  channel, replacing a Lua filter workaround.
- Templates that today read flat `$toc-title$` / `$abstract-title$`
  (populated via per-key meta copies in `src/command/render/pandoc.ts`)
  can migrate to `$quarto.language.toc-title-document$` /
  `$quarto.language.section-title-abstract$`. Once migrated, those
  per-key meta copies and their per-format gating can be removed.

## Localization architecture

Channel 2d (defaults-file `variables:`) is now documented as the
canonical bulk channel for new language keys, alongside Lua filter
params and Pandoc meta. See `llm-docs/localization-architecture.md`.

## Test plan

- [x] `tests/docs/smoke-all/markdown/lang-fr-template-resolve.qmd` — custom Pandoc template + `lang: fr` resolves `$quarto.language.crossref-ch-prefix$` to "Chapitre"
- [x] `tests/docs/smoke-all/markdown/lang-fr-no-yaml-leakage.qmd` — regression guard: `crossref-*` and `toc-title-document` stay out of the rendered markdown YAML header
- [x] `tests/docs/smoke-all/markdown/lang-fr-user-override-quarto-variables.qmd` — user-set `variables.quarto.language.crossref-ch-prefix` wins on collision
- [x] `tests/docs/smoke-all/markdown/lang-ja-template-resolve.qmd` — unicode round-trip: `lang: ja` resolves to "チャプター"
- [x] `tests/unit/render/quarto-template-variables.test.ts` — builder contract: empty in → undefined out; reference-passing; empty-object case

Related to #14524


cscheid:
(I think this is a great idea.)

cderv:
Documentation PR 
- https://github.com/quarto-dev/quarto-web/pull/2038

Issue #14367, opened 2026-04-15
Quarto 1.9.36 fails in enterprise Windows: Dart Sass invoked via temp batch file is blocked by Group Policy (works in 1.8.25)


m20l22:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

Quarto 1.9.36 fails with `Theme file compilation failed` on Windows systems where Group Policy blocks execution of `.bat` files from the `%TEMP%` directory (common in corporate/enterprise environments).

The root cause is a regression in how Dart Sass is invoked. In Quarto 1.8.25, `sass.bat --version` was called **directly** from the Quarto installation directory (`C:\Program Files\...\dart-sass\sass.bat`), which is in a trusted/allowed path. In Quarto 1.9.36, the same Sass check is now routed through a **temporary wrapper batch file** (`quarto-safe-exec*.bat`) written to `%TEMP%` and executed via `cmd /c`. On locked-down systems, Group Policy (Software Restriction Policies / AppLocker) blocks `.bat` execution from `%TEMP%`, causing the Sass check to fail.

Note that the `quarto-safe-exec` mechanism is also used in 1.8.25 for **optional** tool checks (TinyTeX, Chromium, LaTeX, Chrome) — those also fail with exit code 1, but they are handled gracefully as "not installed"/"not detected." The difference is that in 1.9.36, **Dart Sass** — a critical, non-optional dependency — is now also called via this mechanism, and its failure is fatal.

### Steps to reproduce

1. Use a Windows machine with Group Policy / AppLocker rules that block execution of `.bat` files from the user's `%TEMP%` directory (common in enterprise environments).
2. Install Quarto 1.9.36 (e.g., bundled with Positron).
3. Run:

```
quarto check --log-level debug
```

### Actual behavior

`quarto check` crashes with `Theme file compilation failed`:

```
[execProcess] cmd /c C:\Users\<USERNAME>\AppData\Local\Temp\quarto-safe-exec91e465e27a39d068.bat
[execProcess] Success: false, code: 1
[DART path]   : C:\Program Files\Positron\resources\app\quarto\bin\tools\x86_64\dart-sass\sass.bat
[DART args]   : --version
[DART stdout] :
[...]

Issue #14267, opened 2026-03-25
[Bug] Windows execution wrapper mangles paths with accented characters, breaking external commands like dart-sass (v1.9.36)


sbwiecko:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

When attempting to render a **Quarto website (html) project** on Windows using version 1.9.36, the render fails during the Dart Sass bundling step bundling step with an `os error 2` (file not found) if the project path contains an accented character (e.g., `é` in `C:\Users\Sébastien\...`). 

### Steps to reproduce

1. Create a Quarto project inside a directory path containing an accented character on Windows (e.g., `C:\Users\Sébastien\project\`)
2. Run `quarto render --log-level debug > quarto_broken_log.txt 2>&1` using Quarto v1.9.36
3. The render crashes with the following stack trace:

```text
Quarto version: 1.9.36
Render: Using current directory (C:\Users\Sébastien\Documents\website) as implicit input

[1/8] about.qmd
Loaded deno-dom-native
Creating SassCache at C:\Users\Sébastien\Documents\website\.quarto\quarto-session-temp6eb66806188f8e02\sass
Returning SassCache at C:\Users\Sébastien\Documents\website\.quarto\quarto-session-temp6eb66806188f8e02\sass
SassCache.getOrSet(...)
SassCache.getFromHash(hash=6c22de33446cb5b69a0df62f1a98c15c, inputHash=844c4805a38aa141738dc7405baa30e2, force=undefined)
  cache miss
  cache miss, setting
SassCache.setFromHash(6c22de33446cb5b69a0df62f1a98c15c, 844c4805a38aa141738dc7405baa30e2), ...
[execProcess] cmd /c C:\Users\SBASTI~1\AppData\Local\Temp\quarto-safe-exec27c5869bc2933c5d.bat
[execProcess] Success: false, code: 1
[DART path]   : C:\Users\Sébastien\AppData\Local\Programs\Quarto\bin\tools\x86_64\dart-sass\sass.bat
[...]

Issue #13997, opened 2026-02-04
Windows: Error "C:\Program spelling error"


dmenne:
### I have:

- [ ] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [ ] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Expected behavior

I noted this when I installed quarto 1.9.19 for all users:

- Use any article or book
- Render.
- German: "C:\Program ist entweder falsch geschrieben" (this is the complete text it means ..is either incorrectly spelled" (i.e. no complete sentence)

After reinstalling it for the current user, all is fine.

```
PS>quarto preview "c:/Users/Dieter/Documents/mydoc/index.qmd" --no-browser --no-watch-inputs
Rendering qmd embeds [index.qmd]
Rendering output notebook [index.qmd]
Rendering HTML preview [index.qmd]
ERROR: Theme file compilation failed:

Der Befehl "C:\Program" ist entweder falsch geschrieben oder

Stack trace:
    at dartCommand (file:///C:/Program Files/Quarto/bin/quarto.js:78648:11)
    at eventLoopTick (ext:core/01_core.js:179:7)
    at async dartCompile (file:///C:/Program Files/Quarto/bin/quarto.js:78608:3)
    at async file:///C:/Program Files/Quarto/bin/quarto.js:87080:9
    at async SassCache.setFromHash (file:///C:/Program Files/Quarto/bin/quarto.js:79525:11)
    at async compileSass (file:///C:/Program Files/Quarto/bin/quarto.js:86891:19)
    at async resolveSassBundles (file:///C:/Program Files/Quarto/bin/quarto.js:124554:17)
    at async resolveExtras (file:///C:/Program Files/Quarto/bin/quarto.js:129213:14)
    at async runPandoc (file:///C:/Program Files/Quarto/bin/quarto.js:128642:20)
    at async renderPandoc (file:///C:/Program Files/Quarto/bin/quarto.js:130295:24)
```

### Quarto check output

```

PS>quarto check                                                                                         
Quarto 1.9.19
[>] Checking environment information...
      Quarto cache location: C:\Users\Dieter\AppData\Local\quarto
[>] Checking versions of quarto binary dependencies...
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.4.5: OK
[...]

PR #14535, opened 2026-05-21
[backport] Fix dart-sass failure on Windows usernames containing ampersand

### LINKED ISSUE #14534: Quarto render fails when Windows username contains ampersand
### I have:

### Bug description

Originally reported via RStudio, but have confirmed it is a Quarto issue (happens from terminal, outside RStudio, with a trivial example). Details in https://github.com/rstudio/rstudio/issues/17712.

In short, when using Quarto via a Windows user account that has an ampersand in the username, rendering will not work. This worked with Quarto 1.8, but fails with 1.9.

There's a related bug that happens with both 1.8 and 1.9 if you install Quarto per-user with this `gary&ritchie` style username, Quarto fails with a different error message (didn't capture it). Installing for all users (i.e. outside the troublesome path) avoids this problem.

### Steps to reproduce

- Create a Windows user account that has an ampersand, e.g. `C:\Users\gary&ritchie\` and login as that user
- With Quarto 1.9.37, and a trivial Quarto doc (I used whatever RStudio creates for a new Quarto doc project) somewhere in the user's profile (e.g. `C:\Users\gary&ritchie\Documents\bug-repro\bug-repro.qmd`) try to render using the same command-line that RStudio uses

Here's a transcript showing this failing with 1.9, and working with 1.8:
```
C:\Users\gary&ritchie\Documents\bug-repro>quarto --version
1.9.37

C:\Users\gary&ritchie\Documents\bug-repro>quarto preview bug-repro.qmd --to default --no-watch-inputs --no-browse


processing file: bug-repro.qmd
1/3
2/3 [unnamed-chunk-1]
3/3
output file: bug-repro.knit.md

ERROR: Theme file compilation failed:

'C:\Users\GARY' is not recognized as an internal or external command,

Stack trace:
    at processResult (file:///C:/Program Files/Quarto/bin/quarto.js:79152:13)
    at dartCommand (file:///C:/Program Files/Quarto/bin/quarto.js:79169:12)
    at eventLoopTick (ext:core/01_core.js:179:7)
    at async dartCompile (file:///C:/Program Files/Quarto/bin/quarto.js:79120:3)
    at async file:///C:/Program Files/Quarto/bin/quarto.js:87618:9
    at async SassCache.setFromHash (file:///C:/Program Files/Quarto/bin/quarto.js:80054:11)
    at async compileSass (file:///C:/Program Files/Quarto/bin/quarto.js:87429:19)
    at async resolveSassBundles (file:///C:/Program Files/Quarto/bin/quarto.js:125158:17)
    at async resolveExtras (file:///C:/Program Files/Quarto/bin/quarto.js:129822:14)
    at async runPandoc (file:///C:/Program Files/Quarto/bin/quarto.js:129246:20)


-- downgrade to 1.8.27, problem goes away --

C:\Users\gary&ritchie\Documents\bug-repro>quarto --version
1.8.27

C:\Users\gary&ritchie\Documents\bug-repro>quarto preview bug-repro.qmd --to default --no-watch-inputs --no-browse


processing file: bug-repro.qmd
1/3
2/3 [unnamed-chunk-1]
3/3
output file: bug-repro.knit.md

pandoc
  to: html
  output-file: bug-repro.html
  standalone: true
  section-divs: true
  html-math-method: mathjax
  wrap: none
  default-image-extension: png
  variables: {}

metadata
  document-css: false
  link-citations: true
  date-format: long
  lang: en
  editor: visual
  title: bug-repro

Output created: bug-repro.html

Watching files for changes
Browse at http://localhost:4691/bug-repro.html
Listening on http://127.0.0.1:4691/

```

### Actual behavior

- With Quarto 1.8.25 and 1.8.27, this works (RStudio 2026.01 bundled 1.8).
- With Quarto 1.9.37 it fails with `'C:\Users\GARY' is not recognized as an internal or external command,
operable program or batch file.` (RStudio 2026.05 bundled 1.9)

### Expected behavior

Everything works despite the ampersand in the user profile name.

Ideally, installing Quarto per-user with such a name should also work, but probably a separate issue.

### Your environment

Windows-11 Pro, x64

### Quarto check output

I installed Quarto standalone via the MSI installer, using 1.8.27 and 1.9.37.

cderv:
> [!IMPORTANT]
> Backport from #14273

When the Windows user profile path contains `&` (e.g., `C:\Users\Tom & Jerry\`), `quarto render` of any document with sass fails with `'C:\Users\Tom' is not recognized as an internal or external command`. The ampersand is a `cmd.exe` command separator, so the temp `.bat` wrapper that `safeWindowsExec` writes is split mid-path and the second half is interpreted as a separate command.

## Root Cause

v1.9 introduced `safeWindowsExec` to handle spaced paths (#13997), which writes a temp `.bat` and runs it through `cmd.exe /c`. Any `&` in the program path lets `cmd.exe` split the line. The existing #14367 fallback to direct `sass.bat` execution does not help — Windows still spawns `cmd.exe` internally to interpret any `.bat`, and the ampersand splits again there.

## Fix

Backport of the `dart.exe` + `sass.snapshot` direct-invocation rewrite from #14273. Calling `dart.exe` via `Deno.Command` bypasses `cmd.exe` entirely, removing the whole class of `.bat`-related issues on Windows (ampersand splitting, OEM/UTF-8 code page mismatch, enterprise group policy blocks).

The bundled dart-sass version is identical between v1.9 and main (`DARTSASS=1.87.0`), so the `src/dart.exe` + `src/sass.snapshot` layout assumed by the fix is the same on disk in v1.9.

This subsumes the #14367 fallback approach (no `.bat` = no policy block) and reinforces the #14267 accented-path fix already backported via #14274.

Out of scope: other `safeWindowsExec` callers (`texlive.ts`, `zip.ts`, `shell.ts`) still go through `cmd.exe` and remain vulnerable to `&` in paths. Those have separate upstream fixes and are not covered by this backport.

## Test Plan

- [x] Render a document with sass on Windows where the user profile path contains `&`
- [x] Render a document with sass on Windows with a normal user profile (no regression)
- [x] Render a document with sass on macOS/Linux (non-Windows path unchanged)
- [x] Render a document with `QUARTO_DART_SASS` env override pointing to a custom `sass.exe`
- [x] `unit/dart-sass.test.ts` and `unit/windows-exec.test.ts` pass on Windows CI

Fixes #14534


cderv:
@cscheid I am thinking of backporting this fix that I initially did not backport because it was bigger change that I would expect for a backport. 

However it impacts RStudio users (#14534) for which quarto was working in v1.8 and no more in v1.8. It was reported in our repo from a report on RStudio IDE by @gtritchie. 

It is scoped to Windows user only, and should not create problem because it is improvement but... 🤷‍♂️ you never know. 

What do you think ? This is so next RStudio IDE could maybe bundle a patch v1.9 before we do release a stable v1.10. @gtritchie if we do a patch release with such fix, is there greater probability to have a v1.9 patch bundled into next RStudio IDE ? 

Anyhow, this would be available for user to download stable themself. 

@cscheid I just prefered to have your go on this before merging. (Reminder: the fix is already in v1.10.3 following #14273)

cscheid:
I think a backport is ideal here, assuming everything goes smoothly.

But I _also_ think that using `&` in a username is, um, ill-advised. If it turns out that this fix is bad for other reasons, we should feel ok with reverting it and choosing not to fix.

src/core/dart-sass.ts
```diff
@@ -1,20 +1,18 @@
 /*
  * dart-sass.ts
  *
- * Copyright (C) 2020-2022 Posit Software, PBC
+ * Copyright (C) 2020-2025 Posit Software, PBC
```
cscheid: If we're going to fix this, it's 2026 :D

cscheid (approved):

gtritchie:
The upcoming release of RStudio (2026.05) is already locked and loaded, we can't take changes for anything unless incredibly serious. Handling ampersands in usernames would not meet the bar, IMO.

Having a patched 1.9 would be _nice_ since we could tell impacted users to install that separately, without having to downgrade to 1.8.

We are doing monthly releases of RStudio now, and I could incorporate this into 2026.06 if available in the next few weeks.

src/core/dart-sass.ts
gtritchie: Took me a long time to convince Claude Code that it was 2026. I don't think it believed me.

cderv:
```suggestion
 * Copyright (C) 2020-2026 Posit Software, PBC
```

cderv:
Thanks ! I'll merge this and include in next patch release. 

This way you can consider it for next RStudio IDE - the patch will have other fix worth having in a new IDE release I think.

Issue #14281, opened 2026-03-29
Lots of quarto_ipynb files created


memeplex:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

When previewing qmd files a number of auxiliary quarto_ipynb files are created and left alongside the source document. This is the case even if `keep-ipynb: false`.

### Steps to reproduce

Create test.qmd wih contents:

    ---
    execute:
      keep-ipynb: false
    ---

    ```{python}
    2 + 2
    ```

Then run `quarto preview test.qmd`.

### Actual behavior

I get:

```
 test.html  test.qmd  test.quarto_ipynb  test.quarto_ipynb_1  test.quarto_ipynb_2
```

### Expected behavior

Only ` test.html  test.qmd` with no clutter except when `keep-ipynb: true`.

### Your environment

VSCode 1.113.0
macOS Tahoe

### Quarto check output

```bash
❯ quarto check
Quarto 1.9.36
[✓] Checking environment information...
      Quarto cache location: /Users/carlos/Library/Caches/quarto
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.4.5: OK
      Typst version 0.14.2: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.9.36
      Path: /Users/carlos/Desktop/quarto-test/.venv/lib/python3.13/site-packages/quarto_cli/bin

[✓] Checking tools....................OK
      TinyTeX: (not installed)
      Chromium: (not installed)
      Chrome Headless Shell: (not installed)
      VeraPDF: (not installed)

[✓] Checking LaTeX....................OK
      Tex:  (not detected)

[✓] Checking Chrome Headless....................OK
      Using: Chrome found on system
      Path: /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
      Source: MacOS known location

[✓] Checking basic markdown render....OK

[✓] Checking R installation...........OK
      Version: 4.5.2
      Path: /opt/homebrew/Cellar/r/4.5.2_1/lib/R
      LibPaths:
        - /Users/carlos/.rlib
[...]

PR #14539, opened 2026-05-22
fix(preview): detect frontmatter format change on first request

### LINKED ISSUE #14533: `quarto preview` requires trying twice after format change
### I have:

### Bug description

Changing formats and then clicking on Preview in Positron requires doing this twice before the output format is actually changed.

### Steps to reproduce

_No response_

### Actual behavior

_No response_

### Expected behavior

_No response_

### Your environment

- Positron Version: 2026.06.0 build 113
- Mac OS Version 26.5 (25F71)

### Quarto check output

```bash
Quarto 1.10.3
      Quarto cache location: /Users/mine/Library/Caches/quarto
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.4.5: OK
      Typst version 0.14.2: OK
      Version: 1.10.3

      TinyTeX: (external install)
      VeraPDF: (not installed)

      Using: TinyTex
      Path: /Users/mine/Library/TinyTeX/bin/universal-darwin
      Version: 2026

      Using: Chrome found on system
      Source: MacOS known location

      Version: 4.5.2
      LibPaths:
      knitr: 1.51
      rmarkdown: 2.31

      Version: 3.14.3
      Path: /opt/homebrew/opt/python@3.14/bin/python3.14
      Jupyter: (None)

````

cderv:
When the frontmatter `format:` is edited during a preview session, `quarto preview` reports the first render request as compatible with the running process — so the IDE keeps the process alive and re-renders against the previous format. Only the second request after the edit returns 404 and triggers a restart with the new format.

## Scope

The fix needs a reliable "is any render in flight right now?" signal that didn't exist before, so the PR is layered:

1. **`HttpDevServerRenderMonitor` counter** (`src/core/http-devserver.ts`) — replaces the single `renderStart_` timestamp with an `inFlight_` counter and exposes `isRendering()`. Overlapping renders no longer falsely clear the gate.
2. **`trackInFlight()` helper** (same file) — wraps the render promise so the counter increments/decrements exactly once per render, regardless of what callers do with the resolved value.
3. **`ServeRenderManager.submitRender`** (`src/project/serve/render.ts`) — routes start/stop through `trackInFlight`; drops the paired `onRenderStart` / `onRenderStop` calls from `onRenderResult` / `onRenderError`.
4. **`previewRenderRequestIsCompatible` cache invalidation** (`src/command/preview/preview.ts`) — the actual #14533 fix, gated on `isRendering()`.
5. **Tests + docs** — unit test (`tests/unit/preview-compatibility.test.ts`), manual matrix (`tests/docs/manual/preview/`), `llm-docs/preview-architecture.md` update.

Steps 1–3 enable step 4. Commit order matches.

## Root Cause

`previewRenderRequestIsCompatible` resolves the file's current format via `previewFormat` → `renderFormats` → `fileExecutionEngineAndTarget`, which consults `fileInformationCache`. Cache invalidation runs inside `renderForPreview` (introduced by #14281) — _after_ the compatibility check has already returned its answer. So:

1. Click 1: compat check reads stale cache (old format), matches `flags.to`, returns 200. Render runs, invalidates cache, re-populates with the new frontmatter.
2. Click 2: compat check reads the fresh cache (new format), mismatches `flags.to`, returns 404. Extension restarts the process.

The compatibility check is factually wrong on the first request after a frontmatter format edit.

## Fix

Invalidate `fileInformationCache` for `request.path` at the top of `previewRenderRequestIsCompatible`, gated on two conditions:

- `request.format === undefined` — when the caller pins a format, `previewFormat` short-circuits before consulting the cache, so invalidation would just churn transient `.quarto_ipynb` cleanup for no benefit.
- `!HttpDevServerRenderMonitor.isRendering()` — see *Why the in-flight gate* below.

https://github.com/quarto-dev/quarto-cli/blob/c02028729e2ecad6900b2936e1e1f8d565d5e9bc/src/command/preview/preview.ts#L359-L383

## Why the in-flight gate

`invalidateForFile` calls `safeRemoveSync` on the cached `target.input` — the transient `.quarto_ipynb` that an in-flight render is writing or reading. On Windows this throws (file lock); on Linux it unlinks the file out from under the in-flight render. The in-flight render's own `renderForPreview` already invalidates and repopulates the cache at its start, so the cache reflects its view until it completes; a frontmatter edit made during the in-flight window is picked up on the next compatibility check after the render finishes.

A mutex around the cache was the alternative, but it would block the compatibility-check response (200/404) on the in-flight render — visible latency on every IDE-driven render request. The gate keeps the check non-blocking.

## Why a counter, not a timestamp

`HttpDevServerRenderMonitor` previously held a single `renderStart_: number | undefined`. `submitRender` called `onRenderStart` synchronously at queue time; `onRenderStop` fired from `onRenderResult` / `onRenderError`. With two overlapping renders:

- Render A: start sets `renderStart_`. Stop clears it.
- Render B (queued behind A): A's stop has already cleared `renderStart_`, so `isRendering()` would read "not rendering" while B is still queued or running.

A counter (`inFlight_++` on start, `inFlight_--` on stop) makes `isRendering()` return true while any render is outstanding. `lastRenderTime_` is captured on the 0→1 and N→0 transitions, preserving the existing measurement semantics.

## Why `trackInFlight` rather than leaving start/stop at the call site

Three concrete reasons the helper isn't a free-floating refactor:

- **Pairing safety.** Old code called `onRenderStart` in `submitRender` and the matching `onRenderStop` in `onRenderResult` / `onRenderError`. Anything that threw between submission and those handlers stranded the counter. Wrapping pairs start and stop on the same promise.
- **`RenderResult.error` propagation.** Renders fulfill with a `RenderResult` carrying `.error` instead of rejecting (see `render-files.ts`). A pure resolve/reject wrapper would mark a failed render as success. `successFromValue` lets the caller derive the stop signal from the fulfilled value: `submitRender` passes `(result) => !result.error`.
- **Predicate-throw safety.** If `successFromValue` itself throws, the counter must still decrement — the whole point of the wrapper is that *settlement* decrements. The wrapper catches inside the `.then`, decrements, then re-throws.

Result: `submitRender` becomes `trackInFlight(this.renderQueue_.enqueue(render), (r) => !r.error)`. `onRenderError` / `onRenderResult` no longer touch the monitor.

## For unchanged frontmatter

`previewFormat` repopulates the cache with the same value and the compatibility verdict is identical — only the cache lookup runs again. Cost: one cache re-read per IDE-driven render request when no render is in flight, no functional change. Architecture details in `llm-docs/preview-architecture.md` (Render Request Compatibility Check section).

## Testing

Unit: `tests/unit/preview-compatibility.test.ts` exercises the real `previewRenderRequestIsCompatible` → `previewFormat` → `renderFormats` → `fileExecutionEngineAndTarget` chain via `singleFileProjectContext` — no mocks. Nine cases:

- format-edit detection (`false`)
- unchanged frontmatter (`true`)
- reverse-direction format change typst → html (`false`)
- file path with spaces (`false`)
- in-flight gate: stale verdict (`true`) and cache-entry reference unchanged when `HttpDevServerRenderMonitor.isRendering()` is true
- `HttpDevServerRenderMonitor.isRendering` counter survives overlapping start/stop pairs (would fail the previous single-timestamp implementation)
- `request.version === 1` short-circuits to `true` regardless of format mismatch (RStudio path)
- pinned `request.format` matching `flags.to`: returns `true`, cache-entry reference unchanged
- pinned `request.format` differing from `flags.to`: returns `false`, cache-entry reference unchanged

All pass with the fix; the format-edit and counter cases fail without their respective code changes. Full unit suite: 385 passed | 0 failed | 6 ignored.

https://github.com/quarto-dev/quarto-cli/blob/c02028729e2ecad6900b2936e1e1f8d565d5e9bc/tests/unit/preview-compatibility.test.ts

CLI end-to-end against the dev binary with a single-file html document (token + `--no-watch-inputs`, browser registers as client):

| Click | File state                   | Pre-fix    | Post-fix |
|-------|------------------------------|------------|----------|
| 1     | `format: html`               | 200        | 200      |
| 2     | `format: typst` (after edit) | 200 stale  | 404      |
| 3     | `format: typst`              | 404        | 404      |

The pre-fix click 2 returning 200 is the bug; post-fix it returns 404 immediately, so the extension restarts on the first request after the edit instead of the second.

Manual matrix added at `tests/docs/manual/preview/README.md` (T20–T25, covering bidirectional format change, multi-format frontmatter, mid-render edit, path-with-spaces, and an unchanged-frontmatter regression guard) for IDE-level verification.

## Test Plan

- [x] `tests/unit/preview-compatibility.test.ts` passes
- [x] Manual T20 (html → typst) — single click switches output
- [x] Manual T25 (unchanged frontmatter) — no spurious restart

Fixes #14533


cderv:
Did a side-effect pass on this change before merge, focused on whether the new `fileInformationCache` invalidation in `previewRenderRequestIsCompatible` could break anything beyond the format-change case it targets — in particular `.ipynb` inputs and project vs single-file preview.

## What I checked

`.ipynb` as input — the new `invalidateForFile` call only deletes a file from disk when its cache entry is marked `transient: true`. Native notebooks are `transient: false` (set in `src/execute/jupyter/jupyter.ts`), where the input *is* the source; only `.quarto_ipynb` files generated from `.qmd`/percent scripts are transient. So a user's `.ipynb` is never touched — only the cache entry is dropped. The call reuses the same contract already live in `renderForPreview`, so no new deletion path is introduced.

In-flight tracking — `isRendering()` is now counter-based rather than a single timestamp, so overlapping render submits stay correct and the counter always decrements on settle (fulfill, reject, or a throwing success-predicate).

## Live preview runs (dev build + test venv)

- Native `.ipynb`: renders, source file stays intact, no transient `.quarto_ipynb` created.
- `.qmd` + Python (single file): repeated saves re-render correctly, exactly one transient at a time, no `_1`/`_2` accumulation.
- Website project: repeated saves re-render and converge to the latest edit, no accumulation.

## Automated

`tests/unit/preview-compatibility.test.ts` passes 14/14 — covers the overlap counter, predicate throws (including `throw undefined`), pinned-format requests (no invalidation), and file paths with spaces.

No regressions found — the change stays contained to the preview server.

PR #14560, opened 2026-05-29
Fix font fallback crashing lualatex on recent TeX Live

### LINKED ISSUE #14553: Error rendering PDF with ∈ symbol in Julia cells
### I have:

### Bug description

Julia allows the use of unicode like in

```julia
1 ∈ [1, 2, 3]
```

If I add a Julia cell with this operator, I get an error message when I attempt to render a PDF:

```
ERROR: 
compilation failed- error
You can't use `\spacefactor' in math mode.
\@->\spacefactor 
                 \@m {}
l.2920 If you don't know the \(\LaTeX
                                   \) name of a symbol, you can copy/paste it
```

This might be happening with similar mathematical symbols (e.g., `∈`, `⊆`).

### Steps to reproduce

Create a simple Julia document with the following cell and try to render to PDF:

```{julia}
1 ∈ [1, 2, 3]
```

### Actual behavior

Error message from TeX stack.

### Expected behavior

PDF output with Julia cell normally executed.

### Your environment

```
VSCode
Version: 1.117.0
Commit: 10c8e557c8b9f9ed0a87f61f1c9a44bde731c409
Date: 2026-04-21T16:12:14-07:00
Electron: 39.8.7
ElectronBuildId: 13841579
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Linux x64 5.15.202-1-MANJARO
```

### Quarto check output

```
Quarto 1.9.37
      Quarto cache location: /home/juliohm/.cache/quarto
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.4.5: OK
      Typst version 0.14.2: OK
      Version: 1.9.37
      Path: /opt/quarto/bin

      TinyTeX: v2025.02
      Chromium: (not installed)
      VeraPDF: (not installed)

      Using: TinyTex
      Path: /home/juliohm/.TinyTeX/bin/x86_64-linux
      Version: 2024

      Using: Chrome from QUARTO_CHROMIUM
      Path: /usr/bin/chromium

      Version: 4.5.3
      Path: /usr/lib64/R
      LibPaths:
        - /usr/lib/R/library
      knitr: (None)
      rmarkdown: (None)

      The knitr package is not available in this R installation.
      Install with install.packages("knitr")
      The rmarkdown package is not available in this R installation.
      Install with install.packages("rmarkdown")

      Version: 3.14.3
      Path: /usr/bin/python3
      Jupyter: 5.9.1
      Kernels: python3, julia-1.11, julia-1.10, julia-1.9

```

cderv:
When a document sets a font fallback (`mainfontfallback`, `monofontfallback`, ...) and renders to PDF with lualatex on recent TeX Live (luaotfload v3.29), lualatex crashes before producing a PDF and Quarto reports only a generic `compilation failed- error`.

## Root Cause

luaotfload appends an internal `;-fallback` feature to each fallback name and resolves it with `define_font`. Its font-name parser ends a name only at `:`, `(` or `/`, so a bare name like `"DejaVu Sans"` absorbs the appended feature (`DejaVu Sans;-fallback`), the lookup fails, `define_font` returns `nil`, and `luaotfload-fallback.lua` then dereferences the `nil`. The names reach luaotfload without a terminator, so the appended feature is never parsed as one.

Because the crash is a Lua runtime error with no `! ...Here is how much` block, `findLatexError` in `src/command/render/latexmk/parse-error.ts` extracted nothing, leaving the user with a generic failure and no signal.

## Fix

Normalize fallback names in a quarto-pre Lua `Meta` filter, `src/resources/filters/quarto-pre/font.lua`: append a terminating `:` only when the name has no luaotfload terminator (`:`, `(` or `/`). A bare `"DejaVu Sans"` becomes `"DejaVu Sans:"` so the appended `;-fallback` parses as a feature and the lookup resolves, while documented forms are left untouched — `"FreeSans:"` does not become `"FreeSans::"`, and `"Noto:mode=harf"` keeps no trailing colon. This matches the luaotfload maintainer's confirmation in https://github.com/latex3/luaotfload/issues/331 and jgm's guidance in https://github.com/jgm/pandoc/issues/11678 to add the colon only when one is not already present.

The filter is gated by a `has_font_fallback` flag in `src/resources/filters/normalize/flags.lua`, so it is skipped entirely when no fallback is configured. The vendored pandoc partial `src/resources/formats/pdf/pandoc/font-settings.latex` is left byte-identical to pandoc's: there is no upstream divergence to maintain, and `keep-tex` output stays free of injected plumbing.

The crash-detection diagnostic in `parse-error.ts` is kept as a safety net: if a fallback crash is still detected (e.g. a hand-written raw-LaTeX fallback), `findLatexError` returns an actionable message instead of a generic failure.

## Test Plan

- [ ] Render a PDF on lualatex (TeX Live 2026) with `monofont` plus `monofontfallback` entries that are bare, explicitly colon-terminated, and option-bearing — compiles with fallbacks active, no crash
- [ ] `tests/docs/smoke-all/2026/06/01/14553.qmd` asserts the generated LaTeX appends `:` to the bare name, leaves the colon-terminated and option-bearing names unchanged (no `::`, no trailing colon), and compiles
- [ ] `tests/unit/latexmk/parse-error.test.ts` covers the crash-log signature for the diagnostic safety net

Fixes #14553


cderv:
I understand that a mode is expected always. So I think we can try add the colon automatically but we would need Lua processing to be sure not to add to font with mode

PR #14596, opened 2026-06-16
Add a new regexp to also match uri in the form /proxy/{{port}}

### LINKED ISSUE #14595: Reload preview with VSCODE_PROXY_URI ignores the value
### I have:

### Bug description

When using `quarto preview` with the `VSCODE_PROXY_URI` the url displayed is the good one (as was done in to https://github.com/quarto-dev/quarto-cli/issues/1542), but editing the previewed file, cause a reload to the root url ignoring the value in `VSCODE_PROXY_URI`

### Steps to reproduce

In a code server environment (or with an env variable `VSCODE_PROXY_URI) `quarto create` an empty website. Enter the new directory and execute `quarto preview`.

The preview url is the good one (evaluated against `VSCODE_PROXY_URI`).

Then edit the `index.qmd` file, the broser should reload at the root of the url (without the path indicated in `VSCODE_PROXY_URI`)

### Actual behavior

_No response_

### Expected behavior

_No response_

### Your environment

_No response_

### Quarto check output

```bash
Quarto 1.9.37
      Quarto cache location: /home/onyxia/.cache/quarto
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.4.5: OK
      Typst version 0.14.2: OK
      Version: 1.9.37
      Path: /usr/local/lib/quarto-1.9.37/bin

      TinyTeX: v2026.04
      Chromium: (not installed)
      VeraPDF: (not installed)

      Using: TinyTex
      Path: /home/onyxia/.TinyTeX/bin/x86_64-linux
      Version: 2026

      Chrome:  (not detected)

      Version: 3.13.12
      Path: /opt/python/bin/python3
      Jupyter: 5.9.1
      Kernels: python3

```

micedre:
closes #14595

## Description

This PR add a pattern to match, used with the preview subcommand to reload the page

## Checklist

I have (if applicable):

- [ ] added new tests


cderv (approved):
Thanks for the fix.

The new pattern looks right for code-server — `/proxy/{port}/` and `/absproxy/{port}/` are both valid code-server proxy URL formats, so matching both with `(?:abs)?` makes sense.

I want to get sign-off from the Positron team before merging, since this area has known interactions with the VS Code extension (@cscheid flagged this in the issue). 

For reference from what I found: 
* Positron doesn't use `VSCODE_PROXY_URI` — it builds its own proxy — and its format is also `/proxy/{port}/` on desktop, so the new pattern covers it correctly. 
The Workbench path (`/p/{token}/`) was already matched by the first regex. 

No regression expected, but they should still be aware of the change.

@vezwork asking you for the Quarto VSCODE extension in case you see anything. And do you know who is best fit in positron team to ask ? I'll probably slack them.

vezwork:
I don't see anything obvious, I don't see how this could cause issues. Probably @juliasilge is the best person I know to ask.

juliasilge:
Here is my understanding of what is going with this fix:

### Posit Workbench (Positron Pro / VS Code sessions) should be unaffected

Workbench reverse-proxies sessions under a `/p/<id>/` path, which is already captured by the existing first alternative (`/^.*?\/p\/\w+\//`). This PR adds a new alternative but does not change that one, so reload behavior in Workbench is unchanged. Like @cderv noted, Positron does not read `VSCODE_PROXY_URI` at all. It detects the Workbench context via the `RS_SERVER_URL` env var and resolves preview URLs through `vscode.env.asExternalUri()`, so the `/p/<id>/` prefix is what reaches the browser, not a proxy path derived from `VSCODE_PROXY_URI`.

### Positron Desktop should be unaffected

In the desktop app, `asExternalUri()` is effectively identity, so a previewed app stays at `http://127.0.0.1:<port>/` with no `/proxy/<n>/` segment in the path. None of the three regex alternatives match, so it falls through to a plain `window.location.reload()`, same as today.

### Positron Server Web (like [in JupyterHub](https://opensource.posit.co/blog/2026-04-06_positron-server-jupyterhub/)) is newly matched, and likely a fix!

Here Positron forwards previews through its own proxy and produces external URLs of the form `.../proxy/<port>/...`. The existing two patterns did not match that shape, so reloads dropped the prefix, the same bug being reported here. The new `/^.*?\/(?:abs)?proxy\/\d+\//` alternative should now match and fix it in this environment too.

I don't know if we have the ability to test this in Positron Server on JupyterHub before merging (a pretty high lift) so I think we should go for it and look for any regressions. We _do_ have some automated testing on the Positron side for JupyterHub.

cc @testlabauto and @isabelizimm for visibility on this upcoming change to the Quarto CLI, which I believe would _fix_ problems with `quarto preview` for Positron in JupyterHub. I bet we just test `quarto render` right now, so mostly as a heads up!

cderv:
Thank you ! I will merge this and make a new 1.10 build so you can test. 

We're not close to release 1.10 so we have time to revert or adapt.

PR #14602, opened 2026-06-18
fix(accessibility): add aria-label to announce ORCID profile in author metadata HTML/Reveal.js

mcanouil:
Add aria-label (and `alt=""` for decorative image) to announce ORCID profile in author metadata HTML/Reveal.js to improve accessibility.

This was detected when working with https://github.com/mcanouil/quarto-revealjs-a11y which has a feature to flag images with missing alt-text.

~~TO-DO: check upstream for similar fix.~~
`title-fancy` seems to be pure Quarto eventhough under `pandoc` directory.


cderv (approved):
Can you add a quick test ? Just for template regression.

mcanouil:
> Can you add a quick test ? Just for template regression.

too late it seems

PR #14625, opened 2026-06-26
Add `line-numbers` option for Word (docx) output

### LINKED ISSUE #14624: Native line numbering for Word (docx) output
### Problem

Quarto has no built-in way to add line numbering to `docx` output, yet line numbers are a hard requirement for many manuscript-submission, legal, and peer-review workflows. Today the only recourse is to enable them by hand in Word after every render. The need was raised in discussion [#6772](https://github.com/orgs/quarto-dev/discussions/6772) ("Adding line numbering to PDF and DOCX") — the PDF half was answered via the `lineno` package, but the docx half was never resolved.

### Proposal

A format option to turn on continuous line numbering for `docx`:

```yaml
format:
  docx:
    line-numbers: true
```

(Option name is the maintainers' call; it could eventually unify with PDF `lineno` and Typst's native line numbering into one cross-format option — what that discussion originally asked for. This request is scoped to `docx`, where there is no path today.)

### Why this needs format-level support

Word stores line numbering as `<w:lnNumType>` inside a section's `<w:sectPr>`. The correct place for it is the document's existing main `<w:sectPr>` — the one that already carries the `headerReference`s and `pgSz` — so the running head and page size are preserved. A Lua filter/extension can't reach that section (it comes from the reference doc and is written by pandoc *after* filters run), so the most a filter can do is append a *new* section, which has no `headerReference` or `pgSz` and therefore drops the running head and resets the page size to Letter. Doing it correctly requires a format-level post-processing step on the written document.

dhagmann:
Closes #14624.

Adds a `line-numbers: true` option for `docx` that turns on continuous line numbering.

## Approach

Word stores line numbering as `<w:lnNumType>` inside a section's `<w:sectPr>`. That section is written by pandoc from the reference doc *after* filters run, so a filter can't reach it without appending a *new* section (which drops the running head and resets the page size). Instead, this registers a `docx` output post-processor (`formatExtras().postprocessors`) that injects `<w:lnNumType w:countBy="1" w:restart="continuous"/>` into the document's existing `<w:sectPr>` elements — preserving each section's header references and page size.

## Changes

- `src/format/docx/format-docx.ts`: read `line-numbers` and register the post-processor.
- `src/resources/schema/document-options.yml`: declare the `line-numbers` docx option.
- `news/changelog-1.10.md`: changelog entry.
- `tests/docs/smoke-all/2026/06/27/docx-line-numbers.qmd`: smoke test asserting `<w:lnNumType>` is present.

## Verification

Built from source and rendered to `.docx`, then to PDF via LibreOffice:

- Line numbers appear and run continuously.
- With a header-bearing reference doc, the running head **and** page number are preserved.
- An A4 reference doc renders as true A4 (the page size is not reset to Letter).

The smoke test passes (`1 passed | 0 failed`, no errors or warnings).

Marking as **draft** pending discussion of the option name / approach in #14624.

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -23,6 +23,10 @@ All changes included in 1.10:
 
 - ([#14530](https://github.com/quarto-dev/quarto-cli/pull/14530)): Add `quarto.*` Pandoc template variable namespace. `format.language` is now exposed as `$quarto.language.<key>$` in custom Pandoc templates via the defaults-file `variables:` section, with no leakage into rendered output.
 
+### `docx`
+
+- ([#14624](https://github.com/quarto-dev/quarto-cli/issues/14624)): Add `line-numbers` option to enable continuous line numbering in Word output.
+
 ### `pdf`
 
 - ([#13588](https://github.com/quarto-dev/quarto-cli/issues/13588)): Fix Lua error when rendering PDF with `reference-location: margin` and a footnote alongside a figure with `fig-cap`. (author: @mcanouil)
--- a/src/format/docx/format-docx.ts
+++ b/src/format/docx/format-docx.ts
@@ -1,15 +1,16 @@
 /*
-* format-docx.ts
-*
-* Copyright (C) 2020-2022 Posit Software, PBC
-*
-*/
+ * format-docx.ts
+ *
+ * Copyright (C) 2020-2022 Posit Software, PBC
+ */
 
 import { kFilterParams } from "../../config/constants.ts";
-import { Format } from "../../config/types.ts";
+import { Format, FormatExtras, PandocFlags } from "../../config/types.ts";
 import { mergeConfigs } from "../../core/config.ts";
 import { formatResourcePath } from "../../core/resources.ts";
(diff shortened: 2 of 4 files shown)


mcanouil:
I'm copying my reply from https://github.com/orgs/quarto-dev/discussions/14626

---

@dhagmann If you `want` line numbering, I suggest you enable it in a reference document as demonstrated in the documentation, see <https://quarto.org/docs/output-formats/ms-word-templates.html>.

1. Make sure you're in Print Layout view (View tab > Print Layout).
2. Go to the Layout tab.
3. Click Line Numbers in the Page Setup group.
4. Choose Continuous for numbering throughout the whole document, Restart Each Page to start at 1 on every page, or Restart Each Section to start at 1 after each section break.

Then in your Quarto document:

```qmd
---
format:
  docx:
    reference-doc: custom-reference-doc.docx
---

{{< lipsum 3 >}}
```

Render it via `quarto render document.qmd` and you'll have line numbering. You can reuse the reference document as you see fit.

cscheid:
Thanks for the suggestion, but I think we don't want to support that maintenance given the alternative that exists in @mcanouil's comment.

Issue #6411, opened 2023-08-02
Wrong title in revealjs when including a code block


m-guggenmos:
### Bug description

If the following two conditions are met, the title of a revealjs presentation is erroneously set to the heading of the first slide:

1. Using `_quarto.yml` to define the title
2. Including a block of code in one of the slides

### Steps to reproduce

_quarto.yml
```
project:
  output-dir: _slides
  render:
    - test.qmd

title: "My title"
author: "John Doe"

format: revealjs
```

test.qmd
~~~
## First slide heading

```{python}

```
~~~

### Expected behavior

The title on the title slide is "My title".

### Actual behavior

The title on the title slide is "First slide heading".

### Your environment

- VSCode version 1.80.1
- Quarto Extension v1.90.0
- OS: Ubuntu Linux

### Quarto check output

~~~
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.1.6: OK
      Dart Sass version 1.55.0: OK
      Deno version 1.33.4: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.4.251
      Path: /opt/quarto/bin

[✓] Checking basic markdown render....OK

[✓] Checking Python 3 installation....OK
      Version: 3.10.12
      Path: /usr/bin/python3
      Jupyter: 5.3.1
      Kernels: python3

[✓] Checking Jupyter engine render....OK

[✓] Checking R installation...........OK
      Version: 4.2.3
      Path: /usr/lib/R
      LibPaths:
        - ~/R
        - /usr/local/lib/R/site-library
        - /usr/lib/R/site-library
        - /usr/lib/R/library
      knitr: 1.42
      rmarkdown: 2.20

[✓] Checking Knitr engine render......OK
~~~

cderv:
Thanks for the report. I can reproduce in a specific situation. 

* It happens with Jupyter engine and not knitr engine 
* It is not link to revealjs and it will also happen with `format: html`

* `_quarto.yml`
````yaml
title: "My title"
author: "John Doe"

format: html
````

* `index.qmd`
````
## First slide heading

Some content

```{python}
1 + 1
```
````

`title` is correctly identified in the metadata
````
Starting python3 kernel...Done

Executing 'index.ipynb'
  Cell 1/1: ''...Done

pandoc 
  to: html
  output-file: index.html
  standalone: true
  section-divs: true
  html-math-method: mathjax
  wrap: none
[...]

PR #14639, opened 2026-06-30
fix(embed): don't use a non-leading heading as embedded notebook title

### LINKED ISSUE #14577: Python comments wrongly rendered as title in the sidebar of embeded notebooks
### I have:

- [ ✔] searched the issue tracker for similar issues
- [ ✔] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [ ✔] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

I'm embedding a whole `.ipynb` file into a `.qmd` file by using something like 
```qmd
{{< embed 20260527-1d-phonon/Shockley1D-parity.ipynb echo=true >}}
```
to include non-folding source code blocks, with `code-fold: false` in `_quarto.yml` in my website project.

After rendering, the notebook sidebar and the "Source" button actually display an inline Python comment (that starts with a `#`) as the title of the notebook.

The notebook itself doesn't contains any proper markdown blocks containing titles starting with `#`.

Image

Image

### Steps to reproduce

qmd file:
````qmd
---
title: "title"
author: "Guest_1013"
date: 2026-05-27
categories: ["paper", "physics", "typst", "solid state physics"]
---

Check [here](/blog/posts/20260527-1d-phonon.pdf) for whole text.

# Example Calculation: Shockley1D-parity.ipynb

{{< embed 20260527-1d-phonon/Shockley1D-parity.ipynb echo=true >}}
````

notebook file: 
```ipynb
{
 "cells": [
  {
   "cell_type": "markdown",
   "id": "b189ab7b",
   "metadata": {},
   "source": [
    "Lowest frequency standing wave solution"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": null,
   "id": "12c23ca4",
   "metadata": {},
   "outputs": [],
   "source": [
    "n=np.argmin(values_sorted)\n",
    "print(n, values_sorted[n])\n",
    "plt.plot(vectors_sorted[n])\n",
    "plt.title('Lowest Frequency Standing Wave Solution')\n",
    "plt.xlabel(r'$n$')\n",
    "plt.ylabel(r'$u$')\n",
    "\n",
    "# plt.savefig('./fig/Shockley-parity/lowest_solution.svg')"
   ]
  }
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "base",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.13.13"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 5
}

```

### Actual behavior

Embeded `ipynb`s' inline comments are mistaken as titles when there's no other proper markdown titles, displayed in the embeding page.

Image

Image

### Expected behavior

Show the file name rather than the comments.

### Your environment

- IDE: VSCode 1.122.1
- Quarto: 1.9.37
- Python: Python 3.13.13 | packaged by conda-forge | (main, Apr  8 2026, 01:56:49) [MSC v.1944 64 bit (AMD64)] on win32

### Quarto check output

```bash
Quarto 1.9.37
[>] Checking environment information...
      Quarto cache location: C:\Users\26653\AppData\Local\quarto
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.4.5: OK
      Typst version 0.14.2: OK
      Version: 1.9.37
      Path: D:\Quarto\bin
      CodePage: 936

      TinyTeX: (external install)
      Chromium: (not installed)
      VeraPDF: (not installed)

      Using: Installation From Path
      Path: D:\TinyTeX\bin\windows
      Version: 2026

[>] Checking Chrome Headless....................OK
      Using: Chrome found on system
      Path: C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe
      Source: Windows Registry

[>] Checking R installation...........(None)

      Version: 3.13.13 (Conda)
      Path: D:/miniforge3/python.exe
      Jupyter: 5.9.1
      Kernels: python3

[>] Checking Julia installation...
```

cderv:
## TL;DR

An embedded notebook with no title and no heading cell used a code-cell comment (e.g. `# plt.savefig(...)`) as its "Source" title. This PR gates heading-to-title promotion in `findTitle` on "nothing precedes the heading" (**option A**), so a `#` inside a code fence can't be the title and the filename is used instead. A side effect: a real markdown heading with prose before it also falls back to the filename — intentional, because a standalone notebook's own title already behaves that way. A more surgical fence-aware fix (**option B**) is described below but not taken here.

---

When embedding a notebook that has no YAML title and no markdown heading cell, the "Source" link and sidebar showed a Python code comment (e.g. `# plt.savefig(...)`) instead of the notebook filename.

## Root Cause

`findTitle()` in `src/core/jupyter/jupyter-embed.ts` scans every cell's markdown to find a title. A code cell's markdown is its fenced source, and `partitionMarkdown` is not fence-aware, so a `#` comment line inside the fence matches the ATX-heading regex and gets returned as the title.

The non-embed path already guards against exactly this. The Dec-2023 "Improve title snipping in notebooks" change (24672a4af, #5363/#6411) made `fixupFrontMatter` only promote a heading to the notebook title when no content precedes it. `findTitle` predates that work and was never brought in line, so it still promoted any heading-like line.

## Fix (option A)

Expose `contentBeforeHeading` (already computed by `markdownWithExtractedHeading`) through `PartitionedMarkdown`, and gate `findTitle` on `!partitioned.contentBeforeHeading`. A fenced code block's opening delimiter always counts as content before the comment, so the comment is no longer mistaken for a title and the filename is used instead.

This borrows the same content-before-heading gate that `fixupFrontMatter` uses. The two functions are not otherwise equivalent — `findTitle` scans every cell and can't filter on `cell_type`, whereas `fixupFrontMatter` inspects only the first markdown cell.

## Behavior note

Because the gate keys off "content before the heading" rather than "is this a code cell", it also affects a real markdown cell whose heading has prose before it: that heading is no longer promoted, and the embed "Source" title falls back to the filename.

This is intentional and is the reason to prefer option A. A standalone notebook with prose before its first heading already gets **no** title from that heading — `fixupFrontMatter` only promotes a heading when nothing precedes it (since #5363 / #6411). Option A makes the embed "Source" title behave the same way as the notebook's own title, instead of using a heading that the rest of Quarto would not treat as the document title.

The documented precedence is YAML `title`, then the first markdown heading, then the filename. Option A is marginally stricter than the literal "first markdown heading" wording for the prose-before-heading case, but matches the actual title behavior of a standalone notebook.

## Alternative (option B, not taken here)

The root cause mcanouil identified is that heading extraction is not fence-aware. Option B would make `markdownWithExtractedHeading` (or `findTitle`) skip `#` lines inside fenced code, fixing only the code-comment case and leaving a prose-before-heading markdown cell promotable — matching the literal documented "first markdown heading" wording.

Option B is the more surgical fix for the reported bug, but it touches every caller of `markdownWithExtractedHeading` and would leave the embed title behaving differently from the standalone notebook title for the prose-before-heading case. This PR takes option A to keep the two title code paths consistent; option B remains available if we later prefer to track the documented wording exactly.

## Tests

Two smoke-all tests under `tests/docs/smoke-all/2026/06/30/`:

- `14577.qmd` — the reported case: an embed of a notebook whose only cell is a `#`-comment code cell, asserting the Source link uses the filename and never the comment.
- `embed-heading-after-content.qmd` — the general alignment: an embed of a notebook whose only heading has content before it, asserting the Source link falls back to the filename rather than the heading.

Two incidental test-infra cleanups came out of writing that test:

- `render-embed.test.ts` teardowns now use `safeRemoveSync` instead of raw `Deno.removeSync`, so a teardown no longer throws (and aborts, leaving stale state) when a preview artifact is already gone.
- smoke-all's `postRenderCleanup` now removes recursively, so a registered cleanup entry can be a `*_files` support directory and not only a single file.

Fixes #14577


cderv:
I am putting that as a draft, because maybe option B is better... 

I need to think about it.

cderv:
This is replaced by #14643

PR #14664, opened 2026-07-09
Update Pandoc, Typst, Dart Sass, esbuild, and veraPDF

cderv:
Version bump only in `configuration` and `src/command/check/check.ts`, so the new binaries can be archived to S3 and downloaded during `configure`. Bumps Pandoc to 3.10, Typst to 0.15.1, Dart Sass to 1.101.0, and esbuild to 0.28.1, all riding along in one branch since Carlos is archiving from it directly. veraPDF stays at 1.28.2 for now (1.30.2 isn't published to S3 yet).

> [!NOTE]
> Draft: only Pandoc and Typst have had any risk analysis so far (template resync, Lua API compatibility, test coverage). Dart Sass and esbuild are bumped here but not yet verified. None of this is ready to merge until that work lands in follow-up commits.


cderv:
Typst 0.15 seems to have some impact. Already two issues open with Pandoc
- https://github.com/jgm/pandoc/issues/11761
- https://github.com/jgm/texmath/issues/291

I'll see how to handle. 

I noticed a few things to investigate in our failing tests too. Slowly looking in all this

cderv:
I dug into jgm/pandoc#11761 (leading-dot backslash escaping) to see whether we can work around it on our side for v1.10, and traced the full chain through our own code, not just pandoc's.

## Root cause

This only triggers when a path-like metadata value starts with a literal `.` — `bibliography: ./refs.bib` or `../refs.bib`, not `bibliography: refs.bib` or `sub/refs.bib`. I confirmed the trigger condition and the full chain end to end:

1. `bibliography: ./refs.bib` in the YAML front matter is metadata pandoc parses as markdown inlines, not a raw string.
2. We don't strip or rewrite that value for Typst — `isBibliographyOutput()` only covers `bibtex`/`biblatex`/`csljson`, so for `typst` the `bibliography` key flows through untouched:
https://github.com/quarto-dev/quarto-cli/blob/d3a4452c54a9f41fb040cddc25490350bde575bf/src/resources/pandoc/datadir/_format.lua#L103-L110
3. Our own Typst partial interpolates it straight into a quoted string argument, exactly the pattern the pandoc issue describes:
https://github.com/quarto-dev/quarto-cli/blob/d3a4452c54a9f41fb040cddc25490350bde575bf/src/resources/formats/typst/pandoc/quarto/biblio.typ#L16
4. Pandoc's `metaToContext` renders that value through `inlinesToTypst`/`escapeTypst`, which since 80c704a9a (jgm/pandoc#11511) prepends a backslash to any string starting with `.`. Typst 0.14 silently tolerated the stray backslash; Typst 0.15 hard-rejects it (`path must not contain a backslash`).

I reproduced this locally with pandoc 3.10 + Typst 0.15.0: `bibliography: ./refs.bib` fails, `bibliography: refs.bib` renders fine.

## jgm's reply

> This is a tricky one. Pandoc renders everything in metadata as if it's ordinary text, but this poses a problem in the current case, since typst's escaping rules inside a string or path context are different than for ordinary text. Probably the most straightforward fix would be to roll back 80c704a9acb01fbeffc91a0d05a86c98ab0fe2c0 and replace it with something more subtle, but the replacement isn't obvious, and I'm not sure there wouldn't be other similar problems.

So no fix on the pandoc side yet, and no clear timeline. His workaround is to write the bibliography path as raw Typst so it skips pandoc's metadata escaping entirely:
```yaml
bibliography: `./refs.bib`{=typst}
```

I tested this end to end rather than just relaying it: rendered with a single bibliography this way, and separately with a list (one entry as ``- `./refs.bib`{=typst}``, alongside a second plain entry). Both compiled cleanly and, more importantly, the citation and bibliography entry actually resolved and rendered correctly in the PDF — Quarto disables pandoc's own `--citeproc` for Typst and lets Typst resolve citations natively from the `#bibliography()` call, so the raw-syntax value reaches Typst as a plain path string with no knock-on effect on citation resolution. This workaround is safe to document for users hitting this now, at least for single-format Typst output (as jgm notes, multi-format documents would need a Lua filter to vary the value by output format).

## Possible workaround on our side

We could apply the same idea automatically instead of asking users to hand-write raw syntax. Pandoc's Typst writer emits `RawInline (Format "typst")` content completely verbatim, bypassing `escapeTypst`:
[source](https://github.com/jgm/pandoc/blob/5dd191098d45dfe7ba040f37c45dcb0e98d4cac6/src/Text/Pandoc/Writers/Typst.hs#L488-L491)
```haskell
RawInline fmt str ->
  case fmt of
    Format "typst" -> return $ literal str
    _ -> return mempty
```
So a small Lua filter gated on `isTypstOutput()` could rewrite each string in `doc.meta.bibliography` (it's normalized to a list internally, even for a single entry) into a `RawInline("typst", path)`, sidestepping the escaping bug entirely rather than post-processing the generated `.typ` (which is more fragile, per the `mediabag.lua`-style approach we already ruled out for the Windows path-separator issue). Haven't implemented this yet — flagging the direction here before committing to it.

## This isn't just bibliography

The same shape of bug — a Quarto Typst partial quoting a raw `$metadata-var$` — exists at least one more place:
https://github.com/quarto-dev/quarto-cli/blob/d3a4452c54a9f41fb040cddc25490350bde575bf/src/resources/formats/typst/pandoc/quarto/page.typ#L14

`$logo.path$` for brand logos. I tested a brand logo with an explicit leading `./` and confirmed it does not currently trigger this — our own path-resolution normalizes the leading `./` away before the value reaches the template. That's incidental, not a guard against this specific pandoc behavior. Any future Typst partial we write that quotes a metadata value directly is exposed to the same class of bug if that value can ever start with a literal `.`.

Not blocking the version bump investigation, but wanted to get the full picture written down before deciding how much of this to fix for v1.10 vs. just document as a known limitation.

cderv:
We are also hitting again the 200 variables limits, so I'll make probably another PR to reduce the usage.

cderv:
Update on the 200-local-variable limit: it hit us directly on this branch rather than needing a separate PR. The packaged filter bundle concatenates every `import()`ed file into one physical `main.lua` with no per-file scoping, so all their top-level locals share Lua's hard 200-per-function limit. That budget was already at 199/200 before this branch; the Windows path-normalization fix added four `local path = require("modules/path")` call sites, tipping it to 203 and breaking `test-bundle` at compile time.

Fixed by registering `path` in the existing `_quarto.modules` registry (a single field in an already-existing global table, costs zero locals) instead of each site re-requiring the module as its own local:
https://github.com/quarto-dev/quarto-cli/blob/36d9f32b68ab275e1262d2220bd1d6e08bd9271b/src/resources/filters/modules/import_all.lua#L19

Confirmed by rebuilding the inlined bundle standalone and compiling it with pandoc directly (199 locals, restored to the pre-branch margin) — `test-bundle` is green again on the latest push.

This isn't the first time this class of bug has bitten us (found two more independent recurrences on other open PRs going back to 2026-03), and the 200-local budget is structurally fragile as filter code accumulates — wrote up the mechanism, why it recurs, and the ranked fix options (including a `do...end`-wrapping fix at the inliner level that would remove the shared-budget problem entirely):
https://github.com/quarto-dev/quarto-cli/blob/36d9f32b68ab275e1262d2220bd1d6e08bd9271b/llm-docs/lua-filter-bundle-local-limit.md

Not fixing the inliner itself in this PR — keeping this branch scoped to unblocking CI, structural fix and the broader `_quarto.modules` cleanup (about 25 other pre-existing sites redundantly re-requiring already-registered modules) are follow-up work.

Separately, the `typst-poster` extension test was failing on this branch too, and it turned out to be two of the same-shaped upstream bugs, both in `quarto-ext/typst-templates`' own example content rather than our code:

1. The demo formula `\sum_(k=1)^n k` uses paren-grouping instead of proper LaTeX brace-grouping. jgm/texmath#291 confirmed pandoc 3.10's texmath conversion fuses a superscript immediately followed by a space into one Typst identifier (`^n k` → `^nk`), which Typst then rejects as an unknown variable. jgm noted the paren form is "a bit weird" and not proper LaTeX.
2. `institution-logo: ./images/ncstate.png` starts with a literal `.`, hitting the same leading-dot escaping bug from jgm/pandoc#11761 — this time in a template's own logo path rather than bibliography, which is exactly the "isn't just bibliography" concern flagged above. Another data point that any Typst partial (ours or an extension's) quoting a raw metadata value is exposed if that value can start with `.`.

Both are upstream pandoc/texmath bugs with no fix yet, so fixed at the content level instead — proper brace-grouping for the formula, dropped the redundant leading `./` for the logo path:
https://github.com/quarto-ext/typst-templates/commit/79fcc3a4cd1b59543d962367c6d6d257c7beab56

Merged upstream, CI on this branch reran and is now picking up the fix.

cderv:
The Windows feature-format-matrix job failure came from pandoc 3.10 always emitting a newline after `#set text(...)` in the Typst writer (jgm/pandoc#11583) — a fix for lists getting mis-parsed as the first block of a lang/style-attributed Div. The fix applies unconditionally, so it also touches the simple inline cases `dev-docs/feature-format-matrix` asserts on, even though the newline is a no-op there (Typst collapses it to whitespace in markup mode).

I confirmed this is pure test drift, not a Quarto regression: pdftotext on the compiled PDFs shows no visual difference, and reproducing the upstream bug scenario locally confirms the fix behaves as intended (list renders correctly where it previously wouldn't have). Loosened the 38 affected regexes to tolerate the whitespace change.

One related but distinct case: `raw-blocks/interpreted/pandoc-json/document.qmd` now escapes consecutive hyphens individually instead of leaving them literal — verified via pdftotext that the compiled PDF still renders `--`, not an en dash.

Separately, while digging into why only the Windows job caught this: `run-tests.sh` passes its final test-args variable unquoted to `deno test`, so bash pathname-expands the `**` glob itself with no `globstar` set, collapsing `qmd-files/**/*.qmd` to single-`*` semantics and silently dropping every file nested more than one directory deep. Ubuntu's job only ran 10 of the 271 tests in this bucket. That's unrelated to the pandoc bump and out of scope here — tracking it separately.

Issue #14667, opened 2026-07-10
quarto-nav.js's index.html clean-URL rewrite mangles alternate-format links on index pages


jph00:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest) (the live example below is built with 1.9.38; see environment notes)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

`quarto-nav.js` rewrites every link on page load to clean up `index.html` URLs, stashing the original in `data-original-href`:

```js
links[i].dataset.originalHref = links[i].href;
links[i].href = links[i].href.replace(/\/index\.html/, "/");
```

The regex is unanchored, so on a site's index page the "Other Formats" link to `index.html.md` (the markdown twin generated when a site pairs a markdown format with html, e.g. for llms.txt workflows) is rewritten from `https://example.com/index.html.md` to `https://example.com/.md`, which 404s. Only the index page is affected, since no other page has `index.html` as a filename prefix; the same CommonMark link works everywhere else on the site. The served HTML is correct; only the client-side rewrite breaks it.

Still present in `main` today, in [`src/resources/projects/website/navigation/quarto-nav.js`](https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/projects/website/navigation/quarto-nav.js).

The `index.html.md` naming comes from `output-file: index.html` in the page metadata (tools like nbdev set this on every page of the sites they generate).

### Steps to reproduce

Primary evidence is the live example under "Actual behavior" plus the code above; a minimal project for reproducing:

`_quarto.yml`:

```yaml
project:
  type: website

format:
  html: default
  commonmark: default
```

`index.qmd` (the `output-file` line gives the alternate format the `index.html.md` name that triggers the bug):

````qmd
---
title: Home
output-file: index.html
---

Hello.
````

Then:

```bash
quarto render
python3 -m http.server -d _site   # the rewrite is skipped on file:, so serve over http
```

Open the served index page and inspect the CommonMark link under "Other Formats". Note: on 1.8.25 I could confirm the rendered HTML gets the `index.html.md` link, but not the client-side rewrite itself, because that version routes `site_libs` and the `.md` twin outside `_site` when `output-file` is set, so the served page loads no JS; the rewrite is confirmed on the 1.9.38-built live site below.

### Actual behavior

[...]

PR #14671, opened 2026-07-11
Fix keep-md cleanup deleting the markdown output when `output-file` has an `.html` extension

### LINKED ISSUE #14669: `output-file` with `.html` extension loses the markdown-format output in multi-format project renders (website render fails at move step)
### Bug description

In a project render that pairs `format: html` with a markdown format (e.g. `commonmark`, as in llms.txt workflows), setting `output-file` to a name **with an `.html` extension** (e.g. `output-file: index.html`, which nbdev sets on every page it generates) breaks the render of the markdown twin:

- **Website project**: `quarto render` fails with a hard error at the output-move step — it tries to move `<name>.html.md` from the project root into `_site`, but the file was never written. The already-rendered `index.html` and `site_libs/` are left stranded in the project root.
- **Default project**: `quarto render` exits 0, but the markdown output is **silently missing** — no `.md` file is produced anywhere. Presumably the same underlying bug with a quieter symptom.

Notes from narrowing this down:

- `output-file: index` (no extension) works correctly in both project types (`_site/index.html` + `_site/index.md`).
- Not index-specific: `output-file: other.html` on a non-index page fails identically.
- Rendering a single file/format works: `quarto render index.qmd --to commonmark` correctly produces `_site/index.html.md` (so the collision-avoiding `<name>.html.md` naming itself works; it's the multi-format project render that loses the file).
- Reproduces identically with released **1.9.38** and current `main` dev build (8577ed1c6).
- Related history: the reporter of #14667 observed on 1.8.25 that this config routed `site_libs` and the `.md` twin outside `_site`, so some form of this misbehavior is longstanding. (Found while reproducing #14667, which needs this config to exercise its client-side bug end-to-end.)

### Steps to reproduce

`_quarto.yml`:

```yaml
project:
  type: website

format:
  html: default
  commonmark: default
```

`index.qmd`:

```qmd
---
title: Home
output-file: index.html
---

Hello.
```

Then:

```bash
quarto render
```

### Actual behavior

```
ERROR: NotFound: No such file or directory (os error 2): rename '<project>/index.html.md' -> '<project>/_site/index.html.md'

Stack trace:
    at Object.renameSync (ext:deno_fs/30_fs.js:267:3)
    at safeMoveSync (file:///.../src/deno_ral/fs.ts:102:10)
    at renderProject (file:///.../src/command/render/project.ts:526:9)
```

After the failed render, `index.html` and `site_libs/` sit in the project root instead of `_site`, and no `index.html.md` exists anywhere (watching the directory during the render confirms it is never created).

With `project: type: default` instead: exit code 0, `index.html` is produced, and the commonmark output silently does not exist.

### Expected behavior

The render succeeds and produces `_site/index.html` and `_site/index.html.md`, matching what the single-format render (`quarto render index.qmd --to commonmark`) already produces.

### Your environment

- macOS 26.x (Darwin 25.5.0), arm64
- Reproduced with Quarto 1.9.38 (release) and quarto-cli `main` @ 8577ed1c6 (dev build)

### Quarto check output

Dev build at 8577ed1c6; `quarto check` unremarkable (Pandoc 3.8.3, Deno 2.4.5).

cscheid:
Closes #14669.

## Problem

Pairing an html format with a markdown format while `output-file` has an `.html` extension (e.g. `output-file: index.html` with `format: html` + `commonmark`, as nbdev sets up for llms.txt workflows) names the markdown output `index.html.md` — exactly the path the keep-md intermediate convention (`<input-stem>.<base-format>.md`, `executionEngineKeepMd`) assigns to the html format. The html render's cleanup removed what it took for a stale intermediate, deleting the freshly written markdown output:

- **website projects**: the render then failed at the output-move step (`NotFound … rename '<proj>/index.html.md' -> '<proj>/_site/index.html.md'`), leaving `index.html` and `site_libs/` stranded in the project root
- **default-type projects**: exit 0, markdown output silently missing
- **`keep-md: true`**: no crash, but the html render's intermediate silently fought with the markdown output for the same file

Diagnosis was confirmed by instrumenting `renderCleanup` and watching the filesystem at 2ms resolution: the commonmark output is written and deleted 29ms later by the html format's cleanup.

## Fix

Principle: keep-md intermediate handling must never write to or delete a path that a declared format owns.

**Commit 1 (core fix):** compute the projected output paths of the formats being rendered (`projectedOutputFile`, next to `formatOutputFile` in `core/render.ts`), attach them to each `RenderContext`, and

- skip the keep-md deletion in `renderCleanup` when the conventional keep-md path is another format's output
- skip the keep-md intermediate **write** on the same collision, with a warning (the real output wins; warn rather than error for now, may revisit)

**Commit 2 (partial-render protection, split out for review):** a partial render (`quarto render --to html`) must not delete the markdown twin a previous full render produced (outputs live in place in default-type projects). Since `renderContexts` already resolves every declared format (inactive ones are marked `active: false`), the implementation is just dropping the active-only filter from commit 1's computation — one line plus a test.

## Tests

New smoke tests in `tests/smoke/render/render-output-file-collision.test.ts` (fixtures under `tests/docs/output-file-collision/`):

1. website project renders both outputs into `_site`, no root debris
2. default-type project keeps the markdown output
3. `--to html` partial render preserves an existing twin (commit 2)
4. `keep-md: true` warns, and the twin holds the commonmark output rather than the intermediate

Each commit's tests verified to fail without that commit's change. Genuine stale-intermediate cleanup still works (`keep-md: true` render followed by `keep-md: false` removes the leftover). Together with #14670, the full nbdev/llms.txt scenario from #14667 now works end-to-end (verified on a local merge of both branches: render succeeds and the served page's `index.html.md` link resolves with HTTP 200).


cscheid:
(Claude says:)

**Note on the 8 failing windows checks** (analysis after re-running them — they fail identically across 3 attempts):

- The failing tests are unrelated to this PR: babel/TinyTeX package-auto-install docs (`13633.qmd`, `latex-hyphen-lang-{zh,es}.qmd`), `caption-footnotes`, the feature-format matrix, and one search-index test. The new `render-output-file-collision` tests **pass on both platforms**, and every ubuntu job is green.
- Root symptom: TeX Live's `fmtutil.pl`/`tlmgr` aborts with `kpsewhich -var-value=TEXMFROOT failed` / `'kpsewhich' is not recognized`. The PATH it reports has its first two entries joined by `:` (the unix separator) instead of `;`, merging both TinyTeX bin dirs into one invalid entry — this happens inside TeX Live's perl, far below anything this PR touches (keep-md path bookkeeping).
- Why this PR and not #14670 (same runner image `windows-2025-vs2026 20260628.158.1`, same fresh TinyTeX v2026.07 download, green 36 minutes earlier): this PR adds a smoke **test file**, which reshuffles the deterministic test-bucket sharding; #14670 added only playwright files, which don't. A latent test-ordering-dependent TinyTeX corruption (e.g. bucket 10 runs `render-pdf-svg-conversion` / `render-verapdf` before the babel doc) would fire deterministically on exactly this PR's bucket layout — matching the 3-for-3 rerun failures while `main` and #14670 stay green.

I couldn't bisect the specific polluting test without a windows box; happy to file the CI investigation as a separate issue if this diagnosis looks right to maintainers.

cderv (approved):

cderv:
LGTM.

Traced all four paths (default / website / `--to html` / keep-md): the collision check is source-relative on both sides (`keepMd` and `projectedOutputFile`), so it matches before the output-dir move, and `pathsEqual` keeps it separator-safe on Windows. Tests pass locally.

Two changes on top of the fix:

- Pulled the duplicated collision check — it was inline in both `renderCleanup` and the keep-md write path — into `keepMdCollidesWithFormatOutput` in `core/render.ts`, so the "never touch a path another format owns" invariant has one home instead of two copies that can drift.
- Trimmed the changelog entry to the user-facing symptom. The keep-md / output-move mechanism and the nbdev+commonmark example read as internals for a changelog line — that context is already in the PR body.

One thing I noticed while in here: book and manuscript renders call `renderCleanup` on their own paths and never populate `siblingFormatOutputs`, so an html+markdown `output-file` collision there would still hit the original bug. Out of scope for the reporter's case and rare. So I don't plan to follow up on this

PR #14677, opened 2026-07-13
Vendor axe-core, drop the Skypack CDN dependency

cwickham:
## Description

Vendors axe-core as a bundled HTML resource, so the `axe:` accessibility option stops loading it from the Skypack CDN in the reader's browser.

**What changes**

- `configuration` pins `AXE_JS=4.10.3` — exactly the version the CDN import already pinned, so this PR changes *where* axe comes from, not *which* axe runs. Scan results are identical. A follow-up PR will bump to current axe-core (4.12.1), keeping "different source" and "different results" in separate diffs.
- `update-html-dependencies.ts` downloads `axe.min.js` plus axe-core's `LICENSE` (MPL-2.0) and `LICENSE-3RD-PARTY.txt` from the npm package via unpkg — `https://unpkg.com/axe-core@4.10.3/<file>`, the same `updateUnpkgDependency` helper used for anchor-js, popper, and tippy — into `src/resources/formats/html/axe/`. To verify the committed bytes, `curl` those three URLs and compare (or rerun `quarto-bld update-html-dependencies`). The files are vendored byte-for-byte: the `/*!` banner in `axe.min.js` carries the Deque copyright/MPL notice that must ship with every copy. Per-directory LICENSE files follow the existing convention (`src/resources/formats/pdf/pdfjs/LICENSE`).
- `format-html-axe.ts` injects `axe.min.js` as a classic script ahead of the `axe-check.js` module, so `window.axe` is set before the checker runs.
- `axe-check.js` replaces the pinned Skypack dynamic import with `window.axe`.

**Why**

- Accessibility checking now works offline and no longer depends on Skypack's uptime (previously, no network meant the checker silently never ran).
- Viewing a rendered document no longer triggers a request to `cdn.skypack.dev` from every reader's browser.
- The upcoming `quarto axe` scanner and the overlay will provably run the same axe version from the same file.

**A race this exposed (also fixed here):** `quarto-dashboard.js` keeps `<html>` hidden until its `DOMContentLoaded` handler runs, but deferred modules execute before that event — so with the Skypack network latency gone, the first scan ran against a fully hidden document and dashboards reported no violations. `axe-check.js` now waits for the window `load` event before the first scan. The race predates this PR (a sufficiently fast CDN response could have triggered it); vendoring just made it deterministic.

**Size:** ~573 KB committed across three files (541 KB of it `axe.min.js`; ~145 KB gzipped on the wire). Precedent: pdf.js, MathJax, and reveal.js are vendored the same way.

[...]

PR #14680, opened 2026-07-13
Axe overlay polish: left-align report and scroll highlighted elements clear of it

cwickham:
## Description

> [!NOTE]
> Was stacked on #14677 (vendored axe-core); that PR merged, and this branch is rebased onto `main`.

Three quality-of-life fixes to the `axe: output: document` fixed overlay in HTML output.

**Left-align the report overlay.** The overlay is appended into `<main>`, so page-level centering leaks into the report text — e.g. an about page with `about: template: jolla` ships `div.quarto-about-jolla main { text-align: center }` and centers the whole report. The reveal report slide already pins `text-align: left`; the base overlay rule now does the same.

**Don't scroll highlighted elements underneath the overlay.** Hovering a selector in the report scrolls the target element into view with `scrollIntoView({block: "center"})`, but the overlay is `position: fixed; bottom: 3rem; right: 3rem; max-height: 50vh` — a centered element (especially a full-width one) could land partly or wholly underneath it. The scroll target is now computed by a new exported pure helper, `overlayAwareScrollTop`, which centers the element in the viewport band *above* the overlay. Elements with no horizontal overlap keep the default centering, and only the plain-HTML fixed overlay takes this path: reveal navigates slides (the report is its own slide) and the dashboard offcanvas is a static full-height panel that vertical scrolling can't dodge. The overlay itself never moves or fades, so the mouse stays inside it during the scroll and no `mouseleave` → unhighlight flicker can fire.

**Make the overlay itself pass its own scan.** Running axe-core against the injected report (it never audits itself — it's added after the scan) flagged `scrollable-region-focusable` (serious, WCAG 2.1.1): once the report exceeds `max-height: 50vh` it scrolls, but wasn't keyboard-focusable. The overlay is now a focusable, labeled region (`tabindex="0"`, `role="region"`, `aria-label`), added only in the fixed-overlay path — the reveal slide and dashboard offcanvas don't scroll the report element itself, so they skip the extra tab stop. The violation-target spans remain mouse-only; full keyboard operability of the report is a separate issue.

Tests: unit tests for the helper's geometry (`tests/unit/axe-overlay-scroll.test.ts`), a Playwright test that hovers a violation near the bottom of a long page and asserts the highlighted element settles fully above the overlay (verified to fail against the old scrolling behavior), and assertions on the existing HTML overlay test for `text-align: left` and the focusable-region attributes. A self-scan regression test runs the vendored axe-core build (`window.axe`) against the injected overlay on the long-report page and asserts zero violations — the overlay is added after the page scan, so nothing else ever audits it (verified this test fails with `scrollable-region-focusable` when the fix is removed).

## Checklist

I have (if applicable):

- [x] added new tests

<details>

</details>

### DIFF
--- a/src/resources/formats/html/axe/axe-check.js
+++ b/src/resources/formats/html/axe/axe-check.js
@@ -67,15 +67,6 @@ export function overlayAwareScrollTop(
   return Math.max(0, scrollY + elementRect.top - offset);
 }
 
-// Which of a violation node's target selectors a single hover/click should act on.
-// A node.target with several selectors is one element's cross-frame selector chain,
-// not distinct elements, so a hover must act on the primary (first) target only —
-// otherwise one hover fires one competing window.scrollTo per selector (last-wins
-// jank). Pure and DOM-free, like overlayAwareScrollTop, so it is unit-testable.
-export function nodeScrollTargets(node) {
-  return node.target.slice(0, 1);
-}
-
 class QuartoAxeReporter {
   constructor(axeResult, options) {
     this.axeResult = axeResult;
@@ -185,16 +176,11 @@ class QuartoAxeDocumentReporter extends QuartoAxeReporter {
     for (const node of violation.nodes) {
       const nodeElement = document.createElement("div");
       nodeElement.className = "quarto-axe-violation-selector";
-      // Render a span for every selector so the full cross-frame chain is shown.
       for (const target of node.target) {
         const targetElement = document.createElement("span");
         targetElement.className = "quarto-axe-violation-target";
         targetElement.innerText = target;
         nodeElement.appendChild(targetElement);
-      }
-      // Register hover/click on the node's scroll target(s) only — see
-      // nodeScrollTargets — so one hover fires one scroll, not one per selector.
-      for (const target of nodeScrollTargets(node)) {
         if (typeof Reveal !== "undefined") {
           // RevealJS: click navigates to the slide and highlights briefly
           nodeElement.addEventListener("click", () => {
--- a/src/resources/formats/html/axe/axe-check.js
+++ b/src/resources/formats/html/axe/axe-check.js
@@ -67,6 +67,15 @@ export function overlayAwareScrollTop(
   return Math.max(0, scrollY + elementRect.top - offset);
 }
 
+// Which of a violation node's target selectors a single hover/click should act on.
+// A node.target with several selectors is one element's cross-frame selector chain,
+// not distinct elements, so a hover must act on the primary (first) target only —
+// otherwise one hover fires one competing window.scrollTo per selector (last-wins
+// jank). Pure and DOM-free, like overlayAwareScrollTop, so it is unit-testable.
+export function nodeScrollTargets(node) {
+  return node.target.slice(0, 1);
+}
+
 class QuartoAxeReporter {
   constructor(axeResult, options) {
     this.axeResult = axeResult;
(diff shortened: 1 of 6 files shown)


cwickham:
Stacked on #14677 — please don't merge this until #14677 has landed. The base branch is set to `axe/vendor-axe-core`, so GitHub will retarget this PR to `main` automatically once #14677 merges; it should then show only the overlay-polish commits.

src/resources/formats/html/axe/axe-check.js
```diff
@@ -35,6 +35,38 @@ export function axeConformanceLevel(tags) {
...
+
+// Compute the document scrollTop that positions an element in the viewport band
+// the fixed report overlay does NOT cover, so scrolling a highlighted element
+// into view can't park it underneath the report. The overlay is bottom-anchored,
+// so that band runs from the viewport top down to the overlay's top edge.
+// Returns null when default `block: "center"` scrolling can't be obscured (no
+// horizontal overlap with the overlay) or when the overlay leaves no usable
+// band. All inputs are viewport-relative rects plus scroll state, so this stays
+// a pure function (unit-testable without a DOM, like axeConformanceLevel).
+export function overlayAwareScrollTop(
```
cderv, line 50: I wondered why this was exported but I understand this is for the test... 

Though this will export the function to the user client side too I think. So I wonder if this is the best way to do this. 🤔 

I wondered also why it was not inside `QuartoAxeReporter` class directly, but maybe for the same reason.

tests/unit/axe-overlay-scroll.test.ts
cderv, line 1: I think this file is to test `overlayAwareScrollTop` from the .js file and not the .ts file - usually our unit test are for .ts file. I wonder if we should verify this in Playwright test, which would avoid exporting the function. 

I need to run that idea through our robots maybe for advice, but that's my thinking for now. 

Maybe this is ok as is.

cderv (approved):
Looks good ! Nice improvments. 

Something I questioned while reviewing: is exporting `overlayAwareScrollTop` only to be able to test it in unit test a good idea, as this will expose the function in browser console context unscoped. I am curious to ask Claude to see if this is good thinking or just me worrying too much on best practices.

src/resources/formats/html/axe/axe-check.js
cderv: Ran it through the sanity check — safe, no client-side leak.

`axe-check.js` is injected as `<script type="module">`, and ES-module exports
never land on the global object. In the actual rendered output:

- `typeof window.overlayAwareScrollTop` → `"undefined"`
- typing `overlayAwareScrollTop` in the console → `ReferenceError: ... is not defined`
- for contrast, `typeof window.axe` → `"object"`, because `axe.min.js` is a
  *classic* script — that one is global, the module's exports are not.

The `export` is reachable only via `import`, and nothing in the browser imports
this module (it self-initializes). It exists purely so the Deno unit test can
import it — same pattern we already use for `axeConformanceLevel`
(`tests/unit/axe-conformance.test.ts`), which likewise doesn't leak. Keeping it
a module-level pure function (rather than a method) is what makes the geometry
DOM-free unit-testable. Good as-is.

cderv:
Update: I've reverted that follow-up (`3dc213376` → `044702a43`); the 
branch is back to its reviewed state.

On a closer look it wasn't fixing a real bug. A multi-selector 
`node.target` is a cross-frame selector chain, and `highlightTarget` 
resolves each selector with `document.querySelector`, which doesn't 
pierce iframe boundaries — so only the first selector (the frame element)
ever resolves and scrolls; the deeper ones hit the existing `if 
(!element) return null` guard and no-op. There were no competing scrolls
to fix. Sorry for the noise.

PR #14706, opened 2026-07-17
Run smoke tests against a built quarto (binary mode)

cderv:
> [!NOTE]
> Stacked on #14705 (base branch `test/verify-vacuous-pass`). Review/merge that first; this branch includes its commit and will rebase cleanly once it lands.

Smoke tests only ever run against the in-process dev TypeScript sources (version `99.9.9`). Bugs that live in the shipped product — bundling, the packaged Lua filter set, the launcher, signing/notarization side effects — have no CI coverage, because the code path that exhibits them is never executed by the harness.

This adds a binary mode so the same smoke suite can run against a built quarto distribution, with no change to how the verifiers work.

## Harness

Every `testQuartoCmd()`-based test now invokes quarto through one dispatch point, `runQuarto()` in `tests/quarto-cmd.ts`. In dev mode (default) it calls the in-process `quarto()` entry point exactly as before. When `QUARTO_TEST_BIN` points at a built quarto extracted outside the checkout, it spawns that binary as a subprocess with `--log <file> --log-format json-stream`, merges the child log into the test log, and the verifiers run unchanged — they only ever see log records and rendered output.

Dev callers are untouched; binary mode is opt-in via one env var. Around 30 smoke tests that spawned quarto directly or imported `src/quarto.ts` were migrated onto the dispatch helper so they work in both modes; a `requiresDevQuarto` escape hatch skips the few that exercise in-process internals.

## CI

`test-smokes.yml` gains a `quarto-install: dev | release | artifact` input (dev callers unchanged). A new `test-smokes-built.yml` orchestrates three sources for the binary under test:

- `nightly` — reuses the signed artifacts of a nightly `create-release` build, triggered automatically via `workflow_run`. Tests exactly what the release pipeline produces, on all three OSes, before anything is published.
- `build` — builds a fresh linux dist from the current ref; works on forks and PR branches; manual dispatch.
- `release` — installs a published (pre-)release at its tag, for post-publish verification.

`workflow_run` was chosen over modifying the release pipeline so smoke failures can never redden — or block — a real publish. The shared build step is extracted into a `build-dist-tarball` composite action, and `create-release.yml` gains a `smoke-artifacts-only` fast path for cheap branch builds.

Design decisions and their rationale are recorded in `llm-docs/built-version-testing-architecture.md`.

## Test hardening

Migrating tests onto the dispatch seam exposed several that were silently vacuous — assertions the harness swallowed, or fixtures that never existed. Fixed here: a `writer.releaseLock()`-before-`close()` bug in the editor-support helper, and two missing fixtures (`convert/issue-12318`, `editor-support/all.qmd`). The shared `verify.ts` vacuous-pass fixes these rely on are in #14705, which this stacks on.

## Testing

Test-infrastructure only — no user-facing change, no changelog entry. Dev-mode CI exercises the refactored dispatch path on this PR. Binary mode is verified by dispatching `test-smokes-built.yml` with `source=build`, and after merge by the automatic nightly `workflow_run`.


cderv:
None of the new dispatch-only scenarios run automatically on this branch's own checks, so I exercised them manually via `workflow_dispatch` to see what's actually verifiable before merge.

`source=build` (the default path) exercises the full built-artifact leg: build, ff-matrix, smoke, and playwright all run against the same build. I ran it both without a `buckets` value and with a valid one (`["integration/playwright-tests.test.ts"]`), both dispatched fine and are running as expected.

`source=release` can't be validated pre-merge. The preflight check
https://github.com/quarto-dev/quarto-cli/blob/fa012b2a424d954186339f6efec82de717603fec/.github/workflows/test-smokes-built.yml#L147-L150
requires `tests/quarto-cmd.ts` to exist at the resolved release tag, and no released tag has that file yet since it only exists on this branch. So this path will keep failing preflight until this merges and a new (pre-)release is cut afterwards.

`source=nightly` needs a Build Installers run first (`smoke-artifacts-only=true` gives a cheap linux+windows-only build), then a second dispatch pointing `run-id` at it. The installer build I ran for this completed successfully, I haven't chained the nightly smoke dispatch off it yet.

One thing worth adjusting: the `buckets` input description says "e.g. the ff-matrix bucket", but `"ff-matrix"` isn't a valid value there. It's a separate job
https://github.com/quarto-dev/quarto-cli/blob/fa012b2a424d954186339f6efec82de717603fec/.github/workflows/test-smokes-built.yml#L101
gated on `buckets == ''`, so a non-empty `buckets` value disables it rather than scoping to it. Passing `"ff-matrix"` through to the smoke bucket loop
https://github.com/quarto-dev/quarto-cli/blob/fa012b2a424d954186339f6efec82de717603fec/.github/workflows/test-smokes.yml#L488-L508
just fails since it isn't a real test file or bucket name. We could reword the description so it doesn't read as a valid example.

cderv:
I dispatched all six `workflow_dispatch` paths of `test-smokes-built.yml` to verify the built-version testing infrastructure end to end.

`source=build`, full run, no buckets: feature-format matrix and Playwright both passed. The smoke job ran to completion, 2043 passed, 17 failed, and it failed on real content, not on wiring. Four distinct pre-existing quarto bugs surfaced, none introduced by this PR, none of them binary-mode-specific either. They just never got exercised before because this is the first CI path that actually renders through a built (not dev) binary.

1. Rendering two `docs/smoke-all` fixtures using the `{{< version >}}` shortcode crashes with `bad argument #1 to 'concat' (table expected, got string)` in `version.lua`. `quarto.config.version()` in `init.lua` is polymorphic, a pandoc `Version` object when the version string parses under pandoc's strict dotted-integer parser, a plain string otherwise. The workflow stamps a semver build-metadata suffix onto the test build's version so the TS-side `quarto-required` gating still works, and that suffix satisfies semver but not pandoc's parser. Same path is reachable on a real packaged build too, via `QUARTO_FORCE_VERSION` or `quarto-bld prepare-dist --set-version`, so this isn't only a CI artifact. Fix (normalize `init.lua`'s `version()` to always return a Version object) is implemented and pushed to a separate branch, not part of this PR.

2. `quarto call axe` rejects `--log`, `Unknown option "--log". Did you mean option "--help"?`, so axe-core never runs and no `findings.json` gets produced. `appendLogOptions` in `src/core/log.ts` forwards log options to a command's direct subcommands via `getCommands()`, which filters out hidden commands by default, and `axe` is registered `.hidden()`. This predates this PR and reproduces against dev quarto directly, it's just that dev-mode tests capture logs in-process and never actually pass `--log` on the command line. Fix (`getCommands(true)`) is implemented and pushed too.

3. Six `log-level-and-formats` tests fail because renders that should log an ERROR record produce none at all. A pandoc or Lua-filter failure rejects with no reason, gets rewrapped into an empty-message Error, and `logError()` only falls back to printing something when `QUARTO_DEBUG` is set. Dev quarto forces `QUARTO_DEBUG=true`, which is the only reason dev-mode output looks fine here. A real released binary hitting the same failure emits zero error output on exit code 1. This is the one I'd actually worry about, a render can fail silently for anyone running a built quarto. A full implementation plan is written, the fix itself hasn't started yet.

4. Five smoke-all blog listing fixtures warn about being unable to read a sibling item's rendered HTML for description or preview-image content. Turned out to be a test-harness bug, not a product one: `smoke-all.test.ts` deletes a file's own output right after its own test finishes, and a later listing rebuild can need an earlier sibling's HTML that's already gone. Reproduces in dev mode too, so it's ordering, not binary mode. Fix is implemented and pushed.

`source=build` with a valid bucket (`buckets=["integration/playwright-tests.test.ts"]`): succeeded clean.

`Build Installers` with `smoke-artifacts-only=true`: succeeded.

`source=release`: failed as expected. Preflight correctly blocks release mode until a post-merge release actually has `tests/quarto-cmd.ts` in its tree. Not a bug.

`source=build` with an invalid bucket (`buckets=["ff-matrix"]`): failed on my own bad input, not a bug. `ff-matrix` is a separate job gated on `buckets==''`, not a valid bucket value.

`source=nightly`: same four bugs on both the Linux and Windows legs, feature-format matrix and Playwright green on both platforms. Confirms none of this is specific to the build-artifact path.

I'd hold off merging this until at least the version.lua crash and the missing-ERROR-logs bug are fixed or explicitly triaged. Both are real regressions independent of this PR's own scope, this is just the first time anything actually exercised them. The axe bug is real too but lower severity, your call whether it blocks this one. The listing warnings are test-infra noise only, not a blocker.

cderv:
One more read through the full diff before merging turned up four things, none blocking.

`tests/smoke/smoke-all.test.ts`'s project pre-render sits in a bare top-level `for` loop, no try/catch:
https://github.com/quarto-dev/quarto-cli/blob/c9dd785293e63f6bf6ecb878c2b090895fe0e15c/tests/smoke/smoke-all.test.ts#L448-L457
In dev mode a failed pre-render goes through `quarto()`'s `exitWithCleanup(1)`, a clean process exit. In binary mode `runQuarto` defaults `throwOnFailure` to true, so a failing pre-render throws uncaught at module top level instead, which I think aborts the whole test file load rather than failing just that project. Different case from the execute()-swallows-errors anti-pattern since this runs before any test even exists. Worth confirming with a fixture that fails to render before deciding if that's actually what we want.

`test-smokes-built.yml` repeats the same ternary in about 15 job-level `if:` conditions, e.g.
https://github.com/quarto-dev/quarto-cli/blob/c9dd785293e63f6bf6ecb878c2b090895fe0e15c/.github/workflows/test-smokes-built.yml#L44
The logic itself checks out but that many hand copies will drift the next time a source or a leg gets added. Could collapse into one `resolve-mode` job whose output the rest condition on.

The env-strip list for `QUARTO_TEST_BIN` spawns is hand-maintained in three places now, `tests/quarto-cmd.ts`'s `kStripEnvVars`
https://github.com/quarto-dev/quarto-cli/blob/c9dd785293e63f6bf6ecb878c2b090895fe0e15c/tests/quarto-cmd.ts#L18
and the smaller probe-only lists in `run-tests.ps1`/`run-tests.sh`. The shell ones are missing a few vars the TS one has (QUARTO_PROFILE, QUARTO_LOG*, RSTUDIO...). Harmless today since the probe only needs `--version` back, but worth a look at whether they can share one source instead of three copies staying in sync by hand.

`test-smokes-built.yml`'s `workflow_run` trigger has no `branches:` filter:
https://github.com/quarto-dev/quarto-cli/blob/c9dd785293e63f6bf6ecb878c2b090895fe0e15c/.github/workflows/test-smokes-built.yml#L5-L7
so any completed Build Installers run, including a `workflow_dispatch` off a feature branch, fans out the full smoke/playwright/ff-matrix triple. Probably want to gate `resolve-nightly` on `head_branch == 'main'` or similar.

Tracking all four as follow-ups, not holding up the merge on them.

PR #14707, opened 2026-07-19
fix(html): make code-tools toggle work when code-copy is enabled

### LINKED ISSUE #13583: code-tools buttons don't work if code-copy is enabled
### I have:

- [ ] searched the issue tracker for similar issues
- [ ] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [ ] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

When using code-tools: true, the "Show All Code" / "Hide All Code" buttons don't work if code-copy is also true.

I think it's due to code-copy feature adding a `<div class="code-copy-outer-scaffold"/>` around the code block, which makes the code-tools querySelector fail.

When I set code-copy to false, the buttons work.

### Steps to reproduce

Here is a document that shows the problem:

````qmd
---
title: "Reproducible Quarto Document"
format:
  html:
    code-fold: true
    code-tools: true
    code-copy: true
engine: jupyter
---

Example with code-tools:true and code-copy:true

```{python}
print("Hello, world!")
```
````

### Actual behavior

_No response_

### Expected behavior

_No response_

### Your environment

- OS: MacOS Sequoia 15.4.1
- Tested on latest quarto pre-release (1.9.8)

### Quarto check output

```bash
 ~ > quarto check
Quarto 1.9.8
      Quarto cache location: /Users/etienne/Library/Caches/quarto
      Pandoc version 3.6.3: OK
      Dart Sass version 1.87.0: OK
      NOTE: Deno version 2.4.5 does not strictly match 2.3.1 and strict checking is enabled. Please use 2.3.1.
      Typst version 0.13.0: OK
      Version: 1.9.8
      Path: /Users/etienne/Applications/quarto/bin

      TinyTeX: v2025.10
      Chromium: (not installed)

      Using: TinyTex
      Path: /Users/etienne/Library/TinyTeX/bin/universal-darwin
      Version: 2025

      Using: Chrome found on system
      Source: MacOS known location

      Version: 3.12.0
      Path: /Users/etienne/libs/quarto-bug/.venv/bin/python3
      Jupyter: 5.9.1
      Kernels: python3

```

mcanouil:
The `code-copy` scaffold introduced in 1.8.15 wraps each code block in a `div.code-copy-outer-scaffold`, which broke the "Show All Code" / "Hide All Code" handler's direct-child selectors. The menu rendered but clicks did nothing. Selectors now key off `details.code-fold` and `div.sourceCode.cell-code`, which the scaffold cannot displace.

- Fixes the fold toggle, which matched zero elements and also mistook the scaffold for the `<details>`.
- Fixes the hidden/unhidden toggle for `echo: false` cells under `keep-hidden: true`, broken the same way but not reported (found/verified during Claude's review).
- Applies the same selector to the server-side toggle detection, which previously worked only because that post-processor happens to run before the scaffold is inserted.
- Also picks up folded blocks with `filename`, and folded blocks inside tabsets, callouts and listings, which the old direct-child selector never reached.

Verified in headless Chromium by clicking both menu items on a rendered document: all folded blocks open and close together and the hidden block flips both ways. Patching the selectors back to their previous form on the same page reproduces the inert behaviour.

Closes #13583.

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -2,6 +2,7 @@ All changes included in 1.10:
 
 ## Regression fixes
 
+- ([#13583](https://github.com/quarto-dev/quarto-cli/issues/13583)): Fix "Show All Code" and "Hide All Code" in the `code-tools` menu doing nothing when `code-copy` is enabled. (author: @mcanouil)
 - ([#14267](https://github.com/quarto-dev/quarto-cli/issues/14267)): Fix Windows paths with accented characters (e.g., `C:\Users\Sébastien\`) breaking dart-sass compilation.
 - ([#14281](https://github.com/quarto-dev/quarto-cli/issues/14281)): Fix transient `.quarto_ipynb` files accumulating during `quarto preview` with Jupyter engine.
 - ([#14298](https://github.com/quarto-dev/quarto-cli/issues/14298)): Fix `quarto preview` browse URL including output filename (e.g., `hello.html`) for single-file documents, breaking Posit Workbench proxied server access.
--- a/src/command/render/codetools.ts
+++ b/src/command/render/codetools.ts
@@ -353,7 +353,7 @@ function resolveCodeTools(format: Format, doc: Document): CodeTools {
 
   // if we have requested toggle, make sure there are things to toggle
   if (codeToolsResolved.toggle) {
-    const codeDetails = doc.querySelector(".cell > details > .sourceCode");
+    const codeDetails = doc.querySelector(".cell details.code-fold");
 
     // we don't OJS hidden cells in this check, since when echo: false, we emit them hidden
     const codeHidden = doc.querySelector(
--- a/src/resources/formats/html/templates/quarto-html-after-body.ejs
+++ b/src/resources/formats/html/templates/quarto-html-after-body.ejs
@@ -149,16 +149,16 @@
     }
     function toggleCodeHandler(show) {
       return function(e) {
-        const detailsSrc = window.document.querySelectorAll(".cell > details > .sourceCode");
-        for (let i=0; i<detailsSrc.length; i++) {
-          const details = detailsSrc[i].parentElement;
+        const cellDetails = window.document.querySelectorAll(".cell details.code-fold");
+        for (let i=0; i<cellDetails.length; i++) {
+          const details = cellDetails[i];
           if (show) {
             details.open = true;
           } else {
             details.removeAttribute("open");
           }
         }
-        const cellCodeDivs = window.document.querySelectorAll(".cell > .sourceCode");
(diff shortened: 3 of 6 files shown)


news/changelog-1.10.md
```diff
@@ -2,6 +2,7 @@ All changes included in 1.10:
 
 ## Regression fixes
 
+- ([#13583](https://github.com/quarto-dev/quarto-cli/issues/13583)): Fix "Show All Code" and "Hide All Code" in the `code-tools` menu doing nothing when `code-copy` is enabled. (author: @mcanouil)
```
cderv, line 5: Are you planning to backport this to 1.9 once merged ? I don't see a backport PR. I can do it.

Though, as we will release 1.10 soon - I don't know if we'll do a new 1.9 backport 🤔

cderv (approved):
Thanks @mcanouil.

One thing before merge: we're missing a test that actually exercises the toggle. The smoke fixture locks the DOM shape, but since this PR doesn't touch the render pipeline, the pre-fix version produces identical HTML — so the fixture passes with or without the fix. I confirmed it by reverting the two changed files to their pre-fix state and re-running the new fixture still passes green, so it wouldn't have caught the original regression.

The behavior is pure client-side JS, so it needs a Playwright test that clicks **Show All Code** / **Hide All Code** and asserts the `<details>` open/close and the `hidden`↔`unhidden` swap on hidden cells, with `code-copy` enabled. I'll add that before merging.

news/changelog-1.10.md
cderv: We have nothing else on v1.9 so probably v1.10 only and we release this really soon.

cderv:
Follow-up: added the Playwright test and pushed it to the branch.

`tests/integration/playwright/tests/html-code-tools-toggle.spec.ts` plus an `{ojs}` fixture — `{ojs}` so it needs no Python/R kernel while still producing the same folded + `code-copy` scaffold DOM. It clicks **Show All Code** / **Hide All Code** and asserts every `<details>` opens/closes and the `echo: false` cell's source swaps `hidden`↔`unhidden`.

It's a real regression guard: it fails on the pre-fix code (the `<details>` never opens after *Show All Code*) and passes on this PR. I kept your smoke fixture alongside it — the two are complementary, one locks the rendered DOM contract, the other the JS behavior.

mcanouil:
Thanks, not yet used to playwright testing.

cderv:
No problem. 🤖 should know how to write them with the context present. We don't need always some playwright, but for things that impact JS usually we need to verify if the regression can be tested without in-browser behavior or not. Some times our Deno test framework can't test. 

I'll merge. 

Thanks again !

mcanouil:
Claude might have caught this, unfortunately, I did it without (quite limited subscription at the moment)

PR #14708, opened 2026-07-19
Support fully offline builds from a pre-populated deno cache

eejd:
## Summary

Two small changes needed to build `quarto.js` from a source tarball (no `.git`, no network) against a pre-populated, checksummed `DENO_DIR`, as required by distro packaging (we're authoring a MacPorts port — see [macports-ports-local#212](https://github.com/eejd/macports-ports-local/issues/212)).

- **`src/webui/quarto-preview/build.ts`**: `buildFromGit()` unconditionally triggers a network-dependent `npm install && npm run build` whenever `git ls-files` fails — which it always does when building from a release tarball rather than a git checkout, even though `quarto-preview.js` is already committed to the tree. This adds an explicit `QUARTO_SKIP_PREVIEW_BUILD` opt-out so packagers can use the shipped build instead of triggering a spurious (and network-dependent) rebuild.
- **`package/scripts/vendoring/vendor.sh`**: the `deno install` calls that resolve quarto's TypeScript module graph have no way to enforce that resolution stays local to an already-populated cache. This adds `QUARTO_VENDOR_LOCK` (passes `--lock=<path> --frozen`) and `QUARTO_VENDOR_OFFLINE=1` (passes `--cached-only`) so a cache miss fails loudly instead of silently reaching out to the network.

Both are additive, env-var-gated, and no-ops unless explicitly opted into — no change to existing default behavior.

## Test plan

Verified end-to-end on macOS/arm64: a full `configure.sh` + `quarto-bld prepare-dist` run against system-provided `deno`/`pandoc`/`typst`/`dart-sass` plus from-source `esbuild`/`deno_dom`, with `QUARTO_SKIP_PREVIEW_BUILD=1 QUARTO_VENDOR_OFFLINE=1` and a pre-populated `DENO_DIR`, completes with **zero network fetches** and produces a working `quarto.js` that renders documents correctly (`quarto check` all green, `quarto render` produces valid self-contained HTML).

- [x] `configure.sh` completes with 0 downloads
- [x] `quarto-bld prepare-dist` completes with 0 downloads
- [x] `quarto check` passes
- [x] `quarto render hello.qmd --to html` produces correct output


cderv (changes requested):
Thanks for the PR - before we can review can you rebase you commits unto quarto-cli `main` branch from upstream. Looking at the number of commits in this PR, and the conflict, I believe your feature branch is not based on the correct HEAD. 

Thank you

cderv:
Thanks for the update. Your PR still have file change that probably should be part of this PR. 

Please be sure to update your fork. You can also recreate a clean PR is this is easier and cherry pick your commit. 

This way the PR will be way cleaner for me to review and look at. Thank you.

Issue #14711, opened 2026-07-20
CI: cold-cache `renv::restore()` fails on 5-package cluster — test setup unusable on forks / fresh caches


cderv:
### Symptom

On a runner with no renv package cache, the "Restore R packages" step of
`test-smokes.yml` fails before any test runs. `renv::restore()` downloads all
145 lockfile packages, installs 140, and **never attempts** the remaining
five — reporting only circular "dependency failed" messages with no
underlying install error:

```
Successfully installed 140 packages in 46 seconds.

The following package(s) were not installed successfully:
- [chromote]: dependency failed (zip)
- [flextable]: dependency failed (flextable, webshot2, officer)
- [officer]: dependency failed (flextable, webshot2, zip)
- [webshot2]: dependency failed (chromote)
- [zip]: dependency failed (flextable, webshot2)

Error: failed to install "chromote", "flextable", "officer", "webshot2", "zip"
```

Reproduces **identically on ubuntu-latest and windows-latest** (rules out
compilers/system deps). All five are recorded as normal P3M `Repository`
entries in `tests/renv.lock`; their downloads succeed; there is zero
install-attempt output for them between "# Installing packages" and the
failure summary.

- Environment: renv 1.2.3, R 4.5.2 (lockfile records 4.5.1), P3M binary repo
  (`__linux__/noble/latest` on Linux), `RENV_PATHS_ROOT` empty (cache miss).
- Evidence runs (fork, workflow_dispatch): cderv/quarto-cli runs
  29763939514 / 29763944071 / 29763956751 (all three identical, both OSes).
- The affected versions entered the lockfile in `effe145` ("tests - update
  R env").

### Why upstream CI doesn't see it

Every upstream run restores these packages from a **warm renv package
cache** (link from cache, no install scheduling), so the failure is latent.
Any cold-cache environment hits it deterministically: new forks, cache
eviction, a runner OS image bump that changes the cache key, or new OS
targets (e.g. future macOS built-version legs). Upstream CI is one
cache-cleanup away from going red in setup.

### Reproduction

Dispatch `test-smokes.yml` on any fork (or clear the
`*-renv-*` Actions caches upstream) and watch the "Restore R packages"
step. No code changes needed.

### Workaround (verified)

Wrapping `renv::restore()` in `try()` and downgrading the
`renv::status()$synchronized` check to a warning lets the run proceed;
the 5 packages are only needed by the docx/screenshot test stack
[...]

PR #14715, opened 2026-07-21
Group CI logs and surface failures on the default (non-bucketed) test run

cderv:
## Problem

The default (non-bucketed) test path runs one giant `deno test` over every discovered file and emits no workflow commands, so the CI log is thousands of flat lines — finding a failure means scrolling or text search. This is the path the smoke leg of `test-smokes-built.yml` runs by default (empty `buckets`); the playwright and ff-matrix legs pass a non-empty `buckets`, so they already go through the per-file `::group::` bucket loop added by #13787 and #13807. Those PRs grouped only the bucket loops (one file per `run-tests.sh` invocation) — the default path was left flat, and no per-test `::error` annotation or step-summary table existed on either path.

## What it does

Per-test-file `::group::` markers plus per-test `::error` annotations on the default path, and `GITHUB_STEP_SUMMARY` failure tables in every mode:

- Annotations are capped at 9 per-test errors + 1 inline aggregate per step, coordinated across Deno's per-file module instances via a sidecar counter file (a module-level counter is not a per-step budget, since Deno instantiates each test file's module graph separately).
- Step summary tables are the complete failure record, size-budgeted against the ~1 MiB silent-drop bug (self-truncate near 512 KiB, coordinated cross-process via the summary file size).
- All group/annotation emission is gated on the harness owning the step: the `QUARTO_TESTS_GHA_ORCHESTRATED` env contract (set by the bucket loops) keeps the #13807 bucket-log format byte-identical and doubles as a no-revert kill switch. Everything else gates on `GITHUB_ACTIONS`, so local output is byte-identical.
- Includes `tests/tools/check-gha-log.ts`, a mechanical invariant checker to re-run on Deno version bumps (reporter framing is version-sensitive).

Design of record (constraints, spike results, invariants, verification plan) lives in the branch at [`dev-docs/ci-test-log-grouping-design.md`](https://github.com/quarto-dev/quarto-cli/blob/test/gha-log-grouping/dev-docs/ci-test-log-grouping-design.md).

## Evidence

Fork trial matrix with seeded failures, Linux + Windows:

- Default path — [run 29826031431](https://github.com/cderv/quarto-cli/actions/runs/29826031431): grouped log; exactly 9 per-test + 1 aggregate annotations from 12 failing files; checker green on both OSes' raw logs.
- Orchestrated bucket — [run 29767185290](https://github.com/cderv/quarto-cli/actions/runs/29767185290): bucket format unchanged, zero harness annotations.
- 12-failing-file bucket — [run 29767190539](https://github.com/cderv/quarto-cli/actions/runs/29767190539): GitHub silently drops annotations past 10/step (observed); the step summary stays complete. This is the motivating case.

Two real bugs the trial caught and fixed before this PR: Deno's per-file module instances invalidating a per-process budget, and explicit `undefined` args triggering default parameters so unit tests touched the real CI counter/summary/env (null-sentinel fix).

## Notes

> [!NOTE]
> Stacked on #14706; GitHub retargets this to `main` when it merges.

Related but independent finding from the trial: #14711 (cold-cache renv deadlock).

A built-mode (source=build) confirmation run on this branch is linked in a follow-up comment.

<details>

- **Codebase grounding**: local clone, DeepWiki MCP

</details>

## At-scale evidence (2026-07-22/23)

**Run:** full-serial dispatch of `test-smokes.yml` on `test/gha-log-grouping` @ 0476b338f (https://github.com/quarto-dev/quarto-cli/actions/runs/29952972915), Linux + Windows, dev quarto. RED by design: 36 pre-existing order-dependent failures per OS (they pass in isolation; tracked separately) — used here as at-scale evidence for the failure-surfacing machinery.

**Grouping (Phases 2/2.1):** full-log survey found zero invariant violations — groups balanced, never nested, one group per registering harness file, failures and the terminal ERRORS/FAILURES sections always outside groups. Registration-time open captures even heavy module-eval output (smoke-all's ~65 s of registration-phase project renders land inside its group). Expected in this log and out of scope for this PR: ~17 min of ungrouped output around `integration/playwright-tests.test.ts` (raw `Deno.test`, bypasses the harness wrapper; no invariant violated; absent from the built workflow's per-leg structure, where playwright runs as its own leg), plus four other raw-`Deno.test` files (`smoke/create/create.test.ts`, `smoke/logging/log-level-and-formats.test.ts`, and the two CI-merged julia-engine subtree files) — a documented follow-up covers these.

**Failure surfacing (Phases 1/2.2) at scale, verified on both OS jobs:**
- Summary tables `L-F1..L-F36` / `W-F1..W-F36`: ordinals strictly increasing, plain-text labels, every row carrying its own label and doc-path repro.
- Clustering: 36 failures → 6 blocks per OS (30 + 2 + 1 + 1 + 1 + 1); one shared excerpt per cluster; member lists with per-member labels and repros; identical cluster structure on Linux and Windows (deterministic).
- Annotations: exactly (9 + 1 aggregate) × 2 jobs = 20 errors; titles `L-Fn · <test>`; bodies = doc-path repro + trimmed excerpt + "Full output: step summary → L-Fn" pointer; the aggregate points at the step summary for the complete list.
- Ctrl+F navigation unambiguous across OS sections via the L-/W- prefixes; summary size far below the 512 KiB budget; no degraded rows.

**Built-version run:** `test-smokes-built.yml` dispatch on the same head (https://github.com/quarto-dev/quarto-cli/actions/runs/29995036875). Playwright leg GREEN as its own job (the structural fix for the serial run's ungrouped playwright stretch, visible in practice); ff-matrix leg red on a known pre-existing issue; smoke leg red with 8 failures — the summary machinery behaves identically to the dev run (strictly increasing `L-F1..L-F8`, per-signature clusters with member lists, doc-path repros), now demonstrated in binary mode. Notably, the dev run's 30-failure book-crossref family is entirely absent here: dev mode renders in-process (one Deno process for the whole suite) while binary mode spawns a fresh built quarto per render, so that family is a dev-serial-environment artifact users never see — the daily built structure is inherently free of it. Five listings failures reproduce from the dev run (a filesystem-level sibling-output dependency, separate triage); three failures are new in built mode (two missing-output, one pdf compile) and are being investigated as part of built-version testing, independent of this PR.


cderv:
Phase 2.1 (opening each per-file `::group::` at registration time) is on the branch and confirmed in a real Actions run — a narrowed subset of the default path, run on a fork because the full default path is a ~2h serial run and is currently blocked by the cold-cache renv issue #14711

`running N tests from …` and the first-test announcement now sit inside each file's group instead of above it — 5 per-file groups, no nesting, log checker clean. The failure-path labeled step-summary blocks aren't exercised by this passing subset; that UI proof will ride the Phase 2.2 seeded-failure validation run.

One rendering note: the completed-run log viewer prints each `::group::` line twice (literal echo + processed command). That's pre-existing GitHub behavior for consumed workflow commands — identical on pre-2.1 runs, not from this change.

cderv:
Phase 2.2 is on the branch now: the failure step summary is navigable and clustered, and the per-test annotations are trimmed.

Each failure gets a step-wide OS-prefixed label (`L-Fn` / `W-Fn` / `M-Fn`) in a new leading table column and as a `#### L-Fn` heading before its detail block, so you can jump from a table row to its output. Identical errors now collapse into one block (keyed by the first three excerpt lines) with a member list and a single shared excerpt — on the 36-failure run that motivated this, ~36 blocks become ~4. Annotations are trimmed to the repro plus a few excerpt lines and a pointer back to the summary entry, since the full output is already in the summary on the same page.

Labels are plain text rather than in-page links: I checked on a real run and GitHub's step-summary renderer emits headings without slug anchors, so `[L-Fn](#l-fn)` fragment links don't resolve there. Ctrl+F on a label gives the same row↔block navigation, and the OS prefix keeps it unambiguous across the run page's concatenated per-job summaries.

Verified on a fork run (https://github.com/cderv/quarto-cli/actions/runs/29934718951): three seeded failures (two sharing an error, one distinct) produced exactly one 2-member cluster plus one single block, strictly-increasing `L-F1..L-F3` labels, three trimmed annotations, and the grouping checker stayed clean on the log. That run also exercises the labeled `<details>` summaries added earlier on this branch, so it doubles as the UI proof for those.

cderv:
Two follow-up commits since the evidence block (which was gathered at 0476b338f, all of it still valid — neither commit touches grouping or reporting code):

- bfe01cada — `{{< version >}}` no longer errors when `quarto.version` is the string fallback (unparseable version strings).
- 345fc800d — the built-version workflow's source=build mode stamps the plain `X.Y.Z` instead of `+test.<date>` build metadata (which pandoc's Version type cannot parse — the trigger for the above, and a regex mismatch in two version-asserting smoke docs even after the shortcode fix).

Together these fix 2 of the 8 built-smoke-leg failures in run 29995036875; the next source=build dispatch should show issue-12006/13913 green. Requires both commits (and means #14706 should not merge without this PR following).

PR #14718, opened 2026-07-22
Retry stapling notarized macOS pkg on transient failure

cderv:
The macOS release build failed at the pkg stapling step (https://github.com/quarto-dev/quarto-cli/actions/runs/29915034977/job/88906897397), right after notarization had just reported success.

## Root Cause

`stapleNotary()` in `package/src/macos/installer.ts:422` calls `xcrun stapler staple` once, immediately after `notarytool submit --wait` reports success. Apple's notarization ticket can take a few seconds to propagate before it's fetchable, so the immediate staple attempt fails with exit code 65 even though notarization succeeded.

## Fix

Retry stapling up to 5 times with a 15s delay between attempts before giving up.

Verified with `deno check` (no local way to exercise `xcrun`/`stapler` outside macOS CI).

### DIFF
--- a/.claude/rules/copyright.md
+++ b/.claude/rules/copyright.md
@@ -1,0 +1,10 @@
+---
+paths:
+  - "**/*.ts"
+  - "**/*.js"
+  - "**/*.lua"
+---
+
+# Copyright headers
+
+New source files get a single current-year header — `Copyright (C) <year> Posit Software, PBC` (e.g. `2026`), not a back-dated `2020-<year>` range. Leave existing files' headers alone unless you're already substantially editing them.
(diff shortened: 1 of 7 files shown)


cderv:
Running the build installer workflow to verify it works 
- https://github.com/quarto-dev/quarto-cli/actions/runs/29925294593

cderv:
And it is still failing... we have a Mac signing problem again 😢

cderv:
All failure are from Julia problem on Windows it seems...

cderv:
I am merging despited windows failure, all Julia related. Something else to look at

PR #14723, opened 2026-07-23
Prep Pandoc 3.10.1 update, gate version bump behind S3 archival

cderv:
Pandoc released 3.10.1. Quarto currently bundles 3.10.0 (landed in #14664). Actually bumping the bundled version means archiving the new binary to S3 and, since main just recovered from a bundled-binary notarization break (#14718), running the release signing/notarization dry-run before merging. Neither of those happens in this PR - it only prepares, so the eventual bump is lower-risk and can be done incrementally by whoever holds S3 credentials.

Resyncs the Pandoc-derived template files that changed between 3.10.0 and 3.10.1: `tables.tex` (pdf and beamer), `typst.template`, and `html/pandoc/styles.html`, each verified byte-exact against upstream and via real renders.

Fixes stale doc and `llm-docs` claims about which template files are Quarto's wired-up copies versus Pandoc's dev-reference-only copies, across every format. The naming trap those docs describe (`latex.common` vs `common.latex`, `styles.html` vs `html.styles`) caught two patches in this same prep before the doc audit caught it.

Adds an opt-in `pandoc-override-version` input to `test-smokes.yml` so a not-yet-archived Pandoc version can be smoke-tested via `QUARTO_PANDOC` without needing S3 access.

Adds `--skip-archive` to `quarto-bld update-pandoc` so the template and `format-extension.ts` regeneration can be prepared and verified against a `QUARTO_PANDOC` binary without S3 credentials, leaving the result as an uncommitted diff to review - a safety net against the same naming trap for future updates.

`configuration` stays pinned at 3.10.0. The version bump, S3 archival, and release dry-run are Phase 2 in `dev-docs/update-pandoc-checklist.md`, for whoever holds S3 credentials to do right before merge.

## Test Plan

- [ ] Dispatch `test-smokes.yml` with `pandoc-override-version: 3.10.1` (scoped to typst/latex/table buckets) and confirm a clean run
- [ ] Run `./package/src/quarto-bld update-pandoc 3.10.1 --skip-archive` with `QUARTO_PANDOC` pointed at a 3.10.1 binary - expect an empty diff, confirming the templates above already match

### DIFF
--- a/.claude/rules/dev-tools/dependency-updates.md
+++ b/.claude/rules/dev-tools/dependency-updates.md
@@ -1,0 +1,10 @@
+---
+description: "Bundled dependency version bumps (Deno, Pandoc, Dart Sass, Typst, esbuild, veraPDF)"
+paths:
+  - "configuration"
+  - "src/command/check/check.ts"
+---
+
+# Bundled Dependency Updates
+
+Bumping any bundled binary version starts here: see `dev-docs/upgrade-dependencies.md` for the general procedure (version numbers, `check.ts` constraints, the installer signing/notarization dry-run gate before merging). For Pandoc specifically, `dev-docs/upgrade-dependencies.md` links to `dev-docs/update-pandoc-checklist.md` (template resync, schema/lua-types checks, CI smoke-test path).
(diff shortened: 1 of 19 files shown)


cderv:
This was wrongly merged. It seems 3.10.1 is still not in S3 - I'll revert and recreate

Issue #13696, opened 2025-11-17
Support resvg as alternative to rsvg-convert for SVG rendering


cderv:
Installing rsvg-convert on Windows can be challenging for users, requiring either manual binary installation or complex package manager setup. This affects workflows that convert SVG to PDF/LaTeX output, including mermaid diagrams with `mermaid-format: svg`.

resvg (https://github.com/linebender/resvg) is a modern Rust-based SVG rendering library with better cross-platform support. It provides pre-built binaries for all platforms and is available through Windows package managers (scoop, winget), making installation significantly easier for Windows users.

We could add resvg as an alternative SVG converter alongside rsvg-convert and Inkscape. The conversion pipeline in `src/resources/filters/quarto-post/pdf-images.lua` already handles rsvg-convert and has infrastructure for tool detection.

https://github.com/quarto-dev/quarto-cli/blob/c9008d7cb1d088f6837fc41ed1f138b8f443a9d1/src/resources/filters/quarto-post/pdf-images.lua#L28-L61

We could implement this similar to the existing `use-rsvg-convert` option, perhaps with a more general `svg-converter` option that accepts `resvg`, `rsvg-convert`, or `inkscape`, with automatic fallback to whichever tool is available.

This came up while investigating #13661, where better SVG tooling options would help Windows users. Directly addresses the request in #1144 for specifying alternative SVG conversion tools.

cderv:
Though it only convers to PNG... not PDF 
They consider out of scope: https://github.com/linebender/resvg/issues/65

So it can't be used for PDF images right now. Too bad.

jkrumbiegel:
What about using Typst for that? You already bundle it and it sort of contains SVG to PDF conversion machinery. Maybe with something like

```typst
#set page(width: auto, height: auto, margin: 0pt)

#image("some_figure.svg")
```

and then `typst compile input.typ some_figure.pdf`

cderv:
Really interesting idea ! I haven't thought of that. 🤔 Thanks !

cderv:
Following up on the Typst suggestion in this thread — it works well. While fixing #14722 (PR #14724) I had the placeholder shortcode rasterize its generated SVG locally through the bundled Typst binary, and Typst handles both SVG→PNG and SVG→PDF (vector, with selectable text) with no external tool and no network:

```
typst
#set page(width: auto, height: auto, margin: 0pt)
#image("figure.svg")
```

[...]

Issue #1144, opened 2022-06-14
How to specify other SVG conversion tools (Inkscape) into the chain


jjallaire:
### Discussed in https://github.com/quarto-dev/quarto-cli/discussions/1140

<div type='discussions-op-text'>

<sup>Originally posted by **fuhrmanator** June 13, 2022</sup>
I'm using SVG a lot in my markdown, and it renders nicely (most of the time) in PDF by way of `rsvg-convert` part of the LaTeX chain:

> (rsvg-convert.exe:2516): Pango-WARNING **: couldn't load font "Helvetica Not-Rotated 13.640625", falling back to "Sans Not-Rotated 13.640625", expect ugly output.

I'd like to use Inkscape instead, as it does some things better, especially with fonts I've installed in Windows. 

Is there a way to specify that SVG conversions to PDF should be done using Inkscape? </div>

PR #14724, opened 2026-07-23
Fix `{{< placeholder >}}` shortcode failing on non-SVG output

### LINKED ISSUE #14722: placeholder shortcode fails on all non-SVG output: svg2png.deno.dev sunset (404)
The built-in `placeholder` shortcode renders a PNG by fetching from `https://svg2png.deno.dev/`. That service ran on Deno Deploy Classic, which was sunset on 2026-07-20, and now returns a permanent 404:

> 404 Not Found — The requested deployment does not exist. Deno Deploy Classic was sunset on July 20, 2026 and no longer serves deployments.

For every output format except Typst (which uses an inline SVG data URI), the fetch fails and the shortcode raises `Error rendering placeholder`, which aborts the whole render:

https://github.com/quarto-dev/quarto-cli/blob/d30cdbb9e8e81d2aab1a53c997d12aa24ffd242a/src/resources/extensions/quarto/placeholder/placeholder.lua#L28-L41

I can reproduce on both `html` and `pdf` with a minimal document:

```markdown
![]({{< placeholder 300 200 >}})
```

```
ERROR (.../placeholder/placeholder.lua:33) Error rendering placeholder
ERROR (.../placeholder/placeholder.lua:34) nil
```

On `pdf` the raised string then leaks into the LaTeX source as an image path, so compilation fails looking for `Error rendering placeholder.pdf`. The `svg` path (Typst, or an explicit `format=svg`) still works because it never touches the network.

This also breaks CI: the smoke-all test `tests/docs/smoke-all/2026/02/04/issue-13992-figure.qmd` renders a placeholder to PDF and now fails.

We could:

1. render the PNG locally so there is no network dependency,
2. host an equivalent svg-to-png endpoint under Quarto's control, or
3. emit the inline SVG data URI for all formats, not just Typst.

Separately, the error handling at lines 33-35 has dead code: `error()` raises immediately, so the second `error(contents)` and the `pandoc.Str("Error rendering placeholder")` return are never reached, and the raised message is what ends up used as the image path.

cderv:
When rendering `{{< placeholder >}}` to any non-SVG format (html, pdf, docx, …), the shortcode fails. It relied on the `svg2png.deno.dev` service to rasterize its generated SVG to PNG, and that service was permanently retired (Deno Deploy Classic sunset) — it now returns a 404. On html this leaks a broken ``; on pdf it aborts the render with a LaTeX file-not-found.

## Root Cause

The non-SVG branch fetched a PNG from `https://svg2png.deno.dev/<svg-data-uri>` (the graceful error handling around that fetch was added in #11111). With the service gone the fetch always fails. `error()` in Quarto's Lua filters is a logging helper that writes to stderr and returns — it does not raise — so the failure branch logged two lines and then substituted the literal string `Error rendering placeholder` as the image content, which is what surfaced as the broken image / missing file.

## Fix

Rasterize locally with the already-bundled Typst binary — no network, no new bundled assets. The shortcode writes its SVG plus a one-line Typst wrapper to a temp dir, runs `typst compile --format png --ppi 96 --ignore-system-fonts`, and returns a `data:image/png;base64,…` URI. Typst is located via the existing `quarto.paths.typst()` Lua API and invoked with `pandoc.pipe`, mirroring how `src/resources/filters/quarto-post/pdf-images.lua` shells to `rsvg-convert`. `--ignore-system-fonts` still renders the dimension label because Typst embeds portable fonts.

On failure the shortcode now aborts with `fatal()` instead of substituting text — local, offline rasterization is deterministic, so a failure is a real bug rather than the transient the old text-substitution masked.

The `format=svg` and Typst-output paths are unchanged (inline SVG, no rasterization).

## Related

Typst rasterizes SVG→PNG and SVG→PDF locally with portable fonts — the same capability requested for the general SVG-figure conversion pipeline in #13696 and #1144. This PR wires only the placeholder shortcode; extending Typst to the `pdf-images.lua` figure path is separate follow-up.

Fixes #14722

### DIFF
--- a/news/changelog-1.10.md
+++ b/news/changelog-1.10.md
@@ -139,3 +139,4 @@ All changes included in 1.10:
 - ([#14595](https://github.com/quarto-dev/quarto-cli/issues/14595)): Fix reload preview in code-server environment
 - ([#14669](https://github.com/quarto-dev/quarto-cli/issues/14669)): Fix markdown output being deleted when `output-file` has an `.html` extension and an html format is paired with a markdown format.
 - ([#14687](https://github.com/quarto-dev/quarto-cli/issues/14687)): Fix SCSS color-variable export (`--quarto-scss-export-*`) being silently skipped when a theme rule places a declaration on the same line as the opening brace with no space after the colon (e.g. `.example {width:100px;}`).
+- ([#14722](https://github.com/quarto-dev/quarto-cli/issues/14722)): Fix `{{< placeholder >}}` shortcode failing to render on non-SVG formats after an external image service was retired; placeholders now render locally with no network access.
--- a/src/resources/extensions/quarto/placeholder/placeholder.lua
+++ b/src/resources/extensions/quarto/placeholder/placeholder.lua
@@ -25,21 +25,33 @@ return {
     if output_format == "svg" then
       result = svg64
     else
-      local pcallresult, mt, contents = pcall(function()
-        local mt, contents = pandoc.mediabag.fetch("https://svg2png.deno.dev/" .. svg64)
-        return mt, contents
+      local ok, contents = pcall(function()
+        return pandoc.system.with_temporary_directory('placeholder', function(tmpdir)
+          local svg_in = pandoc.path.join({tmpdir, 'in.svg'})
+          local typ_in = pandoc.path.join({tmpdir, 'in.typ'})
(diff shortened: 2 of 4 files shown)


cderv:
This is passing ! Failing test is playwright trensient. I'll merge.

PR #14773, opened 2026-08-13
fix: keep custom language keys in the resolved translations

### LINKED ISSUE #14772: Custom `language` keys are accepted, then silently discarded
### I have:

### Bug description

A key set under `language` that is not part of Quarto's own string set passes validation, is merged over the defaults, and is then discarded before templates see it.
Reading it back through `$quarto.language.<key>$` yields an empty string, with no warning at any point.

`translationsForLang` rebuilds the language table from scratch and copies across only keys present in `kLanguageDefaultsKeys` or matching `crossref-.*-title` / `crossref-.*-prefix`, and `formatLanguage` runs it *after* merging the user's `language` object over the defaults.
So a key such as `custom-key` survives schema validation and the merge, then is dropped on the way out.

- https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/core/language.ts#L161-L173
- https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/core/language.ts#L201-L210
- https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/config/constants.ts#L385-L387

This looks like a bug rather than a deliberate restriction, for four reasons:

- The value is accepted before it is thrown away.
  The `format-language` schema is an ordinary `object` with no `closed: true`, unlike neighbouring definitions such as `website-about` a few lines below, which do set it.
  An unknown key is therefore valid input by the project's own schema rules.
- The same function already supports user-defined keys.
  The `crossref-.*-title` and `crossref-.*-prefix` escape hatches exist because users define their own crossref types, and the comment there reads "crossrefs can be custom, so be more lenient".
- The documentation never describes a closed set.
  <https://quarto.org/docs/authoring/language.html> presents `language` as a YAML file of translations and tells readers to copy `_language.yml` and translate the values.
- The failure mode is the worst available one: neither a value nor a diagnostic.
  In a template the result is an empty string, so the visible symptom is a blank heading or a missing label, at a distance from the cause.
  A typo in a *known* key, say `toc-title-documnt`, behaves identically.

- https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/schema/definitions.yml#L1372-L1376
- https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/schema/definitions.yml#L1439-L1441

It became more consequential with `$quarto.language.*$`.
Now that resolved strings are exposed to Pandoc templates under the reserved `quarto.language` namespace, format extensions can read localised strings but cannot supply any of their own.
An extension whose template contains wording such as "Signature", "Confidential" or "Page N of M" has no supported path to a localised value and must fall back to ordinary metadata, duplicating a mechanism that `language` already looks like it provides.

### Steps to reproduce

`issue.qmd`:

````qmd
---
title: "Custom `language` keys do not reach `$quarto.language.*$`"
format:
  typst:
    template-partials:
      - typst-show.typ
    keep-typ: true
language:
  toc-title-document: "Overridden built-in"
  custom-key: "Custom value"
---

The `language` schema accepts a free-form object, so `custom-key` above passes
validation, but only keys shipped in Quarto's `_language.yml` reach the
`quarto.language` template namespace.
````

`typst-show.typ`, next to it:

```typ
#show: doc => {
  [built-in overridden: `$quarto.language.toc-title-document$`]
  linebreak()
  [built-in untouched: `$quarto.language.section-title-abstract$`]
  linebreak()
  [custom key: `$quarto.language.custom-key$`]
  doc
}
```

Then:

```sh
quarto render issue.qmd
grep -n "built-in\|custom key" issue.typ
```

The same happens through the file form, `language: custom.yml`: the file loads, its override of a known key applies, and its custom key is dropped without an error.

### Actual behavior

```text
415:  [built-in overridden: `Overridden built-in`]
417:  [built-in untouched: `Abstract`]
419:  [custom key: ``]
```

The first two lines confirm the namespace itself works and that user overrides are merged, so the empty value is specific to keys absent from `_language.yml`.

| `language` entry | Read back as | Outcome |
| ---------------- | ------------ | ------- |
| `toc-title-document: "Overridden built-in"` | `Overridden built-in` | override honoured |
| `section-title-abstract` (not set) | `Abstract` | default honoured |
| `crossref-xyz-title: "Custom crossref title"` | `Custom crossref title` | custom key honoured |
| `custom-key: "Custom value"` | *(empty)* | **dropped** |

### Expected behavior

Line 419 to read `` `Custom value` ``.

More generally, either outcome would be fine, but the current silent empty string is neither:

1. **Pass unknown keys through.**
   Keep the filter for locale-variation subkeys but let unrecognised scalar keys survive into `translations`, so extension and template authors can define their own localised strings alongside Quarto's.
   This is what this report proposes, and it needs one extra clause on the condition in `translationsForLang`.
2. **Reject unknown keys.**
   Set `closed: true` on `format-language` so an unrecognised key fails validation with a message.
   This also catches typos in known keys, at the cost of ruling out extension-owned strings, and it would need the crossref patterns expressed in the schema.

### Your environment

- IDE: Visual Studio Code
- OS: macOS, `darwin` 25.6.0

### Quarto check output

Confirmed on a local dev build of `quarto-cli` at `abc6a78ed68f9e8bc9d54e27851093bd687a1cb7`, which reports version `99.9.9`.

```bash
Quarto 99.9.9
      Pandoc version 3.10.0: OK
      Dart Sass version 1.101.0: OK
      Deno version 2.7.14: OK
      Typst version 0.15.1: OK
      Version: 99.9.9
      commit: abc6a78ed68f9e8bc9d54e27851093bd687a1cb7
      Path: /Users/mcanouil/Projects/quarto-dev/quarto-cli/package/dist/bin
      TinyTeX: v2026.07
      VeraPDF: (not installed)
      Using: TinyTex
      Version: 2026
      Using: Chrome from QUARTO_CHROMIUM
      Version: 4.6.1
      Path: /Library/Frameworks/R.framework/Versions/4.6/Resources
      knitr: 1.51
      rmarkdown: 2.31
      Version: 3.14.3
      Jupyter: 5.9.1
      Kernels: julia-1.12, python3
```

mcanouil:
Closes #14772.

`translationsForLang` rebuilt the language table from an allow-list and ran after the user's `language` object was merged over the defaults, so a key Quarto does not ship was accepted by the schema, merged, then discarded before templates saw it.

- `translationsForLang` now also copies non-object values it does not recognise; locale variations are objects, so they stay out and the variation loop below keeps merging them itself.
- Custom keys prefixed `callout-`, `crossref-` or `environment-` now also reach Lua filter params, and `search-` keys reach website search options, because both consumers select by prefix over the whole table.
- There is no namespacing convention, so an extension key can collide with a key Quarto adds later. User values are merged over the defaults, so the extension's value still wins in its own documents, but requiring a prefix is an option if that is not acceptable.
- Side effect: a string added to `_language.yml` without a matching entry in `kLanguageDefaultsKeys` is no longer dropped as a built-in default.
- Possible follow-up, not done here: derive `kLanguageDefaultsKeys` from `_language.yml` at build time, so the YAML file, the constants array and the schema cannot drift.

The test renders one Typst document through a `typst-show.typ` partial and reads back four keys: the custom key, an override of a built-in key, an untouched built-in key, and a key supplied through a locale variation.


cderv (approved):
Let's merge this. I'll probably do quick adaptation related to this too, on schema and others. 

Thank you !

PR #14785, opened 2026-08-18
Restore keyboard focus indicator on bare .btn buttons

### LINKED ISSUE #14774: a11y: buttons with no Bootstrap variant class have no visible keyboard focus indicator
## I have

- [x] searched the existing issues, and did not find a duplicate.
- [x] tried the latest release of Quarto.
- [x] made a minimal reproducible example.

## Bug description

Quarto emits buttons that carry the `btn` class with no `btn-*` variant class. Such a button takes keyboard focus, matches `:focus-visible`, and shows no focus indicator at all. I found three:

| Button | Appears when |
| --- | --- |
| `button.code-tools-button` | `code-tools` is on, at any viewport width |
| `button.quarto-btn-toggle` | a website sidebar collapses (below 992px) |
| `button.quarto-search-button` | a website has sidebar search, below 992px |

This fails [WCAG 2.2 Success Criterion 2.4.7 Focus Visible](https://www.w3.org/TR/WCAG22/#focus-visible) (Level AA): "Any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible." axe-core has no rule for a missing focus indicator, so a clean axe report does not cover this.

### Cause

Bootstrap removes the native ring and substitutes its own ([`_buttons.scss#L69-L79`](https://github.com/quarto-dev/quarto-cli/blob/5d07f3ed1105dbfd19e1abd9f1d847f2b8681d38/src/resources/formats/html/bootstrap/dist/scss/_buttons.scss#L69-L79)):

```scss
&:focus-visible {
  outline: 0;
  box-shadow: var(--#{$prefix}btn-focus-box-shadow);
}
```

`--bs-btn-focus-box-shadow` references `--bs-btn-focus-shadow-rgb` ([`#L21`](https://github.com/quarto-dev/quarto-cli/blob/5d07f3ed1105dbfd19e1abd9f1d847f2b8681d38/src/resources/formats/html/bootstrap/dist/scss/_buttons.scss#L21)). Only the `btn-*` variant classes define that property. On a bare `.btn` the reference is invalid, so `box-shadow` computes to `none`, and `outline: 0` has already removed the native ring.

Bootstrap documents this behavior. Its [Base class](https://getbootstrap.com/docs/5.3/components/buttons/#base-class) section says that by default "`.btn` controls have a transparent border and background color, and lack any explicit focus and hover styles", and that the class "is intended to be used in conjunction with our button variants, or to serve as a basis for your own custom styles". It then gives this instruction:

> If you are using the `.btn` class on its own, remember to at least define some explicit `:focus` and/or `:focus-visible` styles.

Quarto uses `.btn` on its own in the places below, and defines no such styles for them.

The emission sites I found:

- [`codetools.ts#L163-L166`](https://github.com/quarto-dev/quarto-cli/blob/5d07f3ed1105dbfd19e1abd9f1d847f2b8681d38/src/command/render/codetools.ts#L163-L166) — `btn code-tools-button`, plus `dropdown-toggle` when the menu form is used
- [`nav-before-body.ejs#L66`](https://github.com/quarto-dev/quarto-cli/blob/5d07f3ed1105dbfd19e1abd9f1d847f2b8681d38/src/resources/projects/website/templates/nav-before-body.ejs#L66) — `quarto-btn-toggle btn`
- [`nav-before-body.ejs#L86`](https://github.com/quarto-dev/quarto-cli/blob/5d07f3ed1105dbfd19e1abd9f1d847f2b8681d38/src/resources/projects/website/templates/nav-before-body.ejs#L86) — `btn quarto-search-button`

Quarto's `.btn:focus:not(:focus-visible)` rule is not involved. It is scoped to mouse input and does not apply to keyboard focus.

The other Quarto buttons I checked are unaffected, because they either carry a variant class (`btn-primary`, `btn-dark`, `btn-quarto`) or are not a `.btn` at all (`code-copy-button`, `callout-btn-toggle`). `.btn-close` is a separate component with a working focus shadow. I did not audit bslib components, dashboards, or revealjs, so there may be more.

Related: #12118 fixed the same symptom on the code copy button, which is not a `.btn` and had a different cause. #26 is the origin of the mouse-versus-keyboard rule above.

## Steps to reproduce

1. Save the document below as `index.qmd`.
2. Run `quarto render index.qmd`.
3. Open `index.html`, and press <kbd>Tab</kbd>.

The first tab stop is the "Code" button, at the top right. It takes focus with no indicator.

````qmd
---
title: "Focus indicator test"
format:
  html:
    code-tools: true
    theme: cosmo
---

Press Tab. Focus moves to the Code button, which shows no indicator.

```r
1 + 1
```
````

For the two navigation buttons, run `quarto create project website focus-test`, replace `_quarto.yml` with the file below, render, then narrow the window under 992px and press <kbd>Tab</kbd>.

```yaml
project:
  type: website

website:
  title: "focus-test"
  search:
    location: sidebar
  navbar:
    left:
      - href: index.qmd
        text: Home
  sidebar:
    style: floating
    search: true
    contents:
      - index.qmd
      - about.qmd

format:
  html:
    theme: cosmo
```

## Actual behavior

All three buttons take focus with no indicator. Computed `outline` is `none 0px` and computed `box-shadow` is `none`, in light and dark themes. `--bs-btn-focus-shadow-rgb` is unset on all three.

## Expected behavior

Each button shows a visible keyboard focus indicator, as the surrounding controls do.

## Your environment

- IDE: none, Quarto CLI at the command line
- OS: macOS 26.5.2 (build 25F84)
- Browser: Chromium 151.0.7922.34
- Also present in Quarto 99.9.9, built from `main` at `f9dd09cd2aee9f107a6c3f497e9210d55ceeebca`

## Quarto check output

```
Quarto 1.10.18
      Pandoc version 3.10.0: OK
      Dart Sass version 1.101.0: OK
      Deno version 2.7.14: OK
      Typst version 0.15.1: OK
      Version: 1.10.18
      TinyTeX: v2026.04
      VeraPDF: 1.28.2
      Using: TinyTex
      Version: 2026
      Version: 150.0.7871.115
      Version: 4.6.0
      Path: /Library/Frameworks/R.framework/Versions/4.6/Resources
      LibPaths:
        - /Users/charlottewickham/Library/R/arm64/4.6/library
        - /Library/Frameworks/R.framework/Versions/4.6/Resources/library
      knitr: 1.51
      rmarkdown: 2.31
      Version: 3.12.2
      Jupyter: 5.9.1
      Kernels: python3
```

---

AI helped with this investigation. The work was grounded in a local clone of `quarto-cli`, as described in CONTRIBUTING.md.

cwickham:
## Description

Closes #14774.

Three Quarto buttons carry Bootstrap's `btn` class with no `btn-*` variant class: the code tools button, a website's sidebar toggle, and a website's sidebar search button. Bootstrap's `.btn:focus-visible` rule sets `outline: 0` plus a box-shadow that only the variant classes define. On these buttons the box-shadow computes to `none`, so keyboard focus shows no indicator. This fails WCAG 2.2 SC 2.4.7 (Focus Visible).

This PR adds one rule that restores the browser's native focus ring on the three buttons with `outline: revert`.

Reasons to prefer the native ring over Bootstrap's box-shadow halo:

- It matches the indicator that the code copy button already has (#12118).
- An outline stays visible in forced-colors mode, where box-shadow computes to `none`.
- Bootstrap v6 moves button focus styles to `outline` (twbs/bootstrap#42062), so the fix does not depend on a mechanism that upstream is retiring.
- Browsers without `revert` support (Safari before 15.4) drop the declaration, and behavior is unchanged from today.

Site authors can still restyle the focus state, because user SCSS compiles after Quarto's rules. Each of the three selectors is a single class, the same specificity as Bootstrap's own rule. A plain override therefore wins on all three buttons. For example, `.quarto-search-button:focus-visible { outline: 3px dashed red; }` replaces the ring on the sidebar search button. I confirmed this in a rendered site in Chromium, Firefox, and WebKit.

No documentation change is needed: the fix adds no option and changes no documented behavior.

### Tests

The new spec `html-focus-indicator-bare-btn.spec.ts` covers all three buttons in Chromium, Firefox, and WebKit. It moves focus with real Tab presses (Option+Tab on WebKit, which ships Safari's default keyboard-navigation settings), then asserts that the computed `outline-style` is not `none`. Real key presses are necessary because programmatic focus does not match `:focus-visible`. All nine tests fail on a build without the fix and pass with it. The fixtures are a standalone `code-tools` document and a minimal website with a collapsed sidebar.

## Checklist

I have (if applicable):

- [x] added new tests
- [ ] created a separate documentation PR in [Quarto's website repo](https://github.com/quarto-dev/quarto-web/) and linked it to this PR — not needed, see above

<details>

- **Codebase grounding**: local clone of quarto-cli

</details>


src/resources/formats/html/bootstrap/_bootstrap-rules.scss
```diff
@@ -1038,6 +1038,16 @@ td code:not(.sourceCode) {
   padding: 0.7rem;
 }
 
+// These buttons are bare .btn elements (no btn-* variant class), so
+// Bootstrap's .btn:focus-visible leaves them with no focus indicator.
+// Restore the browser's ring; unlike box-shadow it survives forced colors.
+// See https://github.com/quarto-dev/quarto-cli/issues/14774
+.quarto-secondary-nav .quarto-btn-toggle:focus-visible,
+.quarto-secondary-nav .quarto-search-button:focus-visible,
```
cderv: Could we drop the `.quarto-secondary-nav ` prefix from these two selectors? It is the only reason the nav buttons end up with a different override story from the code tools button, and it does not appear to buy anything.

I tested this. Rendering a site with a user SCSS rule setting `outline: 3px dashed red` on all three selectors, the code tools button picks up the override (same specificity, user rules come later) while the two nav buttons keep the user-agent ring. With the prefix removed and nothing else changed, the override wins on all three. And without any user rule, the unprefixed selectors still beat Bootstrap: `outline-style` computes to `auto` on both nav buttons, because Quarto's rules layer already comes after the framework layer, so equal specificity plus later source order is enough.

Scoping is not lost either I think. Both buttons are emitted in exactly one place, `nav-before-body.ejs`, and only ever inside `.quarto-secondary-nav`.

That would leave all three buttons overridable the same way, and the description would no longer need the caveat about authors having to match the new specificity.

Did you check about this maybe ?

tests/docs/playwright/website/bare-btn-focus/_quarto.yml
```diff
@@ -0,0 +1,19 @@
+project:
```
cderv, line 1: This fixture needs a `.gitignore`, otherwise rendering it leaves an untracked `_site/` behind (plus the `.gitignore` Quarto generates for `.quarto/`). The driver calls `cleanoutput()` without `projectOutDir`, so it cannot clean a project render — which is why the neighbouring website fixture ships one:

https://github.com/quarto-dev/quarto-cli/blob/456e129742a31cb10298dea2c3fb693dd0c0504d/tests/docs/playwright/website/issue-14667/.gitignore

```
/.quarto/
/_site/
```

tests/integration/playwright/tests/html-focus-indicator-bare-btn.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+// the fix only applies to keyboard focus, and programmatic locator.focus()
+// does not reliably match :focus-visible. WebKit follows Safari's default
+// of skipping buttons on Tab; Option+Tab visits every focusable element.
+async function tabUntilFocused(
+  page: Page,
+  browserName: string,
+  target: Locator,
+  maxTabs = 25,
+): Promise<boolean> {
+  const tabKey = browserName === "webkit" ? "Alt+Tab" : "Tab";
```
cderv, line 22: Not a change request, just a note for the record: on the Windows WebKit build the three tests also pass with a plain `Tab`, so the Safari tab-order behaviour the comment describes does not reproduce there. `Alt+Tab` passes as well and CI runs on Linux, so I would leave this as is — worth knowing if it ever gets flaky.

cderv (commented):
Thanks for this, and for the very thorough issue that goes with it. I checked the claims in the description against the compiled CSS and a real browser rather than taking them on trust, and the diagnosis holds up in full: with the rule commented out the three buttons compute `outline-style: none`, `box-shadow: none`, and an empty `--bs-btn-focus-shadow-rgb`. I also ran the new spec locally in all three engines: nine pass with the fix and nine fail without it, so the assertion is doing real work.

One substantive request, on the `.quarto-secondary-nav` prefix, plus two small things. Details inline.

The rest checks out.

cwickham:
Thanks @cderv — both changes are in, as 92a59ef.

**Prefix dropped.** `quarto-btn-toggle` and `quarto-search-button` each appear in exactly one template, `nav-before-body.ejs` L66 and L86, both inside `<nav class="quarto-secondary-nav">`. Nothing else in `src/` emits them, so the prefix bought no scoping.

I rendered the same site twice, once per selector variant, with a user SCSS rule setting `outline: 3px dashed red`, and tabbed to each button in all three engines:

| | code tools | sidebar toggle | sidebar search |
|---|---|---|---|
| prefixed | `dashed 3px red` | `auto` (UA ring) | `auto` (UA ring) |
| unprefixed | `dashed 3px red` | `dashed 3px red` | `dashed 3px red` |

Identical in Chromium, Firefox, and WebKit. Your finding reproduces exactly.

I also checked the no-user-rule case against the fixtures: all nine button/engine combinations compute `outline-style: auto`, so equal specificity plus later source order is enough, as you said. The spec passes 9/9 on the unprefixed build.

The specificity caveat is out of the PR description.

**Fixture `.gitignore` added.**

cderv (approved):
LGTM, thanks;

🤖 went through the diff again after and flagged something worth a mention. #14826, already on main, added its own focus-visible rule for the new sidebar-item-toggle buttons:

https://github.com/quarto-dev/quarto-cli/blob/a8b5b2ba61b9d4053a8f8e5afeba640d2ce16180/src/resources/projects/website/navigation/quarto-nav.scss#L414-L418

That one uses `outline: 2px solid currentColor` with `outline-offset: 2px`, while this PR uses `outline: revert`. Same problem, same sidebar region, two different conventions now living next to each other.

Not blocking, both fix the underlying issue and neither is wrong. Do you think we need a follow-up issue to converge on one approach ?

I am merging this anyway, but I figured it was good to mention in case we want common ways to fix axe problems.

cwickham (commented):
Good catch, thanks — #14867 converges them on the browser's ring.

It turned out to be a deletion rather than a choice: nothing suppressed the
#14826 buttons' ring in the first place, so the rule can just go. Details and
measurements are in the PR.

Issue #14810, opened 2026-08-25
a11y: listing pagination aria-label is not localizable


cwickham:
The listing pagination nav has an `aria-label`, but it is hardcoded English: `aria-label="Page Navigation"` ([_pagination.ejs.md#L4](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/projects/website/listing/_pagination.ejs.md#L4)). On a non-English site the label stays English. The same template already uses the localization helper two lines above ([_pagination.ejs.md#L2](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/projects/website/listing/_pagination.ejs.md#L2), `listing.utilities.localizedString("listing-page-no-matches")`), so a language key is a drop-in fix.

Suggested fix: add a `listing-pagination-label` language key and use it via `listing.utilities.localizedString`. A default of "Pagination" also avoids a label collision with the prev/next page navigation ("Page navigation"), which appears on the same page when a site sets `page-navigation: true` (see #14376).

Split from #14376. Part of #8706.

Investigation was AI-assisted, grounded in a local clone (per CONTRIBUTING.md).

Issue #14809, opened 2026-08-25
a11y: dashboard header nav landmark has no accessible name


cwickham:
Every dashboard emits a `<nav class="navbar">` header, with no `aria-label`. The template includes it unconditionally ([template.html#L31](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/formats/dashboard/template.html#L31) → [title-block.html#L2](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/formats/dashboard/title-block.html#L2)), even when the dashboard has no title. For multi-page dashboards this nav is the primary navigation: the page tabs are injected into it ([format-dashboard-page.ts#L43](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/format/dashboard/format-dashboard-page.ts#L43)).

Two problems:

1. The nav has no accessible name. On a standalone dashboard this is a missing-name issue; when a dashboard page renders inside a website project, it can coexist with the website navbar and then fails axe `landmark-unique` (#14376).
2. The navbar toggler has a hardcoded English label, `aria-label="Toggle navigation"` ([title-block.html#L4](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/formats/dashboard/title-block.html#L4)). The website equivalent uses the localized `toggle-navigation` language key ([navtoggle.ejs#L2](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/projects/website/templates/navtoggle.ejs#L2)).

Both are one-line template fixes: Pandoc templates can read localized strings with `$quarto.language.<key>$` variables ([docs](https://quarto.org/docs/authoring/language.html#localized-strings-in-templates)). The mechanism applies to built-in templates too — verified by a render of this template with `lang: fr`, where `$quarto.language.toggle-navigation$` resolves to "Basculer la navigation".

Repro:

````qmd
---
title: "My Dashboard"
format: dashboard
---

# Page 1

Content

# Page 2

Content
````

The rendered header nav has no `aria-label`, and the toggler label stays English regardless of `lang`.

[...]

PR #14813, opened 2026-08-26
Give every website and book `nav` landmark a distinct, localizable label

### LINKED ISSUE #14376: a11y: navbar and sidebar nav landmarks have no aria-label (landmark-unique)
Reported in #14373.

Repro: https://github.com/alexpghayes/quarto-accessibility-reprex

axe rule: `landmark-unique` · impact: moderate · tag: best-practice
Targets: `<nav class="navbar navbar-expand-lg">`, `#quarto-sidebar`

The top navbar and the on-page sidebar are both `<nav>` landmarks without any `aria-label`. Assistive tech sees two unlabelled navigation regions and cannot distinguish them. Same for the right-hand `#quarto-margin-sidebar` if surfaced as a `<nav>`. Each navigation landmark should carry a distinguishing `aria-label` (e.g. `"Site"` for the top navbar, `"On this page"` for the page-local sidebar).

Part of #8706.

cwickham:
## Description

Closes #14376.

Website and book pages emit up to six `<nav>` landmarks. Before this PR, only the breadcrumbs had a label, and it was hardcoded English (`aria-label="breadcrumb"`). Assistive technology listed the rest as identical "navigation" entries, and axe flagged the pages under `landmark-unique`.

This PR gives each landmark a distinct label from a new set of language keys:

| Landmark | Label |
|---|---|
| Navbar | `navigation-main-label` ("Main") |
| Sidebar | Adaptive: "Main" when the site has no navbar, "Section" when it has one |
| Secondary/mobile nav strip | `navigation-secondary-label` ("Secondary") |
| Previous/next page nav | `navigation-page-label` ("Page") |
| Breadcrumbs (both copies) | `navigation-breadcrumbs-label` ("Breadcrumbs") |
| Table of contents | `aria-labelledby` on its existing localized heading ("On this page") — no new key |

Authors can override any key with `language:` metadata, per document or site-wide.

### Label wording

The default labels do not contain the word "navigation". Screen readers announce the landmark role after the label, so a label like "Main navigation" reads as "Main navigation, navigation". The [ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/) says not to use the role as part of the label, and the [WAI page-structure tutorial](https://www.w3.org/WAI/tutorials/page-structure/labels/) uses the bare label "Main" for a nav region. VoiceOver testing confirmed the double announcement with the longer wording.

### Design notes

- The two breadcrumb copies share one label. They hold identical links, and only one is in the accessibility tree at a time (one shows at `lg` and up, the other below). The APG also asks identical navigation sets to share a label.
- The secondary nav keeps a label distinct from the sidebar. Both are visible together when the mobile sidebar is open.
- The TOC label reuses the visible heading via `aria-labelledby`, which is the pattern the WAI tutorial recommends. The TOC nav can also hold the Other Formats / code-links / other-links sections. The label stays acceptable: everything in that nav is page-scoped.
- The TOC change also applies to `revealjs` (same partial pattern, `toc-slide.html`).

### Out of scope

- Dashboard header nav: #14809
- Listing pagination label: #14810
- Several `<a>` elements carry an invalid `role="navigation"` (secondary-nav toggles, sidebar section toggles). Unchanged here — now handled by #14826, which removes both and makes the sidebar section toggles real `<button>` disclosures. The two PRs are complementary: this one labels the genuine `<nav>` landmarks, #14826 deletes the fake ones. They conflict only in `news/changelog-1.11.md`, where both add an entry to `## Accessibility`; keep both.

### Documentation

No documentation PR: the only user-facing change is the new language keys, and the docs already refer users to `_language.yml` itself as the reference for available keys ([Document Language](https://quarto.org/docs/authoring/language.html)).

### Testing

- New smoke-all tests assert every label, the adaptive sidebar branch (with and without navbar), a `language:` override, and a custom `toc-title` in both `html` and `revealjs`.
- Verified axe `landmark-unique` passes on navbar-only, sidebar-only, navbar+sidebar, and book sites, at desktop and narrow viewports, including with the mobile sidebar open. Also reviewed the in-page report from `axe: output: document` on the rendered site. These browser checks were manual; this PR does not add a Playwright case for them.
- Verified the VoiceOver landmark rotor lists each landmark with a distinct name.

## Checklist

I have (if applicable):

- [x] added new tests

<details>

</details>


cderv (commented):
LGTM. I checked out the branch and ran both new smoke-all suites, they pass here on Windows. I also ran axe-core against the rendered fixture site in Chrome, `landmark-unique`, `landmark-one-main`, `region` and `duplicate-id-aria` are all clean at 1400x900 and at 500x900, including with the mobile sidebar expanded. The shared breadcrumbs label holds up, `.quarto-secondary-nav` is `display: none` from `lg` up and `.quarto-title-breadcrumbs` carries `d-none d-lg-block`, so only one of the two is ever in the accessibility tree.

One thing I noticed while looking at the TOC change. With `toc-location: left-body` or `right-body` we clone the whole TOC nav into the body and only rename the nav id:

https://github.com/quarto-dev/quarto-cli/blob/471531762d94e7badf0bd2e2f62461be07afc624/src/format/html/format-html-bootstrap.ts#L353-L362

The `<h2 id="toc-title">` inside is cloned as is, so the page ends up with two elements sharing that id, and now both navs point at it with `aria-labelledby`. The name resolves to the same title for both either way, and I measured `landmark-unique` already failing on that page before your change, so it is not a regression from this PR. Still, the ARIA reference now depends on a duplicated id. Would you rather fold a fix in here, or should I file it separately ?

Nit, no need to act on it: `language[kNavigationBreadcrumbsLabel] || "Breadcrumbs"` is the only language lookup in `website-navigation.ts` with a hardcoded fallback, the others rely on the resolved defaults.

## The new keys and the translation files

This part is not really about your change. The five new keys land in `_language.yml` only, so every translated site gets English landmark names sitting next to otherwise translated chrome. I looked at the wider picture and 25 of the 30 base language files are already behind `_language.yml`, so I filed #14822 for that separately.

For these five keys I produced a candidate set for all 30 base language files. To be explicit about what that is: it is LLM generated. I do not speak most of these languages and no native speaker has looked at any of it. What I verified is only the mechanics, `lang: fr` renders `aria-label="Principale"` and `lang: zh` renders `aria-label="主要"`, so the keys resolve and the encoding is fine. That says nothing about whether the wording is right.

Every row wants a review before it ships. The `de`, `is`, `tr`, `ko` and `eu` ones are the ones I trust least, `Main` and `Secondary` are bare adjectives and those languages want a head noun after them. The breadcrumbs column is the least settled everywhere, there is no single agreed term.

So it is your call. We could take this as a starting point here and be clear in the commit that the strings are machine generated and unreviewed, or leave the five keys English in this PR and let #14822 collect translations from people who actually speak the languages. Which do you prefer ?

<details>
<summary>Candidate translations, LLM generated and unreviewed (30 base languages)</summary>

These are LLM generated and have not been checked by a native speaker. The warning sign marks the rows I trust least, where "main" and "secondary" do not stand alone as adjectives.

| | main | section | secondary | page | breadcrumbs |
|---|---|---|---|---|---|
| bg | Основна | Раздел | Второстепенна | Страница | Път |
| ca | Principal | Secció | Secundària | Pàgina | Ruta |
| cs | Hlavní | Sekce | Sekundární | Stránka | Cesta |
| da | Primær | Sektion | Sekundær | Side | Brødkrummesti |
| de :warning: | Primär | Abschnitt | Sekundär | Seite | Brotkrümelpfad |
| el | Κύρια | Ενότητα | Δευτερεύουσα | Σελίδα | Διαδρομή |
| es | Principal | Sección | Secundaria | Página | Ruta |
| eu :warning: | Nagusia | Atala | Bigarren mailakoa | Orria | Ibilbidea |
| fi | Ensisijainen | Osio | Toissijainen | Sivu | Polku |
| fr | Principale | Section | Secondaire | Page | Fil d'Ariane |
| he | ראשי | מקטע | משני | עמוד | נתיב |
| id | Utama | Bagian | Sekunder | Halaman | Jejak |
| is :warning: | Aðal | Undirkafli | Auka | Síða | Slóð |
| it | Principale | Sezione | Secondaria | Pagina | Percorso |
| ja | メイン | セクション | サブ | ページ | パンくずリスト |
| ko :warning: | 주요 | 섹션 | 보조 | 페이지 | 이동 경로 |
| lt | Pagrindinė | Skiltis | Antrinė | Puslapis | Kelias |
| nb | Primær | Seksjon | Sekundær | Side | Brødsmulesti |
| nl | Primair | Sectie | Secundair | Pagina | Kruimelpad |
| nn | Primær | Bolk | Sekundær | Side | Brødsmulesti |
| pl | Główna | Sekcja | Dodatkowa | Strona | Ścieżka |
| pt | Principal | Seção | Secundária | Página | Trilha |
| ru | Основная | Раздел | Дополнительная | Страница | Путь |
| sk | Hlavná | Sekcia | Sekundárna | Stránka | Cesta |
| sl | Glavna | Razdelek | Drugotna | Stran | Pot |
| sr-Latn | Glavna | Odeljak | Sekundarna | Stranica | Putanja |
| sv | Primär | Sektion | Sekundär | Sida | Brödsmulor |
| tr :warning: | Ana | Bölüm | İkincil | Sayfa | Yol |
| ua | Основна | Розділ | Додаткова | Сторінка | Шлях |
| zh | 主要 | 章节 | 次要 | 页面 | 面包屑 |

All five avoid that language's own word for "navigation", following the rule you documented in `_language.yml`. That comment only lives in `_language.yml` though, so a translator editing `_language-de.yml` later will not see it.

</details>

cwickham:
@cderv, I haven't looked at all your comments, but since the language thing keeps coming up, I think it's worth a convention. I'd lean towards including and acknowledging LLM translations. That prevents the drift, and also exposes them to the people most likely to be able to do a better job at the translation.

cderv:
> I'd lean towards including and acknowledging LLM translations. That prevents the drift, and also exposes them to the people most likely to be able to do a better job at the translation.

Yes I agree on that.

cwickham:
Thanks for the careful review, and for running the axe pass on Windows.

Your warning about `de`, `is`, `tr`, `ko` and `eu` turned out to be the most
useful thing in the review. It is not a problem with those five rows. It is a
problem with the English strings, and it changed the design. `Main` and
`Secondary` are gone.

## The labels are now all nouns

| Landmark | Was | Now |
|---|---|---|
| Navbar | `navigation-main-label` "Main" | `navigation-site-label` "Site" |
| Sidebar | "Main" / "Section" | "Site" / "Section" |
| Toolbar | `navigation-secondary-label` "Secondary" | `navigation-toolbar-label` "Toolbar" |
| Previous/next | `navigation-page-label` "Page" | unchanged |
| Breadcrumbs | `navigation-breadcrumbs-label` | unchanged |

Bare adjectives do not survive translation. Wiktionary gives *main* as a bound
prefix in eight of our languages (`haupt-`, `aðal-`, `hoved-`, `pää-`), so it
cannot stand alone, and Docusaurus ships the identical string
(`theme.NavBar.navAriaLabel = "Main"`, 37 native-speaker locales) where 9
translators left it in English and 2 produced the bare stems `Aðal` and `Hoved`.
So `Primär` and `Aðal` were not weak rows to improve, and a noun removes the
problem in every language at once.

`Toolbar` also describes that strip more honestly than `Secondary` did. It holds
the sidebar toggle, the breadcrumbs and search, and no navigation link of its
own. Whether it needs to be a `nav` at all is a separate question, filed as
#14868.

## Translations

Based on the new nouns Claude found some nice sources:

| Key | Source |
|---|---|
| Section, Page | LibreOffice Writer `sw/messages.po`, all 33 locales |
| Toolbar | Wikidata Q371434, with GNOME `gtk` and `gedit` for 9 gaps |
| Site | Wikidata Q35127 |
| Breadcrumbs | Wikidata Q846205, or the word for "path" (below) |

Each line carries a comment naming its source. Of 165 cells, 134 are a verbatim
attested string. The other 31 are adaptations, marked in the file:

```
grep -rn 'needs review' src/resources/language/
```

You were right that breadcrumbs is the least settled column. 21 of those 31 are
that key. The rule is now explicit: use the attested breadcrumb term when it does
not contain that language's word for "navigation", otherwise use the word for
"path". That gives `Път`, `Cesta`, `Ścieżka`, `Шлях`, which is close to what you
proposed by hand. The difference is only that each value now has a citation.

To be clear about what is verified: no native speaker has read any of this. What
changed is the kind of claim. Each cell is now a sourced adaptation of a
professional localization, so a reviewer can check it against its citation
instead of judging it cold.

One question for you. In English `navigation-section-label` and
`crossref-sec-prefix` are both "Section", and 16 of the translations inherit
that, so the sidebar landmark and a cross-reference end up sharing a word. German,
Swedish and Slovak avoid it by accident, because LibreOffice's word for a Writer
section is an area word: `Bereich`, `Område`, `Oblasť`. I kept it, on the
grounds that the two never share a surface. A cross-reference is inline text with
a number inside a link, and the landmark label is only announced by assistive
technology, with the role after it. Do you read it the same way, or is one word
for both a problem?

## The TOC duplicate id

Filed as #14844 and fixed in #14853, which is stacked on this branch. The clone
now suffixes every id it copies, not just the `nav`. It was worse than one
heading: each `toc-<section>` link was duplicated too, so the count grows with
the size of the TOC. You were right that it predates this PR. The duplicates were
inert until #14376 pointed `aria-labelledby` at `toc-title`.

## The nit

Fixed. `makeBreadCrumbs` now reads `language[kNavigationBreadcrumbsLabel]!` like
the other lookups in that file.

cwickham:
@cderv, one question from merging main: #14773 added 60 language keys to the schema and allowed null for all of them. I initially, put my five `navigation-*-label` keys in that same shape, but then set them back to `string`.

For these keys, a null would result in `aria-label=""`, an unnamed landmark, which is exactly what we are trying to fix in this PR. I don't fully understand the rationale over there, so maybe there is a good reason. Let me know and I can add it.

cderv:
Good catch, and I looked into it. It's not really a pre-existing contract, more schema drift that #14773 ended up formalizing.

The 60 keys #14773 promoted to `anyOf: [string, null]` got that shape only because they were missing from the schema entirely before, an open object, so anything passed through unvalidated. The other ~70 shipped keys have always been plain `string:`, rejecting null, with nothing reported against them. `toc-title-document` sits right next to one of the promoted keys as an example:

https://github.com/quarto-dev/quarto-cli/blob/e9cebc95d7b73e1a35db5ba401e60d1b7aa6c3bf/src/resources/schema/definitions.yml#L1375-L1382

The commit that added the null allowance says the motivation was tolerating a YAML typo, a key written with nothing after the colon parses as `null` and was failing validation:

https://github.com/quarto-dev/quarto-cli/commit/7a15ed8e53dd19c51799fae7953eea420c5cb892

That trades one failure mode for a worse one though, the label renders empty instead of falling back to the default, same symptom this PR exists to fix. Nothing in `_language.yml` or the test suite relies on null clearing a shipped default. `translationsForLang` already passes any `null`-valued key straight through regardless of type, so the schema change didn't even add expressive power:

https://github.com/quarto-dev/quarto-cli/blob/e9cebc95d7b73e1a35db5ba401e60d1b7aa6c3bf/src/core/language.ts#L166-L181

So keeping your five keys `string` is the right call here, it's the actual fix for this class of bug, not scope creep. I'll open a follow-up to tighten the 60 keys back to string-only and document `""` as the clear idiom instead of null, unless you see a use case for null clearing defaults that I'm missing.

cderv (approved):

PR #14815, opened 2026-08-27
Add `quarto call axe`, a hidden experimental site accessibility scanner

cwickham:
## Description

This PR adds `quarto call axe <site-dir>` — a hidden, experimental
command that scans an already-rendered Quarto site for accessibility
violations with quarto-cli's vendored axe-core.

```sh
quarto render
quarto call axe _site
```

What it does:

- Serves the site dir locally and drives headless Chrome (found the same way
  as other Quarto features: `quarto install chrome-headless-shell`, or a
  system Chrome/Edge) over a **page × viewport × color-mode matrix**. Each
  page's modes are discovered from its rendered HTML, so a light-only site
  scans once per page and a `light:`/`dark:` site scans both slots.
- Injects the vendored axe build at scan time — any rendered site scans
  as-is, offline, on the same axe version the render-time `axe:` option uses.
- Groups violations by **root-cause signature** (a normalized selector, or
  the color pair for `color-contrast`), so one template defect on 40 pages is
  one finding with a count, and its id is stable across runs.
- Reconciles a hand-written, committed **baseline** (`_axe-baseline.json`):
  new findings alert, accepted ones are counted but out of the way,
  escalations and out-of-scope recurrences re-alert.
- Writes `_axe-checks/findings.json` (machine contract), `report.md`
  (GitHub-flavored markdown), and a generated `README.md` that documents the
  artifacts for whoever — or whatever agent — finds them in a repo.
- **Fails closed**: a cell that times out, errors, or gets redirected is an
  infrastructure failure in the output and the exit code, never a pass.
  Exit codes: `0` complete, `1` new findings at/above `--fail-on <impact>`
  (opt-in, for CI), `2` incomplete — and 2 takes precedence over 1.

Why hidden under `call`: the semantics (signature scheme, baseline
grain, flag surface) should be validated by real use before becoming public command. Nothing here is a public API commitment.

Docs in this PR:

- `dev-docs/axe-scan.md` — how to use it (flags, baseline workflow, CI
  recipe). Will eventually seed the `quarto-web` docs.
- `llm-docs/axe-scan-architecture.md` — how it works and why (pipeline, mode
  discovery, signatures, baseline semantics, the raw-CDP decision).

Deliberately not in scope (deferred to the public command): `_axe.yml`
config, interaction states, ruleset scoping, source mapping (`.qmd` →
output), a rich report extension. The render-time `axe:` option is
untouched; the scanner mirrors its conformance labellers, pinned by a parity
test.

Testing: 9 unit test files (115 tests) run the aggregate/baseline/discovery/
config logic browser-free against captured real axe payloads; 5 smoke test
files run the full command over fixture sites in `tests/docs/axe-scan/`,
including subprocess exit-code tests covering `--fail-on`, fail-closed
timeouts, and a browser that cannot start. CI already installs
`chrome-headless-shell` unconditionally; a manual `test-smokes-parallel`
dispatch on this branch is green:
<https://github.com/quarto-dev/quarto-cli/actions/runs/33106781443>.

## Checklist

I have (if applicable):

- [x] added new tests
      (not applicable yet: the command is hidden; docs live in `dev-docs/`
      and graduate to quarto-web when it goes public)

<details>

</details>


news/changelog-1.11.md
```diff
@@ -7,6 +7,7 @@ All changes included in 1.11:
 ## Accessibility
 
 - ([#14615](https://github.com/quarto-dev/quarto-cli/issues/14615)): Fix invalid `role="menu"` on the website navbar's collapse toggle button, flagged by axe-core (`aria-allowed-role`) and WAVE (`aria_menu_broken`) when the navbar collapses to the hamburger at narrow viewports.
+- ([#14815](https://github.com/quarto-dev/quarto-cli/pull/14815)): Add `quarto call axe`, a hidden experimental command that scans a rendered site for accessibility violations with axe-core across a page × viewport × color-mode matrix, groups them by root-cause signature, reconciles a committed baseline, and can gate CI with `--fail-on`. See `dev-docs/axe-scan.md`.
```
cderv: I think to be useful this should be a url
```
See `dev-docs/axe-scan.md`
```

But maybe better to have it as a quarto-web page ?

src/command/call/axe/cmd.ts
```diff
@@ -0,0 +1,489 @@
...
+}
+
+export const axeCommand = new Command()
+  .name("axe")
+  .hidden()
+  .arguments("<site-dir:string>")
+  .description(
+    "Scan a rendered site for accessibility violations with axe-core.\n\n" +
+      "EXPERIMENTAL: flags, artifacts and semantics can change between " +
+      "prereleases (see dev-docs/axe-scan.md in the quarto-cli repo).\n\n" +
```
cderv: `see dev-docs/axe-scan.md` => maybe this would be better as an only doc. 

Otherwise, we may need a url so that this is useful

dev-docs/axe-scan.md
```diff
@@ -0,0 +1,192 @@
...
+### Reading report.md
+
+The report is GitHub-flavored markdown, made to be read where markdown
+already renders: your editor's preview, GitHub, or the sticky PR comment
+from the CI recipe below. Rendering it is optional. For a standalone HTML
+view:
+
+```sh
+quarto render _axe-checks/report.md
+```
```
cderv, line 58: Did you consider also providing the HTML rendered ? 

Or is first intent for users to read the .md first ? If so, we need to find a way to make the .md readable without rendering. Like tables formatting, and other formatting ? 

Maybe later improvemend. But I wondered why not calling `quarto render` on the .md (maybe in a try-catch in case of error for some reason).

dev-docs/axe-scan.md
```diff
@@ -0,0 +1,192 @@
...
+
+then open `_axe-checks/report.html` in your browser. The output lands
+beside the report, inside the self-ignoring artifact directory.
+
+`quarto preview _axe-checks/report.md` does **not** work from inside a
+project: `_`-prefixed directories are not project inputs, so preview stops
+with `No output created by quarto render report.md`. Use render-then-open,
+or pass `--report` a path inside your site source (see the flags table) to
+render and preview the report as part of the site.
```
cderv, line 67: I wonder if one way is to make `_axe-checks` its own quarto project. 

This way `quarto render` or `quarto preview` on that directory maybe use this nested project. 

Though this may not be possible until https://github.com/quarto-dev/quarto-cli/issues/3186

dev-docs/axe-scan.md
```diff
@@ -0,0 +1,192 @@
...
+## Running it as a post-render step
+
+The command works as a project `post-render` script, so a full render scans
+itself:
+
+```yaml
+project:
+  type: website
+  post-render:
+    - quarto call axe _site/
```
cderv, line 109: My idea on this is that I wonder if we can make `quarto call axe` self aware to handle correctly all case 

- Detect output dir to check in `QUARTO_PROJECT_OUTPUT_DIR`
- Maybe only run on `QUARTO_PROJECT_RENDER_ALL` set so avoid running on incremental render ? 
- Hide any progress depending on `QUARTO_PROJECT_SCRIPT_PROGRESS` ? 
- Respect quiet when `QUARTO_PROJECT_SCRIPT_QUIET` set ? 

https://quarto.org/docs/projects/scripts.html#pre-and-post-render

dev-docs/axe-scan.md
```diff
@@ -0,0 +1,192 @@
...
+| `1` | complete scan, new findings at/above the `--fail-on` threshold |
+| `2` | scan incomplete — a cell timed out or errored, no browser, nothing to scan. Takes precedence over 1: an incomplete scan never reads as a pass |
+
+A minimal GitHub Actions gate:
+
+```yaml
+- uses: quarto-dev/quarto-actions/setup@v2
+- run: quarto install chrome-headless-shell --no-prompt
+- run: quarto render
+- run: quarto call axe _site --fail-on serious
```
cderv, line 162: idea: we should add that as a parameter in our quarto render github action

dev-docs/axe-scan.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+`report.md` is GitHub-flavored markdown, so posting it as a PR comment is
+workflow configuration, not tooling. A sticky comment beats a plain one:
+it updates in place on each push instead of stacking a new comment per run:
```
cderv: is there anything missing here ? or just did not get the meaning ?

dev-docs/axe-scan.md
```diff
@@ -0,0 +1,192 @@
...
+site can exceed it — trim or attach as an artifact instead.)
+
+## Reading the output as an agent
+
+Point the agent at `_axe-checks/` and let it read the generated `README.md`
+first. Finding ids are stable across runs (`image-alt-6e3b76`), so "fix
+`image-alt-6e3b76`" is a well-defined instruction, and each finding's
+`occurrences[]` carries real selectors and HTML excerpts. Fixes belong in
+Quarto *source* (`.qmd`, `_quarto.yml`, `_brand.yml`, theme `.scss`) — never
+in the rendered site directory.
```
cderv, line 189: Do you think we'll need to provide a skill to help with this ?

llm-docs/axe-scan-architecture.md
```diff
@@ -0,0 +1,289 @@
...
+
+`quarto call axe <site-dir>` scans an already-rendered site for
+accessibility violations. It serves the site dir locally, drives headless
+Chrome over raw CDP, and runs quarto-cli's vendored axe-core against every
+page × viewport × colour-mode cell. It groups violations by root-cause
+signature, reconciles them against a hand-written baseline, and writes
+`findings.json`, `report.md`, and a generated `README.md`.
+
+The command is a hidden prototype under `call` (hidden while experimental —
+it does not show in `quarto call` help). It validates the
```
cderv, line 26: We could also mark it as EXPERIMENTAL in the `quarto call --help`

I am fine with both.

src/command/call/axe/scan.ts
```diff
@@ -0,0 +1,873 @@
...
+      // probing the next page mid-load.
+      await Promise.race([reset.event, sleep(1000)]);
+    } catch (_e) {
+      // if even that fails the next cell will report the real problem
+    }
+    reset.cancel();
+    result = cell("timeout", {
+      message: `cell exceeded --timeout of ${config.timeout}ms`,
+    });
+  }
```
cderv, line 801: if one page is slow, or something goes wrong talking to the browser, the whole scan could freeze on that one page instead of giving up on it and moving on. That breaks the tool's core promise: it's supposed to always finish and give a clear result, never hang.

on timeout we cancel `load` and reset the tab, but the original scan attempt (line 577) is only abandoned, not actually stopped — it can keep running in the background and interfere with the next page's scan. The reset itself (`Page.navigate` to `about:blank`) also has no time limit, so if the browser connection is the thing that's actually stuck, this line never returns either.

Might be worth putting a time limit on the reset too ? Curious if this was tested against a truly frozen browser connection, not just a slow page. It may be hard to do.

src/command/call/axe/aggregate.ts
```diff
@@ -0,0 +1,535 @@
...
+          `Consider combining them into one entry.`,
+      );
+      existing.siteWide = existing.siteWide || entry.pages.length === 0;
+      for (const page of entry.pages) {
+        existing.pages.add(page);
+      }
+      existing.impact = worstImpact(existing.impact, entry.impact);
+      existing.notes.push(entry.note);
+      existing.entryIds.push(entry.id);
+      existing.rules.push(entry.rule);
```
cderv: if the same type of issue gets accepted as "known/OK" on two different pages, one marked less serious and one marked more serious, a genuinely new and worse problem on the "less serious" page can slip through silently instead of being flagged.

two baseline entries for the same signature get merged into one, using the *most lenient* combination (worst accepted impact, applied to every page in the merge). So the stricter page inherits the looser page's tolerance.

Since one entry per signature is already the intended shape (per the warning already in the code), would rejecting duplicates outright be simpler than merging them ?

Maybe I am missing something. Just curious

src/command/call/axe/discover.ts
```diff
@@ -0,0 +1,253 @@
...
+  // pages that remain.
+  if (config.exclude) {
+    const patterns = compile(config.exclude);
+    selected = selected.filter((path) =>
+      !patterns.some((pattern) => pattern.test(path))
+    );
+  }
+  if (config.maxPages !== undefined) {
+    selected = selected.slice(0, config.maxPages);
+  }
```
cderv: `--max-pages` counts before filtering out internal redirect placeholders, so can `--max-pages 1` can report "nothing to scan" even when the site has real pages ? 
just because a redirect placeholder happened to sort first. Worth a quick fix if so, but really not that important.

src/core/cri/cri.ts
cderv, line 1: So we don't use our own CRI for this new feature ?

src/command/call/axe/cmd.ts
```diff
@@ -0,0 +1,489 @@
...
+    "quarto call axe _site",
+  )
+  .example(
+    "Scan two pages, desktop light only",
+    "quarto call axe _site --pages index.html,about.html " +
+      "--viewports 1440x900 --themes light",
+  )
+  // deno-lint-ignore no-explicit-any
+  .action(async (options: any, siteDir: string) => {
+    const code = await axeScan(axeScanConfig(options, siteDir));
```
cderv: a simple typo in a command flag (e.g. `--fail-on serius` instead of `serious`) currently produces the exact same failure signal as "we found a real accessibility problem." 
Maybe we can do better ?  

every quarto command exits 1 for "something went wrong" (bad flag, crash, etc.) via `quarto.ts`'s root-level `.throwErrors()` + catch-all, that's a global convention and fine everywhere else. `axe` is the one command that also gives exit 1 a second meaning ("real findings at/above `--fail-on`"), as part of its own 0/1/2 contract. `axeScanConfig()` runs before `axeScan`'s own try/catch, so a bad flag value falls straight into quarto's generic exit-1 path instead of axe's own incomplete/findings codes, colliding with the meaning axe itself defined.

Not a Cliffy issue, and not fixable by typing `--fail-on` as a Cliffy enum either: quarto's `.throwErrors()` would still normalize it to exit 1 either way. The fix has to live in axe's own action handler, catching the config error there and mapping it to a third code before quarto's generic handler ever sees it.

Worth giving option-parsing errors their own exit code, distinct from both 1 and 2 ? 
I think we can give classes to error for custom handling.

cderv (commented):
First round of review for our 📲

src/core/cri/cri.ts
cwickham: Right — we read `cri.ts` for this and decided against reusing it. The
clearest way to explain is to split "a Chrome driver" into three layers.
The answer is different at each layer.

1. **Launcher** — start the Chrome process: flags, profile dir, wait for
   the CDP port, kill on exit.
2. **Transport** — a generic pipe to the running Chrome. It sends any CDP
   command, awaits the result, subscribes to events, and closes. It knows
   nothing about what it carries.
3. **Task logic** — which commands to send, in what order, with what
   recovery. For mermaid this is the `cri.ts` facade (open /
   querySelector / screenshot). For the scanner it is `scanCell`:
   navigate, viewport and media emulation, seed the color mode before
   parse, evaluate axe with `awaitPromise`.

`cri.ts` bundles all three layers, and its public surface is layer 3,
shaped for mermaid. None of the scanner's commands are reachable through
that facade.

The decisive requirement is fail-closed cells: a hung `axe.run()` must
fail its one cell within `--timeout`, and the scan must continue. That
needs a cancellation surface from layer 2. `cri.ts` has that surface
(in `deno-cri`) but does not expose it.
So the scanner carries a small typed transport of its own (`CdpClient` in
`scan.ts`: send / cancellable wait / close, ~150 lines). Layer 1 got
duplicated as a side effect; browser discovery was shared from day one
(`getBrowserExecutablePath()`).

On not maintaining this in two places — agreed. Proposal: one follow-up
PR that removes all the duplicated code, in two parts.

**Layer 1**: extract one `launchChrome(options)` in `src/core/cri/`,
called by both `criClient` and the scanner's `launchScanBrowser`. 

**Layer 2**: retarget `CdpClient`'s internals onto `deno-cri`, keeping
the small typed interface the unit tests stub. After both parts, no
protocol or launcher code exists twice. The cleaner end state — `cri.ts`
*exports* a typed transport and the scanner consumes it (future work when the scan's needs settle).

**Layer 3 never merges**: mermaid's commands and the scanner's commands
are different jobs.

cderv: OK that makes sense I understand. Let's look at https://github.com/quarto-dev/quarto-cli/pull/14835 to see the proposal. 

@cwickham I believe we can leverage stacked PR. Github UI may propose it to you. 

This will allow to work on both branches separately, but merge all at once easily, and have both branches synced together. 

See here: https://github.github.com/gh-stack/

There is even some skills available to have help for 🤖

cwickham:
Thanks for this — a good first round. Every thread has a reply. Seven
commits on the branch answer it:

| commit | what it does |
|---|---|
| `b0cc18330` | bounds the timeout recovery, and stops the cell it abandoned |
| `f268223cf` | rejects duplicate baseline signatures instead of merging them |
| `55b113e75` | skips the scan on an incremental project render |
| `57ce3228d` | gives usage errors their own exit code, 3 |
| `6fe3e54e3` | `--max-pages` counts scannable pages, not raw HTML files |
| `40b7616dc` | changelog pointer becomes a link; `--help` drops it |
| `be7698477` | rewords the sticky-comment paragraph you flagged |

All seven passed roborev, and the axe suite is green: 162 unit tests and the
four browser smoke tests. Thanks for pointing me at roborev — reviewing my
own commits one at a time before anyone else reads them has earned its keep
already.

Four of your ideas I have kept as follow-ups rather than folding into this
PR, each with the reasoning in its thread: rendering `report.md` to HTML,
respecting the quiet and progress script variables, an axe parameter for
quarto-actions, and a skill for the agent workflow. They are tracked on my
side, and none of them changes this PR's surface.

One question back, in the `post-render` thread: there is no documented way
for a project script to know its phase, so the new incremental-render guard
infers it. Worth an issue?

news/changelog-1.11.md
cwickham: Yes, for now, I've made changelog entry a real link
(`40b7616dc`):

> See [dev-docs/axe-scan.md](https://github.com/quarto-dev/quarto-cli/blob/main/dev-docs/axe-scan.md).

A quarto-web page is the better answer, so I will open a PR against
quarto-web's `prerelease` branch with a page for the command. 
I'll note there once it is live to update the link in the changelog.

src/command/call/axe/cmd.ts
cwickham: Dropped the pointer from the help text entirely (`40b7616dc`) — it now says
only that the command is experimental. A repo path helps nobody at a
terminal, and a full URL is a lot of characters to carry in `--help` for
something that changes once the quarto-web page exists.

When that page lands I can add its URL back here, in the shape `quarto run`
uses (`src/command/run/run.ts:52-56`): a bare URL on its own line after
"For details, see:".

dev-docs/axe-scan.md
cwickham: This is a good idea. Rendering to HTML was the first thing I did every time. 
I've captured this as a strand in my local braid for now.

dev-docs/axe-scan.md
cwickham: You never edit `report.md`. Every scan overwrites it. So there is no edit
loop for preview to serve — the want is "show me this rendered", not
"show me this as I change it". If we give people HTML, I think the preview question goes away.

dev-docs/axe-scan.md
cwickham: Good list. Split three ways.

**`QUARTO_PROJECT_RENDER_ALL` — fixed, `55b113e75`.** A post-render scan now skips incremental renders with a note and exit 0. 

One thing I want your opinion on. To skip only in that case, the command has
to know it is a project script at all — `RENDER_ALL` is equally unset at a
plain command line, so reading it alone would make the command a no-op
everywhere except inside a full render. There is no documented way to know
the phase. I infer it from `QUARTO_PROJECT_OUTPUT_DIR` being present. A documented indicator — `QUARTO_PROJECT_SCRIPT_PHASE=pre|post`, or similar — would make this a fact instead of an inference. Is that worth an issue? 

 Side note: the comment at
`src/command/render/project.ts:830` says "run post-render if this isn't
incremental", but the only guard there is `if (!projResults.error)`. That comment seems stale.

**Output dir — already on my list** as "default the site dir from the
project". Your route via the env var is cheaper than the one I had written
down, and I have noted it there.

**Quiet and progress — captured as a strand.** Partly reachable today:
`--quiet` is a global option on every quarto command, and every per-cell
line goes through `info()`, so `post-render: quarto call axe _site --quiet`
works. The strand is to stop making people write it.

dev-docs/axe-scan.md
cwickham: This is a good idea once we've got a bit more experience running it in CI.
Captured as a strand on my side. Happy to open an issue on quarto-actions
when it is time, so it lives where you'd act on it.

dev-docs/axe-scan.md
cwickham: Yeah, this was confusing. Reworded in `be7698477`:

> `report.md` is GitHub-flavored markdown, so a workflow can post it straight
> into the PR — no Quarto-side tooling needed. Use a sticky comment, which
> updates in place on each push rather than adding one comment per run.

Tell me if that was not the part that tripped you.

dev-docs/axe-scan.md
cwickham: Probably, but not yet. Captured as a strand.

The generated `_axe-checks/README.md` is the shipped answer for now. It
travels with the artifacts, it is regenerated every scan so it always matches
the build that wrote it, and it helps anyone who ran a scan — where a repo
skill only helps people working inside quarto-cli.

The reason to wait is that what the skill needs to say can only come from
running the fix loop with an LLM *without* a skill first. Written now it
would be a guess.

llm-docs/axe-scan-architecture.md
cwickham: Let's leave hidden, the help text says EXPERIMENTAL.

src/command/call/axe/scan.ts
cwickham: Both holes were real. Fixed in `b0cc18330`, and no, it had not been tested
against a frozen connection — it is now.

**The unbounded reset.** The load-event wait was capped, but the
`Page.navigate` send was not, and neither was the seed removal after it. If
the transport is what wedged, that send never returns, and the scan stops
finishing cells: a hang, not a timeout — the one thing `--timeout` exists to
prevent. Both recovery commands now go through a `sendWithin` helper that
gives up after a second and never rejects. A failure there is not worth
reporting, because the next cell hits the same dead connection and says so.

**The abandoned run.** Its in-flight round-trip still completes, and then its *next* step runs against the page the following cell has navigated to — a stale `Runtime.evaluate` injecting axe and starting a second scan underneath it. Sends from the run body now go through a wrapper that refuses once the cell is abandoned, so it stops at its next step.

**Testing a frozen connection.** Easier than expected, because the unit
tests already stub the transport: a stub whose `send` never resolves is a
wedged connection, deterministically. Two tests in
`tests/unit/axe-scan-cell.test.ts` pin it. I ran them against the old code to
make sure they were not vacuous — the wedged-transport one hangs until the
harness's two-minute guard, which is your failure exactly.

src/command/call/axe/aggregate.ts
cwickham: Yes. Done in `f268223cf`, and thank you — this was worse than the warning
suggested.

Your reading is right, and it is loose in both dimensions at once. The merge
unioned the pages *and* took `worstImpact`. Acceptance records the impact it
was written at and re-alerts on escalation past it, so "most severe" is the
lenient direction: a `minor` acceptance on one page beside a `critical` one
on another became "critical, everywhere", and the stricter page inherited the
looser page's tolerance in scope and impact both.

What I had not appreciated is that the two entries can disagree honestly, so
this is not a copy-paste slip worth absorbing:

- axe assigns impact per *check*, not per rule. In the vendored axe-core
  4.10.3, 16 of the rules I could parse have checks with differing impacts.
  `aria-roles` is the clean case: an invalid `role` is critical, an abstract
  one is serious.
- A finding's impact is the worst across the occurrences in *that* scan
  (`aggregate.ts:212`), so a `--pages` scan and a full scan can report the
  same signature at different impacts.

`parseBaseline` now reports each repeat, naming the entry that already
accepted the signature. It reads the raw data rather than the parsed result,
so a duplicate and a typo both surface in one pass — same reason the entry
schema is `.strict()` rather than a `superRefine`. `acceptances()` builds the
map straight and treats a duplicate as an internal error.

src/command/call/axe/discover.ts
cwickham: It can, exactly as you describe. Fixed in `6fe3e54e3`.

The walk and the classification are now one pass that stops once the cap
holds that many real pages, so `--max-pages n` means n scannable pages. One
consequence worth naming: a capped scan's redirect ledger now only covers the
paths walked before the cap filled. That reads consistently with the rest of
a subset scan, whose counts already describe the subset and not the site.

src/core/cri/cri.ts
cwickham: One heads-up: the timeout work above (`b0cc18330`) touches the same code
#14835 rewrites onto `deno-cri`. I fixed it here rather than there because
this PR has to be correct standing alone — "a cell always finishes" should
not wait on a refactor. I will merge down and re-express it against the new
transport.

src/command/call/axe/cmd.ts
cwickham: Yes. Exit 3 now means usage error, in `57ce3228d`. Your diagnosis was exact,
including that the fix has to live in the action before quarto's root handler
sees the error.

Chasing it turned up a second inconsistency on the same theme. `--themes`
matching no cells also throws an option error, but from *inside* `axeScan`,
where it was caught and returned exit 2. So one kind of typo read as "new
findings" and another read as "the scan did not finish". Both throw sites now
raise the same named error and land on 3, and the docs table has a fourth
row.

cwickham:
@cderv round one is fully addressed — every thread has a reply, and six of them have a commit behind it.

Since those replies I also merged `main` in: it had moved 15 commits and collided on the changelog. CI is green on the merge commit, including the full parallel smoke matrix on Ubuntu and Windows. I re-ran the axe suite locally after the merge too, because `main` brought markup changes the smoke tests scan (the skip-to-main-content link, sidebar toggles as buttons) — 166 tests, all passing.

Ready for another look. If it looks good to you, I think it is ready to merge.

Two things that follow this PR rather than block it:

- #14835 stacks on this branch, so it needs a merge-down once this is in. That is also where the timeout fix gets re-expressed against the `deno-cri` transport.
- The quarto-web page for the command comes separately, against the `prerelease` branch. When it is live I will point the changelog link at it and put the URL back in `--help`.

One question is still open, in the `post-render` thread: there is no documented way for a project script to know whether it is running pre- or post-render, so the new incremental-render guard infers it from `QUARTO_PROJECT_OUTPUT_DIR`. Worth an issue?

cderv:
Thanks, this is in good shape. I did a second pass over the branch before merging and found two things worth fixing, so rather than open more threads I pushed the commits straight to the branch. Revert either if you disagree.

## `--report` can overwrite the baseline (`1741f3b38`)

`--report _axe-baseline.json` runs a clean scan and then destroys the ledger it just read. The baseline is read before the scan and the report is written after it, with no check on where it lands, so the scan completes, reports correctly, and replaces the file with markdown on its way out. `findings.json`, `README.md` and `.gitignore` go the same way.

A destination the scan owns is now a usage error before any work starts, on the exit code 3 you added in `57ce3228d`. The comparison folds case on Windows and macOS, otherwise `--report _AXE-BASELINE.JSON` walks past the check on exactly the filesystems where it matters.

I tried `Deno.realPathSync` for that comparison first and dropped it. It corrected the casing in one temp directory and left it alone in another, in the same run on the same machine, which is not something to back a destructive check with. The reasoning is in the `pathIdentity` comment so nobody tries it again.

## A stale baseline entry is not always resolved (`b9c685222`)

`report.md` said an unseen entry is resolved "if this was a full-site scan". That covers `--pages`, but a `--themes light` scan never runs a dark cell, a narrowed `--viewports` never runs the 320px one, and a cell that failed closed inside an otherwise full run is the same hole. In all of those the report was inviting someone to prune an acceptance that nothing had looked for.

`staleIsConclusive` now requires every axis full and every cell completed, and `report.md` and the console say which of the two cases they are in. Coverage is membership rather than count, so `--themes light,light` still reads as narrowed and a superset like `--viewports 1440x900,320x568,768x1024` does not. Same shape as the duplicate-signature one, the ledger is hand-written and the code should not degrade it quietly.

Three smaller things ride along in that commit:

- `parsePositiveInt` accepted a fraction, so `--max-pages 1.5` compared with `===` against a running page count and the cap silently never applied. `Number.isInteger` now. Same failure as the redirect-stub counting in `6fe3e54e3`, different input.
- `--report` is recorded in `findings.json` so the README's regenerate command reproduces the destination instead of reverting to the default. Nullish and additive, the scheme does not move.
- `sniffRedirectStub` matches its script markers inside `<script>` only. The near-empty-body gate already made that unreachable in practice, so it is hardening rather than a fix. Revert it if you would rather keep the function simple.

## What this does not touch

No new flags, no new artifacts, no scheme bump. Nothing in `scan.ts`, so #14835 should merge down without meeting any of it.

Ran the four touched unit files with `./run-tests.sh`, 35 passing.

Does the stale-entry wording read right to you ? That one changes what `report.md` tells people to do, so it is the one I would most like your eye on.

dev-docs/axe-scan.md
cderv: It is a documented fact today, just split across two variables instead of one: `QUARTO_PROJECT_INPUT_FILES` is pre-only and `QUARTO_PROJECT_OUTPUT_FILES` is post-only.
https://github.com/quarto-dev/quarto-cli/blob/e9cebc95d7b73e1a35db5ba401e60d1b7aa6c3bf/src/command/render/project.ts#L330
https://github.com/quarto-dev/quarto-cli/blob/e9cebc95d7b73e1a35db5ba401e60d1b7aa6c3bf/src/command/render/project.ts#L850
`OUTPUT_DIR` is the only variable that is genuinely phase-symmetric (both scripts get it), which is why it does not answer your question on its own.

The catch is the #10828 escape hatch: when `QUARTO_USE_FILE_FOR_PROJECT_OUTPUT_FILES` (or the input-side twin) is set, the variable goes missing and the file list is written to disk instead, so a phase check built on presence/absence of these two breaks for anyone using that path.

I do not think a new `QUARTO_PROJECT_SCRIPT_PHASE` pulls its weight given that. It would be a permanent public surface for something already derivable in the ordinary case, for one hidden experimental command. I opened #10314 with a link back here, since that thread is asking for the same kind of missing invocation context (preview vs. render, in that case) and is a better place for a phase variable to land if we ever want one.

Separately: the comment at
https://github.com/quarto-dev/quarto-cli/blob/e9cebc95d7b73e1a35db5ba401e60d1b7aa6c3bf/src/command/render/project.ts#L831
says post-render is skipped on an incremental render. The only guard around it is `if (!projResults.error)`, so that is stale, `behavior.incremental` never gates the scripts. Confirmed by reading, not something I have changed here.

Issue #14655, opened 2026-07-07
a11y: `code-line-numbers` emits per-line links with no accessible name


cwickham:
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

With `code-line-numbers` enabled for `format: html`, each code line is emitted as
an empty `<a href="#cbN-n">` (the line-number anchor, where the number is rendered
via a CSS `::before` counter). Running with axe (`axe: output: document`) reports:

> Serious · WCAG 2.0 A (2.4.4, 4.1.2): Ensure links have discernible text
> Links must have discernible text
>   a[href$="#cb2-1"]
>   a[href$="#cb2-2"]

This is closely related to jgm/pandoc#1030, but that covers the case when line-numbers are **off**.

The per-line anchors are produced upstream by [`jgm/skylighting`](https://github.com/jgm/skylighting), so the fix, if there is one, belongs upstream — filing here to track it from Quarto.

> Investigation was AI-assisted (reproduced and inspected the rendered output with
> Quarto 1.10.12), per CONTRIBUTING.md's note on using AI tools.

### Steps to reproduce

````qmd
---
title: "Line numbers"
format:
  html:
    code-line-numbers: true
    axe:
      output: document
---

```r
x <- 1
y <- 2
```
````

### Actual behavior

Axe reports:

> Serious · WCAG 2.0 A (2.4.4, 4.1.2): Ensure links have discernible text
> Links must have discernible text
>   a[href$="#cb2-1"]
>   a[href$="#cb2-2"]

### Expected behavior

No violations.

### Your environment

- IDE: n/a — reproduced with the Quarto CLI from the terminal (editor: Positron)
- OS: macOS 26.5.1

### Quarto check output

```
Quarto 1.10.12
[✓] Checking environment information...
      Quarto cache location: /Users/charlottewickham/Library/Caches/quarto
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.8.3: OK
      Dart Sass version 1.87.0: OK
      Deno version 2.7.14: OK
      Typst version 0.14.2: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.10.12
      Path: /Applications/quarto/bin

[...]

Issue #14684, opened 2026-07-13
No skip-to-content link: keyboard users must tab through all navigation on every page (WCAG 2.4.1, Level A)


cwickham:
## I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

## Bug description

Quarto's HTML outputs render navigation — navbar, search, sidebar, page tabs, TOC — before the main content, with no "skip to main content" link. Keyboard and switch users must therefore tab through every navigation stop on **every page** before reaching the content, which fails [WCAG 2.2 SC 2.4.1 "Bypass Blocks"](https://www.w3.org/WAI/WCAG22/Understanding/bypass-blocks.html) (Level A) — a criterion Section 508, EN 301 549, and the US ADA Title II rule all incorporate.

Authors have no remedy short of hand-injecting HTML (the `announcement` workaround from [#11602](https://github.com/quarto-dev/quarto-cli/issues/11602) consumes that feature and leaves a visible bar), so I'm filing this as a bug rather than a feature request; the standing request is discussion [#10096](https://github.com/orgs/quarto-dev/discussions/10096).

Most of the machinery already exists:

- The **target** has shipped since 2022: [59616f22e](https://github.com/quarto-dev/quarto-cli/commit/59616f22ea487bbcc4839de75ca842e71f5aadc5) added `<main id="quarto-document-content">` explicitly "for things like skip links" ([before-body-article.ejs#L20](https://github.com/quarto-dev/quarto-cli/blob/8577ed1c69b3b9a551f1669c8256abc4a1731266/src/resources/formats/html/templates/before-body-article.ejs#L20), [nav-before-body.ejs#L128-L130](https://github.com/quarto-dev/quarto-cli/blob/8577ed1c69b3b9a551f1669c8256abc4a1731266/src/resources/projects/website/templates/nav-before-body.ejs#L128-L130)); the link never followed.
- Bootstrap's `.visually-hidden-focusable` helper (what Bootstrap's own docs use for their skip link) is already compiled into every Quarto theme bundle.
- Exception: `format: dashboard` output has no `<main>` and no id on its content container, so there is currently nothing to target — itself a missing-landmark issue.

Related:

[...]

PR #14816, opened 2026-08-27
Make scrollable code blocks and cell output keyboard-focusable in HTML output

### LINKED ISSUE #14378: a11y: scrollable code blocks and output pre lack tabindex=0 (keyboard inaccessible)
Reported in #14373.

Repro: https://github.com/alexpghayes/quarto-accessibility-reprex (post/post-1)

axe rule: `scrollable-region-focusable` · impact: serious · tag: wcag2a (2.1.1, 2.1.3)
Targets: `<div class="sourceCode cell-code" id="cb...">`, `<pre>` inside `.cell-output`

When code blocks or cell output overflow the viewport width they become scrollable, but Quarto doesn't add `tabindex="0"` — so keyboard-only users cannot scroll to see the hidden content. The fix should conditionally add `tabindex="0"` when the element's scrollWidth exceeds its clientWidth, or unconditionally for `.sourceCode` / `.cell-output pre`.

Part of #8706.

cwickham:
Closes #14378. Part of #8706.

## What this does

A code block or cell output that overflows is scrollable, but keyboard users cannot reach or scroll it in Safari. Chrome and Firefox already make such regions focusable natively, so this is a Safari gap for users and a missing accessible name everywhere. axe reports `scrollable-region-focusable` (serious, WCAG 2.1.1/2.1.3) in all three, because it reads markup rather than browser behavior. Compatibility under Design notes has the detail.

This adds a runtime module to `quarto.js`, for bootstrap HTML formats only. While a region overflows it gets `tabindex="0"`, `role="group"`, a localized `aria-label`, and a `data-quarto-scrollable` marker so removal only touches attributes we added. The attributes come off when the region fits again. The sync runs on `DOMContentLoaded` and from the existing throttled body `ResizeObserver`, so it adds no new observers.

The candidates are the three scroll containers in rendered HTML:

| Candidate | Overflow source |
|---|---|
| `div.sourceCode` | Pandoc's `div.sourceCode { overflow: auto }` for highlighted code |
| `pre` | Bootstrap reboot's `pre { overflow: auto }` |
| `.cell-output-display:not(.no-overflow-x)` | [`_quarto-rules.scss`](https://github.com/quarto-dev/quarto-cli/blob/1f4ff1871010c559d33719d4a9251c6afb6e131b/src/resources/formats/html/_quarto-rules.scss#L384-L387) |

## Which regions get marked

A region qualifies when its computed overflow is `auto` or `scroll`, which is axe's own condition, and it overflows by more than 1px. Three kinds of region are skipped:

- **Author markup**, meaning an existing `tabindex` or `aria-label`.
- **No usable size**: a closed `<details>`, an inactive tab pane, or a `visually-hidden` code alternative. The last is clipped to 1px, which makes its content height read as overflow. Marking it would add an invisible tab stop, and quarto-web's get-started pages carry about nine each.
- **Real focusable content**, such as a link in a wide table, which the browser already scrolls into view.

Pandoc's line anchors are the exception to that third rule. Without line numbers they carry `tabindex="-1"`, so they are out of the tab order anyway. With line numbers they are real focusable links, kept that way on purpose by #14655. But each one sits at the start of its line, so focusing it leaves `scrollLeft` at 0 and the clipped content unreachable. A numbered block therefore satisfies axe while staying unscrollable, so it gets its own region tab stop alongside its per-line links.

## When it re-syncs

One rule, not one per widget: the module re-syncs when the page's layout changes size. Collapsed callouts, folded `<details>`, and tab panes of differing heights all make the page taller, so a single `ResizeObserver` covers them with no mechanism-specific code.

The boundary is a reveal that changes no size, and two tab panes of equal height are the clean example. That case keeps today's behavior rather than regressing anything. A `shown.bs.tab` hook would fix tabsets while still missing equal-height accordions and overlay panels, trading one documented boundary for a special case plus a boundary.

## Design notes

**Runtime JS, not a static `tabindex`.** A Lua filter would put a tab stop on every code block, whether or not it scrolls. It would also change every rendered HTML file, breaking downstream snapshot tests and scrapers. Rendered HTML stays byte-identical here.

**Compatibility.** Chrome (since 132, January 2025) and Firefox already make these regions focusable natively ([Chrome blog](https://developer.chrome.com/blog/keyboard-focusable-scrollers)). These tab stops have therefore been live on every Quarto site for over a year, with no reports. That bounds the risk: this polyfills shipped browser behavior rather than inventing a new interaction. What it adds on top is Safari coverage, an accessible name, and axe conformance. Every failure mode leaves an element exactly as it is today.

**Labels.** Two new language keys, `scrollable-code-label` and `scrollable-output-label`. `quarto.js` is static, so the after-body script passes the resolved strings through a window global. The module falls back to the English defaults.

**`role="group"`, not `region`.** Pages carry many code blocks, and `region` landmarks would flood the screen-reader rotor.

## Tests

- 15 unit tests (`tests/unit/scrollable-regions.test.ts`): overflow geometry, usable size, the focusable-content rule, label merging.
- 30 Playwright tests (`tests/integration/playwright/tests/html-scrollable-regions.spec.ts`) on chromium, firefox, and webkit: marking and removal across viewports, the three skips, one tab stop per block, real-Tab reach with a visible focus ring, and arrow-key scrolling. Arrow keys are asserted on chromium and firefox only, because Playwright's WebKit routes them to the page even when a scroller has focus.
- `axe-accessibility` still passes (116 tests), with its code-line-number fixtures re-rendered.
- On quarto-web pages rendered with this build, the violations on `about.html` (`#cb1`) and `docs/troubleshooting/index.html` (bare `pre`) are gone, and axe reports none for the rule at 390x844.
- A full sync costs 0.6ms median on a 200-block page (400 candidates), against the existing 50ms throttle.

## Checklist

I have (if applicable):

- [x] ensured the present test suite passes (the suites listed above; not the full matrix)
- [x] added new tests

<details>

</details>


cwickham:
Filed #14817 for the revealjs side, which this PR does not cover. Same axe rule, different target: there the scroll container is `.reveal pre code` rather than `div.sourceCode`, and axe stays silent because revealjs enables code line numbers by default.

src/resources/formats/html/scrollable-regions/scrollable-regions.js
```diff
@@ -0,0 +1,133 @@
...
+  "div.sourceCode, pre, .cell-output-display:not(.no-overflow-x)";
+
+// Mirrors axe's pass condition (and Chrome's native heuristic): a region with
+// keyboard-focusable content is already reachable, and adding tabindex would
+// create a double tab stop. tabindex="-1" removes an element from the tab
+// order, so Pandoc's per-line anchors (`a[href][tabindex="-1"]`) don't count.
+const kFocusableSelector =
+  'a[href]:not([tabindex="-1"]), button:not([tabindex="-1"]), ' +
+  'input:not([tabindex="-1"]), select:not([tabindex="-1"]), ' +
+  'textarea:not([tabindex="-1"]), [tabindex]:not([tabindex="-1"])';
```
cderv, line 31: `kFocusableSelector` doesn't exclude disabled controls. `button:disabled`, `input:disabled`, etc. aren't in the tab order, so `hasFocusableContent` returns `true` for a region whose only "focusable" descendant is a disabled control, and `syncScrollableRegions` skips adding a tab stop there. The region ends up scrollable but unreachable by keyboard. Add `:not(:disabled)` to the button/input/select/textarea clauses.

src/resources/formats/html/scrollable-regions/scrollable-regions.js
```diff
@@ -0,0 +1,133 @@
...
+      !isScrollable(el, getComputedStyle(el)) &&
+      el !== document.activeElement
+    ) {
+      el.removeAttribute("tabindex");
+      el.removeAttribute("role");
+      el.removeAttribute("aria-label");
+      el.removeAttribute(kMarker);
+    }
+  }
+}
```
cderv, line 133: The skip guard (lines 110–116) checks `tabindex` and `aria-label` but not `role`. An element with an author-set `role` (e.g. `<pre role="log">`) isn't skipped here: it gets `role="group"` overwritten at line 119, then `role` is stripped entirely at line 128 once the region stops overflowing. That contradicts the comment above `kMarker` at the top of the file ("removal never touches author markup") for this one attribute. Add `role` to the skip condition, or save/restore the original value instead of unconditionally removing it.

src/resources/formats/html/scrollable-regions/scrollable-regions.js
```diff
@@ -0,0 +1,133 @@
...
+      }
+    } else if (
+      !isScrollable(el, getComputedStyle(el)) &&
+      el !== document.activeElement
+    ) {
+      el.removeAttribute("tabindex");
+      el.removeAttribute("role");
+      el.removeAttribute("aria-label");
+      el.removeAttribute(kMarker);
+    }
```
cderv, line 131: Cleanup is skipped while `el === document.activeElement` (line 125), which avoids pulling attributes out from under focused content mid-interaction — reasonable. But nothing reruns the check once focus leaves that element. Sequence: Tab into a scrollable region, resize so it no longer overflows, Tab away. `tabindex`, `role`, `aria-label`, and the marker all stay stale until the next unrelated resize or reload. Worth a `focusout` listener on marked regions that reruns `syncScrollableRegions` (or just the check for that one element) once focus leaves.

src/resources/language/_language.yml
```diff
@@ -82,6 +82,10 @@ toggle-dark-mode: "Toggle dark mode"
 toggle-reader-mode: "Toggle reader mode"
 toggle-navigation: "Toggle navigation"
 
+# Accessible names for scrollable regions (never shown as visible text)
+scrollable-code-label: "Scrollable code"
+scrollable-output-label: "Scrollable output"
```
cderv, line 87: These two keys only exist in the English `_language.yml` — none of the 33 locale files got them. That's ok (precedent: #14376 added nav-landmark labels English-only) but it's the exact drift #14822 was filed to track, and the more recent #14684 (skip-to-content link) already raised the bar by shipping translations for the major locales (de/es/fr/it/ja/nl/pt/pt-BR/zh/zh-TW) in the same PR. Worth adding at least those before merge rather than leaning on #14822 to backfill later — it's aria-label-only text, so translation cost is two short strings per locale, not a big lift.

cderv (changes requested):
Nice piece of work, thorough tests and design write-up. Left a few inline comments on things worth fixing before merge.

Two are real correctness issues in `scrollable-regions.js`: disabled controls are currently counted as focusable content, so a region whose only focusable descendant is a disabled button/input skips its tab stop entirely. And the guard meant to respect author markup checks `tabindex`/`aria-label` but not `role`, so an author-set role gets overwritten then stripped once the region stops overflowing.

There's also a stale-state case: cleanup correctly skips a focused region, but nothing reruns it once focus leaves, so `tabindex`/`role`/`aria-label` can stick around until the next unrelated resize.

Last one is smaller: the two new `_language.yml` keys are English-only. That's legal, we did the same for #14376, but #14684 already raised the bar by shipping major-locale translations in the same PR, and #14822 exists because we keep letting this drift. Could we add at least the major locales here rather than leaning on #14822 to backfill?

cwickham:
@cderv Thanks — all four are in.

**Disabled controls.** `:not(:disabled)` now sits on the button, input, select and textarea clauses, and on the generic `[tabindex]` one, which also catches `<button disabled tabindex="0">`. It matches only form controls, so it is inert elsewhere.

**Author `role`.** I added `role` to the skip guard rather than save and restore it. That makes the guard uniform: any of `tabindex`, `role` or `aria-label` means the author has taken control, so the runtime leaves the element alone. A `<pre role="log">` that overflows keeps its role and gets no tab stop, and Chrome and Firefox still focus it natively.

**Stale state.** Cleanup now attaches a one-shot `focusout` listener to a region it had to skip, and re-runs the sync when focus leaves. A `WeakSet` stops repeat syncs stacking listeners on one element.

Three new Playwright cases cover these, green on all three engines.

**Language keys.** All 34 locale files now carry both keys (`fr-CA` is a three-key overlay on `fr`, so it is skipped — the set #14376 covered). "Scrollable" comes from Firefox DevTools' tooltip for an element with scrollable overflow, `code` from each file's own `code-summary`, and `output` from LibreOffice Base's Direct SQL dialog, where it labels the results of the command you ran.

This is weaker than #14376, where 134 of 165 cells were attested strings used verbatim. Every value here is a phrase assembled from two attested parts, so all 68 name their sources and end `needs review`. Six locales have no attested word for "scrollable" at all — `az`, `fi`, `is`, `nb`, `nn`, `sr-Latn` — and each says so on its own line.

```
grep -rn 'needs review' src/resources/language/
```

Three extra commits you did not ask for.

The labels reach the page through `JSON.stringify` in the after-body template, which escapes quotes but not `</script>`, so a label holding one closed the script element early. `<` is now escaped to `\u003c`. A new `lang: fr` fixture has a `language:` override that holds `</script>`, and it checks that both labels reach `aria-label` intact.

The body `ResizeObserver`'s throttle only ran on the leading edge, so a resize that ended inside the 50ms wait was never synced. It now also runs once at the end of a burst, for that observer only. `layoutMarginEls()` and the reader-mode check share the callback, so they get the fix too. The TOC and sidebar scroll handlers are unchanged.

I also merged `main`.

PR #14685, opened 2026-07-14
Add skip-to-content link to Bootstrap-themed HTML output

cscheid:
Closes #14684.

> **Stacked on #14820.** This PR targets `fix/14818-dashboard-non-page-hash`, not `main`. The diff shown here is only the skip-link work. #14820 fixes two dashboard hash bugs that predate this PR and that this PR would otherwise expose on every dashboard. Merge #14820 first.

## What this does

Adds a "Skip to main content" link as the first element of `<body>` for all Bootstrap-themed HTML output — standalone documents, websites, books, manuscripts, and dashboards — so keyboard and switch users can bypass navigation blocks (WCAG 2.2 SC 2.4.1 "Bypass Blocks", Level A).

- The link is injected by the bootstrap HTML postprocessor, so a single implementation covers every Bootstrap-based output shape (dashboards call the base HTML format extras, so they get it too). `theme: none` / `minimal` output is untouched.
- Target resolution falls back gracefully: `#quarto-document-content` → first `<main>` → `div.page-layout-custom` (dashboards, standalone custom layouts) → `#quarto-content` (website custom layouts). If the target has no id it is assigned `quarto-document-content`, and it gets `tabindex="-1"` so activating the link moves focus, not just scroll position.
- Visually hidden via Bootstrap's `.visually-hidden-focusable` helper (already compiled into every theme); when focused it appears as a fixed-position pill above the fixed/headroom navbar (`$zindex-skip-link: 1070`), styled from theme variables so it works in light and dark themes.
- Link text is localizable via a new `language` key `skip-to-content`, with seeded translations for fr, de, es, pt, pt-BR, it, ja, zh, zh-TW, nl (other languages fall back to English; regional variants inherit their base language).
- Dashboards additionally mark their content container with `role="main"`, addressing the missing main-landmark note in the issue.

## Dashboards need JavaScript for the link

On a dashboard the link cannot be a plain anchor jump. A dashboard uses the URL hash to select a page, so jumping to `#quarto-document-content` pushes a history entry that names no page. Back then landed on a dead step: the URL changed but the visible page did not, and the reader had to press Back twice.

The dashboard now handles the click, moves focus to the target, and leaves the URL alone. Back and Forward behave as they do without the link, and a page hash such as `#sales` survives using the link, so the shareable URL still works. Without JavaScript the link is still an ordinary in-page anchor.

Everything else is unchanged: no JavaScript is involved for documents, websites, books, or manuscripts.

## Behavior

[...]

PR #14820, opened 2026-08-27
Fix dashboard URL hash handling

### LINKED ISSUE #14818: A URL hash that is not a page name blanks a multi-page dashboard
## I have:

## Bug description

In a dashboard with more than one page, a URL hash that does not name a page hides every page. The content area becomes empty. The navbar stays visible.

Hashes that do not name a page are common. Examples are footnote links, cross-references, and hand-written in-page anchors.

`showPage()` removes the `active` class from every nav tab and every `.tab-pane`. It then adds `active` back only to the pane whose id matches the hash ([quarto-dashboard.js#L242-L266](https://github.com/quarto-dev/quarto-cli/blob/b69ed2dde16d1b502dcdddace4eebae5038550e3/src/resources/formats/dashboard/quarto-dashboard.js#L242-L266)). When no pane matches, no pane is active, so the dashboard shows nothing.

Two places pass the hash to `showPage()` without a test:

- the hash processing on load ([#L109-L117](https://github.com/quarto-dev/quarto-cli/blob/b69ed2dde16d1b502dcdddace4eebae5038550e3/src/resources/formats/dashboard/quarto-dashboard.js#L109-L117))
- the `popstate` handler, which runs on every hash change ([#L120-L123](https://github.com/quarto-dev/quarto-cli/blob/b69ed2dde16d1b502dcdddace4eebae5038550e3/src/resources/formats/dashboard/quarto-dashboard.js#L120-L123))

`QuartoDashboardUtils.isPage()` already does this test ([#L229-L232](https://github.com/quarto-dev/quarto-cli/blob/b69ed2dde16d1b502dcdddace4eebae5038550e3/src/resources/formats/dashboard/quarto-dashboard.js#L229-L232)), but neither place calls it.

This is the general form of #9411, which #11264 closed. That fix stopped external links from reaching `showPage()`. It did not change the load path or the `popstate` path, so in-page anchors still blank the dashboard.

One related problem affects any fix: `isPage()` builds a CSS selector by string concatenation. It throws a `SyntaxError` for a hash that is not a valid selector, such as `#`, `#1foo`, or `#a:b`. On load, the code that removes the `hidden` class runs after this point. A guard that calls `isPage()` must therefore not throw, or the whole dashboard stays invisible.

## Steps to reproduce

````qmd
---
title: "Two page dashboard"
format: dashboard
---

# Heat

## Row

Content A

# Cool

## Row

Content B
````

1. Render the document.
2. Open the HTML output in a browser.
3. Add `#no-such-thing` to the end of the URL, then load the page.

## Actual behavior

The content area is empty. Both panes lose the `active` class, and no navbar tab is active. The navbar still shows both page names.

A reload of the same URL shows the same empty state.

## Expected behavior

The dashboard keeps the current page. If the anchor exists in the page, the browser scrolls to it.

## Your environment

- Positron 1.124.0
- macOS 26.5.2 (build 25F84)

## Quarto check output

```
Quarto 1.10.18
      Pandoc version 3.10.0: OK
      Dart Sass version 1.101.0: OK
      Deno version 2.7.14: OK
      Typst version 0.15.1: OK
      Version: 1.10.18
      TinyTeX: v2026.04
      VeraPDF: 1.28.2
      Using: TinyTex
      Version: 2026
      Version: 150.0.7871.115
      Version: 4.5.2
      Path: /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources
      LibPaths:
        - /Users/charlottewickham/Library/R/arm64/4.5/library
      knitr: 1.51
      rmarkdown: 2.30
      Version: 3.12.2
      Jupyter: 5.9.1
      Kernels: python3
```

---

*Investigation was AI-assisted, grounded in a local clone (per CONTRIBUTING.md).*

### LINKED ISSUE #14819: a11y: a dashboard page hash leaves the navbar unreachable by forward Tab
Found while reviewing #14685.

A dashboard navbar links to its own pages with a URL hash, such as `dashboard.html#cool`. These are the URLs that a dashboard author shares.

When a page loads with that hash, the browser sets the sequential focus navigation starting point to the target element. The target is the whole `.tab-pane` for that page. The first Tab press therefore moves focus into the page content.

The navbar comes before that starting point in DOM order, so forward Tab never reaches it. A keyboard user must press Shift+Tab, or tab through the rest of the document to wrap around.

**Repro:** render a dashboard with two pages, open `…#cool`, then press Tab. Focus lands on the first control inside the Cool page. The navbar tabs and the page title are not reachable by forward Tab.

Verified in released Quarto 1.10.18 and on `main` (b69ed2dd), in Chromium.

A page hash selects a page. It is not a position within one. On load, when the hash names a page ([`isPage()`](https://github.com/quarto-dev/quarto-cli/blob/b69ed2dde16d1b502dcdddace4eebae5038550e3/src/resources/formats/dashboard/quarto-dashboard.js#L229-L232)), the dashboard should return the focus starting point to the top of the document.

This also makes the skip-to-content link of #14684 unreachable, because that link is the first element of `<body>`.

Part of #8706.

---

*Investigation was AI-assisted, grounded in a local clone (per CONTRIBUTING.md).*

cwickham:
## Description

Two dashboard bugs in URL hash handling. Both predate #14685; both surfaced while reviewing it.

Closes #14818. Closes #14819.

### A hash that names no page blanked the dashboard (#14818)

`showPage()` removes `active` from every nav tab and every `.tab-pane`, then restores it only on the pane that matches the hash. When no pane matches, nothing is restored and the content area is empty.

Two callers passed the hash through without a test: the processing on load, and the `popstate` handler, which runs on every hash change. Any in-page anchor therefore blanked a dashboard with more than one page. A footnote link, a cross-reference, or a hand-written anchor all did it, and a reload of that URL showed the same empty state.

Both callers now test with `isPage()` first.

`isPage()` needed work before it could be used as a guard. It built a CSS selector by concatenating the hash, which throws a `SyntaxError` for a hash that is not a valid selector, such as `#`, `#1foo`, or `#a:b`. On load, that throw would happen before the `hidden` class is removed, and the whole dashboard would stay invisible. It now uses `getElementById`.

#9411 was a narrower instance of the same root cause. #11264 closed it by keeping external links away from `showPage()`, and left the load and `popstate` paths unchanged.

### Landing on a page hash put the navbar out of reach (#14819)

A dashboard navbar links to its pages with a URL hash, so `dashboard.html#sales` is the shareable URL for a page. Loading such a URL makes the browser set the sequential focus navigation starting point to the target element, which for a dashboard is the whole `.tab-pane`. The first Tab press landed inside the page content, and the navbar comes earlier in DOM order, so tabbing forward never reached it.

A page hash selects a page. It is not a position within one. When the hash names a page, the focus starting point now returns to the top of the document. This runs after `load`, because the browser applies its own fragment behavior later than `DOMContentLoaded`.

## Tests

New Playwright spec at `tests/integration/playwright/tests/dashboard-hash-navigation.spec.ts`, with a fixture at `tests/docs/playwright/dashboard/hash-navigation.qmd`. This is the first dashboard spec in the suite.

Five tests, passing on Chromium, Firefox, and WebKit:

- a hash that names no page keeps the current page, on load and on a hash change
- a hash that is not a valid CSS selector does not leave the dashboard hidden
- page navigation, Back, and Forward still work
- landing on a page hash starts tabbing at the top, and the navbar tab is reachable

Verified against a build without the fix: four of the five fail. The fifth is the navigation regression guard, which passes either way.

WebKit only tabs to links when Alt is held, which matches Safari's default setting, so the focus test picks its key by browser.

## Checklist

I have (if applicable):

- [x] added new tests

<details>

</details>


cwickham:
Stacked below #14685, which adds the skip-to-content link. That PR targets this branch and exposed both of these bugs on every dashboard, but neither is caused by it — both reproduce on `main` and in released Quarto 1.10.18. Merge this one first.

cderv (approved):
LGTM. I'll push a small tweak before merging: decodeURIComponent(window.location.hash) can throw on malformed percent-encoding (e.g. #%), which would leave the dashboard hidden the same way the cases this PR already fixes did. Adding a try/catch fallback and a regression test, then merging.

PR #14853, opened 2026-09-04
Suffix the ids in the cloned body TOC

### LINKED ISSUE #14844: `toc-location: left-body` clones the TOC without renaming its ids
## I have:

## Bug description

With `toc-location: left-body` or `toc-location: right-body`, Quarto renders the
table of contents twice. The `useDoubleToc` branch clones the whole `nav` and
renames only the `nav` itself, from `TOC` to `TOC-body`:

https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/format/html/format-html-bootstrap.ts#L352-L363

Every other id inside the clone is copied as is. The page then holds two
elements for each id in the TOC: the `h2#toc-title` heading, and one
`a#toc-<section>` link per entry. The `id` attribute must be unique. From the
HTML Standard, [The `id` attribute](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute):

> When specified on HTML elements, the `id` attribute value must be unique
> amongst all the IDs in the element's tree and must contain at least one
> character.

The duplicate ids are old. #14813 makes one of them load-bearing. That PR labels
the TOC landmark with `aria-labelledby="toc-title"`:

https://github.com/quarto-dev/quarto-cli/blob/471531762d94e7badf0bd2e2f62461be07afc624/src/resources/formats/html/pandoc/toc.html#L1-L6

Both `nav` elements now point at an id that resolves to one of two elements.
This is not a name bug, because both headings carry the same text. It does turn
an inert duplicate id into an ARIA reference. In axe-core 4.10.3, the version
Quarto vendors, `duplicate-id-aria` moves from *inapplicable* to *needs review*
(impact `critical`) on the same page.

This came out of review on #14813 (thanks @cderv). The duplicate id predates
that PR — measured below.

## Steps to reproduce

````qmd
---
title: "Left-body TOC"
format:
  html:
    toc: true
    toc-location: left-body
    toc-title: "On this page"
---

## First section

Some text.

## Second section

More text.
````

Then count the ids in the output:

```sh
quarto render index.qmd
grep -c 'id="toc-title"' index.html
```

## Actual behavior

Two elements carry `id="toc-title"`, and each TOC link id appears twice:

```html
<nav id="TOC" role="doc-toc" aria-labelledby="toc-title" class="toc-active">
  <h2 id="toc-title">On this page</h2>
  <ul>
    <li><a href="#first-section" id="toc-first-section" ...>First section</a></li>
...
<nav id="TOC-body" role="doc-toc" aria-labelledby="toc-title">
  <h2 id="toc-title">On this page</h2>
  <ul>
    <li><a href="#first-section" id="toc-first-section">First section</a></li>
```

A DOM count of duplicated ids on that page gives
`toc-title`, `toc-first-section`, `toc-second-section` — one entry per TOC line,
plus the heading. The count grows with the size of the TOC.

axe-core 4.10.3 over the rendered page, in headless Chromium 151, Firefox 153
and WebKit 26.5:

| rendered with | `duplicate-id-aria` | `landmark-unique` |
|---|---|---|
| current `main` | inapplicable | violation (moderate) |
| #14813 branch | **incomplete, critical** | violation (moderate) |

`landmark-unique` fires either way, so it is not a regression from #14813.
`duplicate-id-aria` is new, because `toc-title` only becomes an accessible
reference once `aria-labelledby` points at it. The rule reports as *incomplete*
rather than a violation because it carries `reviewOnFail: true`.

All three engines resolve the reference the same way, to the first element in
tree order — the heading inside `nav#TOC`. Both landmarks compute the same
accessible name, "On this page". Nothing user-facing breaks today. The only
id-based lookup in Quarto's own JS is
[`titleSelector: "#toc-title"`](https://github.com/quarto-dev/quarto-cli/blob/abc6a78ed68f9e8bc9d54e27851093bd687a1cb7/src/resources/formats/html/quarto.js#L556-L572),
which reads `document.body.querySelector("#toc-title")` and so takes the first
match. Both headings hold identical text, so the sidebar toggle reads correctly
by accident.

## Expected behavior

Each id appears once. The clone gets its own heading id, and its
`aria-labelledby` points at that id:

```html
<nav id="TOC-body" role="doc-toc" aria-labelledby="toc-title-body">
  <h2 id="toc-title-body">On this page</h2>
```

Both landmarks keep the accessible name "On this page". That shared name is
correct here, and it is what the ARIA Authoring Practices Guide asks for. From
[Landmark Regions, Step 3: Label areas](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/):

> There is one rare circumstance where providing the same label to multiple
> instances of a landmark can be beneficial: the content and purpose of each
> instance is identical.

and, in the same document, on navigation landmarks:

> If a `navigation` landmark has an identical set of links as another
> `navigation` landmark on the page, use the same label for each `navigation`
> landmark.

So the fix is the id, not the label. The two TOCs must keep the same name, which
means axe's `landmark-unique` result stays. That rule is `best-practice`, not
WCAG, and this is the case the APG carves out.

The duplicated `a#toc-<section>` link ids are the same fault from the same clone.
They are not ARIA references, and axe's `duplicate-id` rule is deprecated and
off by default, so they produce no finding. They are still invalid HTML. Worth
handling in the same place, or in a follow-up.

## Your environment

- Positron 1.124.0
- macOS 26.5.2 (build 25F84)

Reproduced with released Quarto 1.10.18 and with a dev build of 1.11.

An AI assistant helped investigate, grounded in a local clone
(CONTRIBUTING.md, "Using AI tools to investigate").

## Quarto check output

```
Quarto 1.10.18
      Pandoc version 3.10.0: OK
      Dart Sass version 1.101.0: OK
      Deno version 2.7.14: OK
      Typst version 0.15.1: OK
      Version: 1.10.18
      TinyTeX: v2026.04
      VeraPDF: 1.28.2
      Using: TinyTex
      Version: 2026
      Version: 150.0.7871.115
      Version: 4.5.2
      Path: /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources
      LibPaths:
        - /Users/charlottewickham/Library/R/arm64/4.5/library
      knitr: 1.51
      rmarkdown: 2.30
      Version: 3.12.2
      Jupyter: 5.9.1
      Kernels: python3
```

cwickham:
Closes #14844.

## Description

With `toc-location: left-body` or `right-body`, the `useDoubleToc` branch clones
the whole table-of-contents `nav` and renames only the `nav` itself, from `TOC`
to `TOC-body`. Every id inside the clone was copied as is. A page therefore
carried two elements for the `toc-title` heading, and two for each
`toc-<section>` link. The number of duplicates grows with the size of the TOC.

Ids must be unique in the tree ([HTML Standard, the `id` attribute](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute)).
The duplicates were inert until #14376 labelled the TOC landmark with
`aria-labelledby="toc-title"`. That reference makes one of them load-bearing.

This PR suffixes every id the clone carries with `-body`, and repoints the
clone's `aria-labelledby` at the renamed heading:

```html
<nav id="TOC" role="doc-toc" aria-labelledby="toc-title" class="toc-active">
  <h2 id="toc-title">Contents</h2>
  ...<a href="#section-1" id="toc-section-1">Section 1</a>

<nav id="TOC-body" role="doc-toc" aria-labelledby="toc-title-body">
  <h2 id="toc-title-body">Contents</h2>
  ...<a href="#section-1" id="toc-section-1-body">Section 1</a>
```

### Why every id, and not only the heading

The heading is the one id that ARIA references, so a minimal fix could rename
only that. I widened it because nothing outside the `nav` reads the link ids.
They come from Pandoc's TOC writer, and a search across `src/` for `"toc-"`,
` #toc-$ `, and `getElementById("toc` finds no consumer in Quarto's JS, SCSS, or
Lua filters.

The originals stay on the sidebar copy, which is first in tree order. So
`document.getElementById("toc-section-1")` returned the sidebar element before
this change and returns the same element after it. Third-party CSS or JS that
targets these ids sees no difference.

### The accessible name does not change

Both landmarks keep the name from the TOC title. That shared name is the correct
outcome here, and it is what the ARIA Authoring Practices Guide asks for
([Landmark Regions, Step 3](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/)):

> If a `navigation` landmark has an identical set of links as another
> `navigation` landmark on the page, use the same label for each `navigation`
> landmark.

axe-core's `landmark-unique` still reports these two `nav` elements. That is a
`best-practice` rule, not WCAG, and this is the case the APG carves out. Do not
"fix" it by giving the two TOCs different labels — that would make the page
worse for screen reader users. It belongs in an axe baseline.

## Measured effect

axe-core 4.10.3, the version Quarto vendors, over a rendered
`toc-location: left-body` page:

| rendered with | duplicate ids | `duplicate-id-aria` | `landmark-unique` |
|---|---|---|---|
| `main` | 4 | inapplicable | violation (moderate) |
| #14813 | 4 | **incomplete, critical** | violation (moderate) |
| this PR | **0** | **passes** | violation (moderate) |

`duplicate-id-aria` now passes rather than being inapplicable, because the
reference exists and resolves to one element.

## Testing

Extended the two existing fixtures for this layout,
`tests/docs/smoke-all/issues/3473-toc-side-body/{left-body,right-body}.qmd`,
with positive assertions for the suffixed ids and negative assertions that the
clone no longer carries `toc-title` or `toc-section-1`.

Both pass. With the change to `format-html-bootstrap.ts` reverted,
`left-body.qmd` fails, so the assertions catch the fault they describe.

Also ran locally, all passing:

- `3473-toc-side-body/{body,left,right}.qmd` — the layouts that render one TOC
  are unaffected.
- `accessibility/toc-nav-label.qmd` — the fixture #14813 adds.

Checked by hand:

- Both landmarks still report the name "Contents" in Chromium's accessibility
  tree, for `nav#TOC` and `nav#TOC-body`.
- `right-body` behaves the same as `left-body`.
- `toc-title: ""` produces no heading and no `aria-labelledby`. The guard holds,
  the link ids are still suffixed, and no id repeats.
- `toc-title: false` fails YAML validation on the base branch too, so no TOC
  renders without a heading through metadata.

I did not run the full suite locally. CI covers it.

## Checklist

I have (if applicable):

- [x] added new tests

The ids are internal, and no documented option changes, so there is nothing to
document on quarto.org.

<details>

</details>


src/format/html/format-html-bootstrap.ts
```diff
@@ -355,6 +355,21 @@ function bootstrapHtmlPostprocessor(
...
+        // originals stay on `toc`, so existing lookups resolve as before.
+        const clonedLabelId = clonedToc.getAttribute("aria-labelledby");
+        const clonedIdEls = clonedToc.querySelectorAll("[id]");
+        for (let i = 0; i < clonedIdEls.length; i++) {
+          const clonedIdEl = clonedIdEls[i] as Element;
+          clonedIdEl.id = `${clonedIdEl.id}-body`;
+        }
+        if (clonedLabelId) {
+          clonedToc.setAttribute("aria-labelledby", `${clonedLabelId}-body`);
+        }
```
cderv, line 371: The suffix isn't guaranteed unique against the rest of the document. Pandoc dedupes identical heading slugs, but it doesn't check whether a slug already matches another heading's slug plus `-body`.

`# Setup` produces `toc-setup`. `# Setup Body` produces `toc-setup-body`, already sitting in the sidebar TOC. Suffixing the clone's `toc-setup` gives `toc-setup-body` too, colliding with the second heading's own entry, and we're back to the `duplicate-id-aria` failure this PR closes.

I don't know if this case can happen a lot or if this is really edge case ?  What do you think ? 

If this is possibly, I think this needs a real uniqueness check against the existing ids rather than a fixed suffix, plus a fixture with a heading pair that triggers this (something like `Setup` / `Setup Body`) so it doesn't come back silently.

src/format/html/format-html-bootstrap.ts
cderv, line 376: Code above runs the id-suffix loop over the whole cloned nav, including the `.toc-actions` subtree, before it's removed here. This is not wrong, but just wasted work if that subtree carries ids — we could just remove `.toc-actions` first and so do the toc id processing after. What do you think ?

cderv (commented):
The id-suffix approach doesn't check for collisions against the rest of the document. Left a note on the specific line. Tell me if you think this is too much of a edge case. 

I also fixed conflict by merging main, and fixing the artifacts

PR #14864, opened 2026-09-08
chore(julia-engine): Update subtree to 0.2.2

jkrumbiegel:
Pulls in julia-engine [v0.2.2](https://github.com/PumasAI/quarto-julia-engine/releases/tag/v0.2.2), which closes the notebook's worker process when a non-daemon render fails. Before, the worker was only closed on success, so a failing render left it running until the server exited.


cderv (approved):
Thank you !

PR #14866, opened 2026-09-08
[Tests] update R, Python and Julia

cderv:
This update 

- R 4.6.1 and latest packages
- Python 3.14.7, latest uv and latest packages
- Julia 3.12.7 and update of packages


cderv:
CI was failing on `docs/smoke-all/jupyter/image-dimensions/matplotlib-quarto-retina.qmd` on both `ubuntu-latest` and `windows-latest`. Root cause: this PR's matplotlib bump (3.10.8 -> 3.11.1 in `tests/uv.lock`) shifted the tight-bbox crop width of that test's figure by one pixel (537 -> 538), so the hardcoded `img[width="537"]` assertion no longer matched. Reproduced locally with the updated env and confirmed the new correct value is 538. Fixed in 969bfe92f by updating the expected width.

Issue #10696, opened 2024-09-03
Add some in-browser testing for revealjs using puppeteer 


cderv:
Some of revealjs feature, and especially our quarto-support plugin can't be tested unless we do automated headless testing. 

This issue is to store the potential test to implement. 

- https://github.com/quarto-dev/quarto-cli/pull/10695
- https://github.com/quarto-dev/quarto-cli/pull/10688

cderv:
@mcanouil I believe puppeteer tag is about the rendering using chrome print we have internal to quarto, for example to have mermaid graph be image in PNG. it is different that the playwright stuff we have for tests. I mistitled indeed, but correctly put the testing tag, which seems enough.

mcanouil:
Oh, thanks it add some sense to another comment.

PR #14869, opened 2026-09-08
Make Quarto Posit Workbench aware

### LINKED ISSUE #14865: Quarto doesn't recognize same-origin links on Posit Workbench
Clicking a link inside a Quarto preview in Positron on Workbench opens the page in a new browser tab instead of navigating inside the pane.

## Reproducing

Render this dashboard and preview it in Positron's Viewer pane on Workbench:

````
---
title: "Test Dashboard"
format: dashboard
---

# Tab 1

This is a tab

# Tab 2

This is another tab
````

Click Tab 2. The tab switches inside the Viewer, and the whole dashboard also opens in a new browser tab.

The dashboard is the smallest reproduction, but this is not specific to dashboards or to tabs. Any internal link in a Quarto preview is affected, on any deployment that serves the preview through a proxy.

## What's happening

`handleExternalLinks` ([links.ts:14-16](https://github.com/quarto-dev/quarto-cli/blob/main/src/webui/quarto-preview/src/frame/links.ts#L14-L16)) hooks every link it decides is external, so those clicks can be handed to the browser. It decides by comparing the link's `href` against an `origin` that the preview server infers from the request (`devserverOrigin`, [http-devserver.ts:329](https://github.com/quarto-dev/quarto-cli/blob/main/src/core/http-devserver.ts#L329)).

On Workbench the server sees `http://localhost:5158` while the browser sees `https://<workbench-host>/s/<session>/p/<token>/…#tab-2`. The comparison is a `startsWith` between those two, so it fails, and every internal link gets hooked and posted as `openExternal`. The two values agree only when the browser reaches the preview at the same address the server sees, which a proxy rules out.

Comparing `window.location.origin` instead of the server-inferred value would classify these correctly, since the code runs in the browser and the browser already knows its own origin.

Tracked on the Positron side as posit-dev/positron#14459.

melissa-barca:
## Description

Clicking a link inside a Quarto preview in Positron on Posit Workbench opened the page in a new browser tab instead of navigating inside the Viewer pane. Links are now classified against the origin the browser is on, so internal links stay in the pane.

Quarto's own preview server infers its origin from the request and sees localhost, while the browser is on the Workbench host. Comparing those two classified every internal link as external, on any deployment that serves a preview through a proxy. Comparing parsed origins also resolves relative hrefs, which the previous prefix comparison always treated as external.

Verified in Positron on a Posit Workbench deployment: a dashboard tab click stays in the Viewer pane, a relative in-document link stays in the pane, the toolbar "Open in Browser" still opens the system browser, and a genuinely external link still opens the system browser.

## Checklist

I have (if applicable):

- [x] added new tests

<details>

</details>

Fixes #14865


cscheid:
`quarto-preview` is a fraught bit of our codebase, so we especially appreciate the PR. This looks good to me. It would be even better if we knew for sure that we have an e2e test that exercises `quarto preview` in the common case. (@cderv will know for sure).

cderv (approved):
No, we don't have an automated e2e test that exercises `quarto preview`. Nothing in `tests/` starts the preview server, and the Playwright suite serves pre-rendered files, so `quarto-preview.js` is never loaded there. Expanding Playwright to cover it is tracked in #10696. 
We only have manual testing through `/quarto-preview-test` and `tests/docs/manual/preview` cases. 

I verified this one by hand instead, with a viewer-like iframe host page and a reverse proxy on a second origin, so the browser origin differs from the one the preview server infers. On main an internal link click posts `openExternal` to the parent, with this PR it stays in the pane, and a genuinely external link still posts `openExternal`. On a plain localhost preview both behave identically, so the common local case is unchanged.

PR #14876, opened 2026-09-09
fix(html): prepend floating color-scheme toggle

### LINKED ISSUE #14838: a11y: the floating color-scheme toggle is last in reading and focus order
Split from the review of #14739.

On a document with no navbar, the color-scheme toggle is injected by the after-body script and its container is appended to `body`, so it is the last element in the DOM. Reading order and focus order both follow DOM order here, since nothing sets `tabindex`. The toggle is drawn at the top right of the page, so it is the first control a sighted user sees and the last one a keyboard or screen reader user reaches.

Measured on a document with a title block banner (author link, affiliation link, DOI link):

```
visual order      TOGGLE (top=17) -> title (top=42) -> A Person (226) -> Example Org (226) -> DOI (284)
focus order now   A Person -> Example Org -> DOI -> body link -> TOGGLE
```

A website is not affected. There the toggle sits in the navbar and is reached in navbar order, matching where it is drawn.

This is not a WCAG failure. Focus order still preserves meaning and operability, and axe reports nothing. It is a usability wrinkle, and it predates #14739: the toggle was appended to `body` before that change as well.

Inserting the container at the start of `body` instead of appending it corrects the order and leaves the drawn position untouched:

```
prepended -> position: {"top":17,"fromRight":17}   (unchanged)
prepended -> focus   : TOGGLE -> A Person -> Example Org -> DOI -> body link
```

The container must stay a direct child of `body`: the toggle is positioned against the page, so any ancestor that establishes a containing block captures it (see #14739).

Two things to check before making the change:

- It alters focus order for every existing plain document with a light/dark theme pair.
- The container would become `body`'s first child, so any `body > *:first-child` rule in a theme could pick it up.

On a plain document without a banner the toggle is drawn level with the title but to its right, so prepending puts it one position early rather than exactly right. That is still much closer than last on the page.

Part of #8706.

johnfinnerty-nz:
## Description

Place the floating color-scheme toggle container first in the body so its DOM order matches its visual position. Add a Playwright assertion that protects the insertion order.

Closes #14838

## Testing

- `npx playwright test tests/html-color-scheme-toggle.spec.ts --list` (9 tests discovered)
- `npx prettier --check src/resources/formats/html/templates/quarto-html-after-body.ejs tests/integration/playwright/tests/html-color-scheme-toggle.spec.ts`
- Source contract check and `git diff --check`

Full browser execution was not run because the generated fixture sites and browser dependencies are not available in this checkout.

## Checklist

I have (if applicable):

- [x] added new tests
- [ ] created a separate documentation PR in Quarto's website repo and linked it to this PR

<details>

- **AI tool used**: OpenAI Codex
- **Codebase grounding**: Local clone of `quarto-dev/quarto-cli`
- **Human review**: John Finnerty reviewed the prepared change summary and authorized submission. Verification performed by Codex is listed above.

</details>

### DIFF
--- a/src/resources/formats/html/templates/quarto-html-after-body.ejs
+++ b/src/resources/formats/html/templates/quarto-html-after-body.ejs
@@ -55,7 +55,7 @@
       }
       container.classList.add('quarto-color-scheme-toggle-container');
       container.appendChild(toggle);
-      window.document.body.appendChild(container);
+      window.document.body.prepend(container);
     }
     setColorSchemeToggle(hasAlternateSentinel())
 
@@ -716,4 +716,3 @@
   
   </script>
   <% } %>
-  
--- a/tests/integration/playwright/tests/html-color-scheme-toggle.spec.ts
+++ b/tests/integration/playwright/tests/html-color-scheme-toggle.spec.ts
@@ -20,6 +20,9 @@ test("fallback color-scheme toggle sits in a header landmark", async ({
     "body > header.quarto-color-scheme-toggle-container",
   );
   await expect(container).toHaveCount(1);
+  await expect(page.locator("body > :first-child")).toHaveClass(
+    /quarto-color-scheme-toggle-container/,
+  );
   await expect(
     container.locator("button.quarto-color-scheme-toggle"),
   ).toHaveCount(1);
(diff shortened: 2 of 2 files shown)


cscheid:
What bug does this fix? We're not inclined to make changes to everyone's web pages with little reason.

johnfinnerty-nz:
Thanks for asking. The issue I was trying to address is the focus-order mismatch documented in #14838: on a plain document, the floating color-scheme toggle is drawn at the top right but appended to the end of body, so keyboard and screen-reader users reach it after the page content. It is a usability concern, not a WCAG failure.

I agree that changing DOM order on every affected page needs stronger regression evidence, especially around theme selectors and full browser behavior. I did not complete that validation, so I am closing this rather than asking the project to accept an under-tested global change. Sorry for the noise.

PR #14878, opened 2026-09-10
Add Azerbaijani localization

abdanar:
This adds Azerbaijani (`az`) translations for Quarto and updates the language list in the README.


cderv (approved):
Thanks for adding the Azerbaijani localization!

PR #14925, opened 2026-09-22
Consolidate binary-mode env-strip lists into a shared data file

cderv:
`tests/quarto-cmd.ts`'s `kStripEnvVars`, `run-tests.ps1`'s `$probeStrip`, and `run-tests.sh`'s inline `env -u` chain hand-maintained three copies of the same environment-variable list, introduced together in #14706. They had already drifted at that point: the shell/PowerShell probes were missing 7 names present in the TypeScript list, including `QUARTO_VERSION_REQUIREMENT`.

## Fix

All three now read a single tracked file, `tests/binary-mode-strip-env.txt`, at runtime, with identical parse semantics: trim (including a trailing CR), skip blank lines and `#` comments, validate each name against `^[A-Za-z_][A-Za-z0-9_]*$`, and fail closed on a missing, unreadable, empty, or malformed file.

`buildBinaryEnv()` is split into a pure `sanitizeBinaryEnv(ambient, overlay)` so the strip/overlay contract (overlay applied after removal, may reintroduce a stripped name) is unit-testable without mutating `Deno.env` directly.

## Verification

A CI guard that set `QUARTO_VERSION_REQUIREMENT` on the binary-mode workflow steps was considered and dropped. Both launcher scripts special-case `--version` to print a static version and exit without invoking Deno at all, so no environment variable can affect that probe's outcome regardless of whether the strip list is correct — the guard would have been dead code.

## Test plan

- [x] New unit tests (`tests/unit/binary-mode-strip-env.test.ts`) cover the parsed list shape and the `sanitizeBinaryEnv()` strip/overlay/non-mutation contract
- [x] Verified each of the three readers fails closed when the list file is renamed away, and rejects a malformed entry such as `FOO=bar`
- [x] Ran the real preflight probe in both `run-tests.sh` and `run-tests.ps1` against an installed quarto outside the checkout, confirming the parsed 16-name list strips correctly end to end

## Checklist

I have (if applicable):

- [x] added new tests


cderv:
@copilot resolve the merge conflicts in this pull request

PR #14930, opened 2026-09-22
[backport] Fix original.code leaking into every chunk's .cell div

cderv:
> [!IMPORTANT]
> Backport from #14928

Cherry-pick fix for `original.code` leaking into every chunk's `.cell` div when rendering with dev knitr (yihui/knitr#2521).


cderv:
It seems I need to update R environment too in v1.10 branch...

